Issue #11, opened 2021-01-27
Website Reloads on routing


MehtaManan07:
Everywhere in the project, hrefs are given which reloads the website every time someone navigates. It needs to be corrected and instead, **Links** from **react-router-dom** should be used. 
I am willing to work on the same and few other aspects of the project Casdoor

hsluoyz:
Already fixed by: https://github.com/casbin/casdoor/commit/ee1e08ae209832bf6c32059c04cc0e6cf01eb0f4

PR #12, opened 2021-01-27
Implemented routing efficiently

MehtaManan07:
The solution to issue #11 
I have implemented routing efficiently which will now prevent reloading of the website every time someone navigates through it. I have also removed unnecessary imports of some components which were not used in the particular files. Now the website acts as a Single Page Application. 
#11


hsluoyz:
@MehtaManan07 plz fix DCO.

PR #22, opened 2021-03-14
Support avatar tailoring and updating

Kininaru:

fixed type errors

fixed the wrong folder

Add: https://github.com/casbin/casdoor/issues/18


hsluoyz:
We should not depend on any specific cloud provider.

PR #32, opened 2021-03-29
Fix: Avatar uploading failed

Kininaru:
GetAccount API has been rewritten, so avatar uploading crashed. Fixed this bug.


hsluoyz:
Thanks for the quick fix!

PR #42, opened 2021-04-14
fix: Added support to MsSQL

GhostyJade:
Hi. I've added support to interface casdoor with a MsSQL database.
The only issue is that you can't create the database by engine.Exec and you must specify the database name in the connection string (I've updated the README to explain it, if needed)

### DIFF
--- a/authz/authz.go
+++ b/authz/authz.go
@@ -27,7 +27,7 @@ var Enforcer *casbin.Enforcer
 func InitAuthz() {
 	var err error
 
-	a, err := xormadapter.NewAdapter("mysql", beego.AppConfig.String("dataSourceName")+beego.AppConfig.String("dbName"), true)
+	a, err := xormadapter.NewAdapter(beego.AppConfig.String("dbAdapter"), beego.AppConfig.String("dataSourceName"), true)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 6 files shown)


authz/authz.go
```diff
@@ -27,7 +27,7 @@ var Enforcer *casbin.Enforcer
 func InitAuthz() {
 	var err error
 
-	a, err := xormadapter.NewAdapter("mysql", beego.AppConfig.String("dataSourceName")+beego.AppConfig.String("dbName"), true)
+	a, err := xormadapter.NewAdapter(beego.AppConfig.String("dbAdapter"), beego.AppConfig.String("dataSourceName"), true)
```
hsluoyz, line 30: why remove dbName?

hsluoyz (changes requested):

authz/authz.go
GhostyJade: I did it because mysql concat the db name to the connection string, mssql not.

I didn't want to add an if/else statement since, if you plan to support multiple databases, it will make the code a bit unreadable...

Kininaru: We have rewritten `app.conf` to adapt Postgres. Now, if you want to use Postgres as database, you can write like this: 
``` ini
db = postgres
dataSourceName = "user=postgres password=xxx sslmode=disable dbname="
dbName = casdoor
```
Please notice here:  `dbname=`. This is the end of `dataSourceName`. I think MsSQL and Postgres are similar here. So I suggest you pull our latest code and keep the parameter `dbName` in `app.conf`. This way, `app.conf` will suit both MySQL MsSQL and Postgres.

Yesterday, we added a new parameter `db` in `app.conf`. I think it has the same effect as your `dbAdapter`. You can use `db` instead.

Issue #38, opened 2021-04-11
calling "api/register" even with correct data feilds gives error


sourabhmandal:
I called "api/register" with following json object
`
{
    "owner" : "builtin",
    "username" : "sourabh",
    "password" : "123",
    "name" : "sourabh mandal",
    "email" : "19mandal97@gmail.com",
    "phone" : "999"
}
`
But I got the following error

![a](https://user-images.githubusercontent.com/39632788/114298694-4b7b1080-9ad5-11eb-9828-62e2340a645c.png)

The `GetOwnerAndNameFromId()` expect a `username : "sourabh/organization`

The route works fine when I pass following payload
`
{
    "owner" : "builtin",
    "username" : "sourabh/organisation",
    "password" : "123",
    "name" : "sourabh mandal",
    "email" : "19mandal97@gmail.com",
    "phone" : "999"
}
`

Please tell me if its an error in CODE or the JSON PAYLOAD

sourabhmandal:
I would suggest refactoring `GetOwnerAndNameFromId()` to eleminate this error

hsluoyz:
@sourabhmandal this is not a bug. And we don't have enough time to do the input checking for now because this is endless.

Plz do some issues listed here: https://github.com/casbin/casdoor/issues

PR #43, opened 2021-04-14
fix: User registration route (api/register)

sourabhmandal:
This PR solves #38 .


hsluoyz:
See: https://github.com/casbin/casdoor/issues/38#issuecomment-828945491

PR #52, opened 2021-05-09
fix: removed a duplicate function GetId

HatsuneBQ:


HatsuneBQ:
Same function here: https://github.com/casbin/casdoor/blob/master/object/user.go#L248

PR #55, opened 2021-05-12
feat: add reset Email and phone by using verification code.

Kininaru:


Kininaru:
Deployed at: https://door.chromium.link/account

User can only reset their own email or phone.

Kininaru:
Now, admin users can set phone prefix for organizations in org setting pages.

PR #56, opened 2021-05-15
Fixed reset phone email data error and removed dom API from ResetModal.js

Kininaru:
Fixed: 
- reset phone email data error
- send code button background color error
- remoted dom api from `ResetModal.js`


Kininaru:
Deployed at: https://door.chromium.link/

`admin 123456`

PR #63, opened 2021-05-17
feat: check email and phone number when signing up

Kininaru:

Click to try sign up: https://door.chromium.link/signup


Kininaru:
@hsluoyz Resolved conflicts.

PR #76, opened 2021-06-01
feat: authorize via clientId and clientSecret

Kininaru:
If there is a client id and a client secret in URL, Casdoor will treat this request as `built-in/service`'s.

### DIFF
--- a/object/application.go
+++ b/object/application.go
@@ -116,7 +116,7 @@ func GetApplicationByUser(user *User) *Application {
 	return GetApplicationByOrganizationName(user.Owner)
 }
 
-func getApplicationByClientId(clientId string) *Application {
+func GetApplicationByClientId(clientId string) *Application {
 	application := Application{}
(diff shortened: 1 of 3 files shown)


Kininaru:
Deployed at: https://door.chromium.link/

- `admin` is global admin
- `user_2` is admin of `built-in`
- `user_1` is a normal user

All of their passwords are `123456`

hsluoyz:
@Kininaru plz use Casbin

Kininaru:
@hsluoyz I tried to use Casbin but it seems like that the Enforcer in `authz/authz.go` is not able to suit the situation. Can we add another Enforcer to manage user permissions?

hsluoyz:
@Kininaru 

1. What are `organization owners`?
2. What's the difference between built-in and other orgs now? built-in org members are supposed to have more privileges.

Kininaru:
@hsluoyz plz review, deployed at: https://door.chromium.link

PR #85, opened 2021-06-06
fix: redirect urls null in app-built-in

Kininaru:


hsluoyz:
@Kininaru thanks for finding the cause! Fixed in another way: https://github.com/casbin/casdoor/commit/98ec187019edc0a1a6fad0d22b857c19fad89463

PR #92, opened 2021-06-08
feat: allow user signup with oauth

WindSpiritSR:
1. Allow user sign up with OAuth account
2. Fix that `GetUserByField(org, providerType, providerId)` if `providerId` is empty, may return a wrong user

Deployed at: https://casdoor.leviatan.cn
But this site does not have phone verify so user cannot signup. 

I test this pr in local machine by disable all verify, user can successful signup with OAuth account:

image

Signed-off-by: WindSpirit <simon343riley@gmail.com>

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -50,6 +50,12 @@ type RequestForm struct {
 	EmailCode   string `json:"emailCode"`
 	PhoneCode   string `json:"phoneCode"`
 	PhonePrefix string `json:"phonePrefix"`
+
+	Github   string `json:"github"`
+	Google   string `json:"google"`
+	Qq       string `json:"qq"`
+	Wechat   string `json:"wechat"`
+	Facebook string `json:"facebook"`
 }
 
 type Response struct {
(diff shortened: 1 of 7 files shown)


hsluoyz:
@WindSpiritSR I don't think you implemented it in the correct way..

@Kininaru plz also have a review.

WindSpiritSR:
@hsluoyz emmmmmm
demo gif:

demo

Kininaru:
I think your way of implement is not correct...

Now we have already supported OAuth login, look at here: 
https://github.com/casbin/casdoor/blob/master/controllers/auth.go#L246

Please focus on the else statement.

Maybe you can just modify the else statement to implement the function: If a new user logged in via OAuth and the application signup is enabled, add this user directly into database instead of responding `The account for provider: %s and username: %s does not exist, please create an account first`.

Kininaru (commented):
I think your way of implement is not correct...

Now we have already supported OAuth login, look at here: 
https://github.com/casbin/casdoor/blob/master/controllers/auth.go#L246

Please focus on the else statement.

Maybe you can just modify the else statement to implement the function: If a new user logged in via OAuth and the application signup is enabled, add this user directly into database instead of responding `The account for provider: %s and username: %s does not exist, please create an account first`.

WindSpiritSR:
sorry, i misunderstand the issue😂

PR #96, opened 2021-06-09
fix: resolved nil pointer dereference bugs.

sh1luo:

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -178,8 +178,12 @@ func (c *ApiController) GetAccount() {
 	var resp Response
 
 	user := object.GetUser(userId)
-	organization := object.GetOrganizationByUser(user)
-	resp = Response{Status: "ok", Msg: "", Data: user, Data2: organization}
+	if user != nil {
+		organization := object.GetOrganizationByUser(user)
+		resp = Response{Status: "ok", Msg: "", Data: user, Data2: organization}
+	} else {
+		resp = Response{Status: "error", Msg: "Please sign in first"}
+	}
 
 	c.Data["json"] = resp
 	c.ServeJSON()
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -94,7 +94,10 @@ func (c *ApiController) Login() {
 	var form RequestForm
 	err := json.Unmarshal(c.Ctx.Input.RequestBody, &form)
 	if err != nil {
-		panic(err)
+		resp = &Response{Status: "error", Msg: err.Error()}
+		c.Data["json"] = resp
+		c.ServeJSON()
+		return
 	}
 
 	if form.Username != "" {
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -163,8 +163,9 @@ func (c *ApiController) Login() {
 				break
 			}
 		} else {
+			tmpUser := object.GetUserByFields(form.Organization, form.Username)
 			password := form.Password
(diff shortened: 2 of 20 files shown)


sh1luo:
This bug caused me to encounter a null pointer reference, which happened when calling `GetOrganizationByUserwhen(user *User) *Organization`, argument *user is nil

context：
```
2021/06/09 15:00:25.086 [C]  the request url is  /api/get-account
2021/06/09 15:00:25.086 [C]  Handler crashed with error runtime error: invalid memory address or nil pointer dereference
2021/06/09 15:00:25.086 [C]  /usr/lib/golang/src/runtime/panic.go:969
2021/06/09 15:00:25.086 [C]  /usr/lib/golang/src/runtime/panic.go:212
2021/06/09 15:00:25.086 [C]  /usr/lib/golang/src/runtime/signal_unix.go:742
2021/06/09 15:00:25.086 [C]  /root/casdoorr/object/organization.go:97
2021/06/09 15:00:25.086 [C]  /root/casdoorr/controllers/account.go:181
2021/06/09 15:00:25.086 [C]  /usr/lib/golang/src/reflect/value.go:476
2021/06/09 15:00:25.086 [C]  /usr/lib/golang/src/reflect/value.go:337
2021/06/09 15:00:25.086 [C]  /root/go_workpath/pkg/mod/github.com/astaxie/beego@v1.12.3/router.go:897
2021/06/09 15:00:25.086 [C]  /usr/lib/golang/src/net/http/server.go:2843
2021/06/09 15:00:25.086 [C]  /usr/lib/golang/src/net/http/server.go:1925
2021/06/09 15:00:25.087 [C]  /usr/lib/golang/src/runtime/asm_amd64.s:1374
2021/06/09 15:00:25.087  [HTTP] http: superfluous response.WriteHeader call from github.com/astaxie/beego/context.(*Response).WriteHeader (context.go:230)
```

sh1luo:
wait, it seems that the front-end cannot accept the organization as nil.

image

hsluoyz:
@sh1luo when does this happen?

sh1luo:
When the user does **not exist** in the database, the function `func (c *ApiController) GetAccount()` under the controllers package will still be called, and the return value user is nil. At this time, calling `object.GetOrganizationByUser(user)` will causes the program to crash because of  nil pointer reference. even if the object.GetOrganizationByUser(user) function is judged, the front end will still report an error, "can't set property "organization" of null" as seen in the previous picture.

I think the function `func (c *ApiController) GetAccount()` should be modified to:
```go
func (c *ApiController) GetAccount() {
    userId, ok := c.RequireSignedIn()
    if !ok {
        return
}

var resp Response

user := object.GetUser(userId)
    if user != nil {
    organization := object.GetOrganizationByUser(user)
    resp = Response{Status: "ok", Msg: "", Data: user, Data2: organization}
} else {
    resp = Response{Status: "error", Msg: "Please sign in first"}
}

c.Data["json"] = resp
c.ServeJSON()
}
```
In this way, [line #148 in web/src/App.js](https://github.com/casbin/casdoor/blob/master/web/src/App.js#L148) can throw an error normally.

This is a screenshot, there are only default users in the database (users without name=主管理员)

image

sh1luo:
sorry,It seems that there is still a problem, let me take a look

sh1luo:
This screenshot is a little strange, I think it has no reference value.

However, the problem with function `func (c *ApiController) GetAccount()` still exists, it's just latent,  we haven't found it yet, so this code is still necessary :)

PR #102, opened 2021-06-13
feat: Allow the user to sign up with OAuth

WindSpiritSR:
Allow the user to sign up when he tries to use OAuth to login but there's no account linked to the OAuth

Deployed at: https://casdoor.leviatan.cn/login/oauth/authorize?client_id=00c5f3410de2ebf737e4&response_type=code&redirect_uri=https://www.leviatan.cn&scope=read&state=casdoor

image

### DIFF
--- a/web/src/ApplicationEditPage.js
+++ b/web/src/ApplicationEditPage.js
@@ -201,7 +201,7 @@ class ApplicationEditPage extends React.Component {
           </Col>
           <Col span={22} >
             <UrlTable
-              title="Redirect URLs"
+              title={i18next.t("application:Redirect URLs")}
               table={this.state.application.redirectUris}
               onUpdateTable={(value) => { this.updateApplicationField('redirectUris', value)}}
             />
--- a/web/src/UrlTable.js
+++ b/web/src/UrlTable.js
@@ -16,6 +16,7 @@ import React from "react";
 import {DownOutlined, DeleteOutlined, UpOutlined, LinkOutlined} from '@ant-design/icons';
 import {Button, Col, Input, Row, Table, Tooltip} from 'antd';
 import * as Setting from "./Setting";
+import i18next from "i18next";
 
 class UrlTable extends React.Component {
   constructor(props) {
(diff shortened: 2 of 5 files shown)


hsluoyz:
@Kininaru plz review.

Kininaru:
image

It looks like that something wrong with the facebook OAuth APP...

Kininaru:
image

WindSpiritSR:
@Kininaru @huazhikui Has changed to use GitHub OAuth now

controllers/account.go
```diff
@@ -115,7 +118,29 @@ func (c *ApiController) Signup() {
...
+
+			object.UpdateUserForOriginal(user)
+
+			object.DisableVerificationCode(form.Email)
+			object.DisableVerificationCode(checkPhone)
+			util.LogInfo(c.Ctx, "API: [%s] is signed up as new user", userId)
+			resp = Response{Status: "ok", Msg: "", Data: userId}
+		} else {
+			resp = Response{Status: "error", Msg: msg, Data: ""}
+		}
```
Kininaru, line 143: These lines do not add the new user to the database. `object.UpdateUserForOriginal` can only update user, but not insert a new user.

And I think we can not add a user here, because when `msg != ""`, the new user information is invalid, and we should not add an invalid user to the database.

web/src/auth/ProviderSignupPage.js
```diff
@@ -0,0 +1,440 @@
+// Copyright 2021 The casbin Authors. All Rights Reserved.
```
Kininaru, line 1: There are some bugs in this page, like this:
image

And we already have a signup page, how about reusing the page here?

Kininaru (changes requested):

web/src/auth/ProviderSignupPage.js
WindSpiritSR: This cause I didn't set email and SMS verification server, already working on it.
Reusing the page may transfer too many params, is that ok?

controllers/account.go
WindSpiritSR: At f672045, already adds user basic info(like org, owner, id) at `controllers/auth.go` line `264` if the application enable user signup
These lines only for complete and validate user info from OAuth if they want.

PR #105, opened 2021-06-14
fix: limit token display length at token page

WindSpiritSR:
1. Token string is too long caused performance low, so limit it to 20 characters only for display
2. Add `Tooltip` for `getClickable()` function
3. Add param `display` for `getClickable()`, what is it used for display text is different from the text you want to copy
4. fixed `Access token` row cannot be hidden if the browser width too small

### DIFF
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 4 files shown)


WindSpiritSR:
Deployed at: https://casdoor.leviatan.cn/tokens
user: `admin`
password: `123`

PR #126, opened 2021-06-22
fix: Application null error

oranges-eating:
Added application default organization to avoid reporting null errors
Signed-off-by: killer <1533063601@qq.com>


hsluoyz:
lgtm

PR #127, opened 2021-06-22
fix: Continue to resolve application null error

oranges-eating:
demo:http://20.194.6.21:8000/
usename:admin
password:123
fix: #126


hsluoyz:
@oranges-eating signupItems is well initialized for new app, you should modify your DB data. This PR is unnecessary.

https://github.com/casbin/casdoor/blob/3274bd0c7c5e8443aab4e312f5a95232b2913377/web/src/ApplicationListPage.js#L56-L65

PR #130, opened 2021-06-24
fix: cant jump to sign in page after registration

sh1luo:


sh1luo:
In order not to conflict with the previous pr demo, I created another app(application_1), to fix the bug raised by users.

demo sign up link: http://test.rescld.cn/signup/application_1

Log in with your account after registration.

Management background: http://test.rescld.cn
```
admin,123
```

One more user belonging to this app(application_1) is added to the user table.

PR #131, opened 2021-06-24
feat: add password signupItem forcibly when enabled password checked

sh1luo:


sh1luo:
demo link: http://test.rescld.cn/applications/app-built-in

add password signupItem forcibly when enabled password and set password Item cannot be deleted

sh1luo:
replaced by: #132

PR #133, opened 2021-06-25
feat: admin can update user's organization

oranges-eating:
demo:http://www.killerwing.xyz:8000/
admin/123
Signed-off-by: killer <1533063601@qq.com>

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -141,7 +141,7 @@ func UpdateUser(id string, user *User) bool {
 
 	user.UpdateUserHash()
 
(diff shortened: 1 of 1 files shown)


hsluoyz:
@oranges-eating plz avoid the user itself from changing his own organization

oranges-eating:
> @oranges-eating plz 避​​免用户自己改变自己的组织

Now users cannot choose their own organization in their profile

hsluoyz:
@oranges-eating but they can call API to change it, right?

oranges-eating:
> @oranges-eating but they can call API to change it, right?


They can’t use the API to change their organization and will return 403

Issue #89, opened 2021-06-07
Add tooltips for all fields in all XXXListPage.js and XXXEditPage.js


hsluoyz:
We should add tooltip for each field, when a user hovers on it, it will show the description for that field. Like:

![image](https://user-images.githubusercontent.com/3787410/120970594-3998af80-c79e-11eb-9b75-0f905d9e0bfc.png)

I have made an example: https://github.com/casbin/casdoor/commit/234a9b906076657be1bc83ac48e1eeba5301a9ef

It will show in your localhost instance at: http://localhost:7001/applications/app-built-in

![image](https://user-images.githubusercontent.com/3787410/123299219-df447080-d54b-11eb-83c7-dfe39fbc06a8.png)

sh1luo:
Which file is related to this modification

hsluoyz:
@sh1luo all xxx-edit pages

sh1luo:
ok

hsluoyz:
@sh1luo

PR #135, opened 2021-06-25
feat: add tooltips for fields of edit pages

sh1luo:
Fix: #89 


sh1luo:
demo link: http://testb.rescld.cn/

PR #145, opened 2021-07-04
fix: null properties bug

sh1luo:


sh1luo:
This bug will report an error when using the admin default account to access the `http://xx.xx.xx.xx/account` page, and the page cannot be displayed normally. 

Because the user.Properties are not initialized, the value of the field in the database is `null`, resulting in a null pointer reference error

PR #153, opened 2021-07-07
feat: auto login session will expire after 24h

Kininaru:
Fix: https://github.com/casbin/casdoor/issues/109

Video: (Modified database to simulate login expiration)

https://user-images.githubusercontent.com/42638489/124791965-259fc380-df7f-11eb-84f7-8209a256ac24.mp4

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -78,8 +78,8 @@ type HumanCheck struct {
 func (c *ApiController) Signup() {
 	var resp Response
 
-	if c.GetSessionUser() != "" {
-		c.ResponseErrorWithData("Please sign out first before signing up", c.GetSessionUser())
+	if c.GetSessionUsername() != "" {
+		c.ResponseErrorWithData("Please sign out first before signing up", c.GetSessionUsername())
 		return
 	}
 
@@ -161,7 +161,7 @@ func (c *ApiController) Signup() {
 
 	if application.HasPromptPage() {
 		// The prompt page needs the user to be signed in
-		c.SetSessionUser(user.GetId())
+		c.SetSessionUsername(user.GetId())
 	}
 
 	object.DisableVerificationCode(form.Email)
@@ -181,10 +181,10 @@ func (c *ApiController) Signup() {
 func (c *ApiController) Logout() {
 	var resp Response
 
-	user := c.GetSessionUser()
+	user := c.GetSessionUsername()
 	util.LogInfo(c.Ctx, "API: [%s] logged out", user)
 
-	c.SetSessionUser("")
+	c.SetSessionUsername("")
 
 	resp = Response{Status: "ok", Msg: "", Data: user}
 
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -52,6 +52,8 @@ type RequestForm struct {
 	EmailCode   string `json:"emailCode"`
 	PhoneCode   string `json:"phoneCode"`
 	PhonePrefix string `json:"phonePrefix"`
+
+	AutoSignin bool `json:"autoSignin"`
 }
 
 type Response struct {
@@ -185,6 +187,7 @@ func (c *ApiController) Logout() {
 	util.LogInfo(c.Ctx, "API: [%s] logged out", user)
 
 	c.SetSessionUsername("")
+	c.SetSessionData(nil)
 
 	resp = Response{Status: "ok", Msg: "", Data: user}
 
--- a/controllers/base.go
+++ b/controllers/base.go
@@ -14,13 +14,32 @@
 
 package controllers
 
-import "github.com/astaxie/beego"
+import (
+	"time"
+
+	"github.com/astaxie/beego"
+	"github.com/casdoor/casdoor/util"
+)
 
 type ApiController struct {
 	beego.Controller
 }
 
+type SessionData struct {
+	ExpireTime int64
+}
+
 func (c *ApiController) GetSessionUsername() string {
+	// check if user session expired
+	sessionData := c.GetSessionData()
+	if sessionData != nil &&
+		sessionData.ExpireTime != 0 &&
+		sessionData.ExpireTime < time.Now().Unix() {
+		c.SetSessionUsername("")
+		c.SetSessionData(nil)
+		return ""
+	}
+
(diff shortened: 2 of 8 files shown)


Kininaru:
@hsluoyz plz review

object/user.go
```diff
@@ -22,10 +22,11 @@ import (
...
 type User struct {
-	Owner       string `xorm:"varchar(100) notnull pk" json:"owner"`
-	Name        string `xorm:"varchar(100) notnull pk" json:"name"`
-	CreatedTime string `xorm:"varchar(100)" json:"createdTime"`
-	UpdatedTime string `xorm:"varchar(100)" json:"updatedTime"`
+	Owner           string `xorm:"varchar(100) notnull pk" json:"owner"`
+	Name            string `xorm:"varchar(100) notnull pk" json:"name"`
+	CreatedTime     string `xorm:"varchar(100)" json:"createdTime"`
+	UpdatedTime     string `xorm:"varchar(100)" json:"updatedTime"`
+	LoginExpireTime string `xorm:"varchar(100)" json:"loginExpireTime"`
```
hsluoyz, line 29: Login -> Signin

controllers/account.go
```diff
@@ -52,6 +52,7 @@ type RequestForm struct {
 	EmailCode   string `json:"emailCode"`
 	PhoneCode   string `json:"phoneCode"`
 	PhonePrefix string `json:"phonePrefix"`
+	Remember    bool   `json:"remember"`
```
hsluoyz, line 55: Remember -> AutoSignin

hsluoyz (changes requested):

controllers/base.go
```diff
@@ -26,6 +32,24 @@ func (c *ApiController) GetSessionUser() string {
...
+		return ""
+	}
+
+	userObj := object.GetUser(userId)
+	if userObj == nil {
+		return ""
+	}
+
+	// if user login expired, then clean the session and return an empty string
+	if userObj.LoginExpireTime != "" && userObj.LoginExpireTime < util.GetCurrentTime() {
```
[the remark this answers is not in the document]
Kininaru: `LoginExpireTime` and `Util.GetCurrentTime` are strings like `2021-06-07T21:45:58+08:00`. 

`year - month - day - hour - minute - second`

We can use `<` to compare the ASCII of the first different char between 2 time strings to tell which is earlier.

object/user.go
Kininaru: @hsluoyz Fixed.

controllers/account.go
Kininaru: @hsluoyz Fixed.

controllers/base.go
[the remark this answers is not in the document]
hsluoyz: @Kininaru but what about the timezone?

+08:00 > +07:00 (may be right)
-08:00 > -07:00 (wrong!)

You should use a formal date comparison in Go.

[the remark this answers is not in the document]
hsluoyz: @Kininaru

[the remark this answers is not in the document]
Kininaru: Got that, I am going to use Unix time stamp instead of this, is that OK?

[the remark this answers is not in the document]
Kininaru: @hsluoyz Fixed that, wrote a new function:
```go
func IsTimeExpired(timeString string) bool {
	t, err := time.Parse(time.RFC3339, timeString)
	if err != nil {
		panic(err)
	}

	return t.Unix() < time.Now().Unix()
}
```

controllers/base.go
```diff
@@ -26,6 +32,24 @@ func (c *ApiController) GetSessionUser() string {
...
+		return ""
+	}
+
+	userObj := object.GetUser(userId)
+	if userObj == nil {
+		return ""
+	}
+
+	// if user login expired, then clean the session and return an empty string
+	if userObj.SigninExpireTime != "" && util.IsTimeExpired(userObj.SigninExpireTime) {
```
hsluoyz, line 46: What if the user logoff and then login?

What if he has multiple sessions?

hsluoyz (changes requested):

controllers/base.go
Kininaru: These codes change the session, and is not that strongly related to accounts.

If a user has many sessions, each of the session is independent. Sessions with "auto sign in" won't expire, and sessions without "auto sign in" will expire.

hsluoyz: But won't `object.UpdateUserInternal(userId, userObj)` write to DB?

Kininaru: @hsluoyz Oh, that's right... I did not take this into consideration.

Can I add a time stamp to the session, so the session will become something like this: `built-in/admin 1626358811`

Then add some code to `GetSessionUser` to recognize the time stamp.

Kininaru: @hsluoyz Resolved that. Now, exipre timestamps are stored in session.

Checked auto signin:


Did not check:
$( _LRYFMW4ACOUVB4OES5N

hsluoyz: @Kininaru  We can store a struct in the session, you can see Casnode's code. We don't need to put all stuff in one string and then parse it. This code is fragile..

Kininaru: @hsluoyz Can I add another session called `ExpireTime`? Or store a struct in a session called `SessionInfo`...
I found that many places use the session `username` like `built-in/admin`, turning the session to a struct (or even just add a timestamp behind user id) may cause many unexpected bugs...

hsluoyz: @Kininaru we can extract `GetSessionUsername()` first before this PR like this:

https://github.com/casbin/casnode/blob/9c66ea1de047a22c9d99e36bcd86dbc9280acb5b/controllers/base.go#L54-L60

Then we can easily change the underlying implementation of session data structure.

Kininaru: @hsluoyz Finished. Now, for each session, there is a struct called `SessionData` stored in session like this: 
```go
type SessionData struct {
	ExpireTime int64
}
```
We can add more info to the struct in the future.

hsluoyz (approved):

Issue #168, opened 2021-07-14
Translation hot reload


ErikQQY:
Casdoor is using i18next as our internationalization tool, every time we want to change to other language, we click the button and Casdoor will refresh the page and load the language we want, however, I think using hot reload might be more elegant and suitable.

Since Casdoor is based on React, I think [react-i18next](https://react.i18next.com/) can help us achieve that feature.

A react-i18next based demo: https://codesandbox.io/s/1zxox032q

More details:
https://github.com/i18next/react-i18next
https://dev.to/adrai/how-to-properly-internationalize-a-react-application-using-i18next-3hdb

hsluoyz:
Hot reload will bring the `[object object]` bug, see: https://github.com/casbin/casdoor/pull/169#issuecomment-881206568

And I don't think removing refresh is good, because changing language is very low-frequency operation, and the users need feedback like page refresh.

So closed here for now.

PR #169, opened 2021-07-15
feat: use react-i18next for translation

WindSpiritSR:
fixed: #168 

### DIFF
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -16,11 +16,11 @@ import {message, Tooltip} from "antd";
 import {QuestionCircleTwoTone} from "@ant-design/icons";
 import React from "react";
 import {isMobile as isMobileDevice} from "react-device-detect";
-import "./i18n";
-import i18next from "i18next";
+import i18n from "./i18n";
 import copy from "copy-to-clipboard";
 import {authConfig} from "./auth/Auth";
 import {Helmet} from "react-helmet";
+import { Translation } from 'react-i18next';
 
 export let ServerUrl = "";
 
@@ -285,18 +285,25 @@ export function getAvatarColor(s) {
   return colorList[random % 4];
 }
 
+export const I18n = (text) => {
+  return (
+    <Translation>
+      {t => {return t(text)}}
+    </Translation>
+  )
+}
+
 export function setLanguage() {
   let language = localStorage.getItem('language');
   if (language === undefined) {
     language = "en"
   }
-  i18next.changeLanguage(language)
+  i18n.changeLanguage(language)
 }
 
 export function changeLanguage(language) {
   localStorage.setItem("language", language)
-  i18next.changeLanguage(language)
-  window.location.reload(true);
+  i18n.changeLanguage(language)
 }
 
 export function getClickable(text) {
--- a/web/package.json
+++ b/web/package.json
@@ -22,7 +22,7 @@
     "react-dom": "^17.0.2",
     "react-github-corner": "^2.5.0",
     "react-helmet": "^6.1.0",
-    "react-i18next": "^11.8.7",
+    "react-i18next": "^11.11.2",
     "react-router-dom": "^5.2.0",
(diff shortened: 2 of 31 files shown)


ErikQQY:
Wow, it’s awesome, maybe we should communicate with each other in advance. That issue is an idea, I think that feature could bring comfortable experiences to our users, do you have demo links for this PR? I would love to collaborate with you on this feature.

WindSpiritSR:
This is a great idea and easy to implement.

Deployed at https://door.leviatan.cn

hsluoyz:
@WindSpiritSR only "Home" is translated:

image

WindSpiritSR:
@hsluoyz `locales/zh/data.json` doesn't translate

ErikQQY:
I think it is because since [this commit](https://github.com/casbin/casdoor/commit/2297251dd7523892983ce46d3289a45f597828be) the translations going wrong and there are no translations, I am checking on it

web/src/Setting.js
```diff
@@ -285,18 +285,25 @@ export function getAvatarColor(s) {
   return colorList[random % 4];
 }
 
+export const I18n = (text) => {
+  return (
+    <Translation>
```
sh1luo, line 290: Why not continue to use `i18n`, `i` is lowercase.

sh1luo (commented):
Code looks good, but I want to know what is the effect after replacement using `react-i18next`, is the performance better, or other effects?

ErikQQY:
Sorry for the Crowdin commit mistake, I think it is all good now

casdoortrans

Next time a commit is pushed to the repo, the translation will working

@WindSpiritSR If you want to try the Chinese translation on you demo, you can substitute your original Chinese translation to [this](https://github.com/ErikQQY/casdoor/blob/master/web/src/locales/zh/data.json), just use this data.json instead of the original in your repo.

ErikQQY:
> Code looks good, but I want to know what is the effect after replacement using `react-i18next`, is the performance better, or other effects?

After clicking other languages, the pages will not refresh, but directly change to the language you choose

hsluoyz:
@ErikQQY the translation is still wrong, also:

1. Put the language selection to top-right, a select box.
2. Show an antd alert to know language is changed. The current way is too silent.

hsluoyz (changes requested):
@WindSpiritSR @ErikQQY the translation is still wrong, also:

1. Put the language selection to top-right, a select box.
2. Show an antd alert to know language is changed. The current way is too silent.

WindSpiritSR:
```js
export function I18n(text) {
  return (
    <Translation>
      {t => {return t(text)}}
    </Translation>
  )
}
```

These thing was **wrong**, because this function return a DOM object, if the text like `<input placeholder={Setting.I18n("text")}/>`, we will get a input box that shows `[object object]`

The right way to use `react-i18next`:

```js
import { withTranslation } from 'react-i18next';
@withTranslation()
class Test extends Component {
  //...code
  {this.props.t("Text")}
  //...code
}
export default Test;
```

or

```js
import { withTranslation } from 'react-i18next';
class Test extends Component {
  //...code
  {this.props.t("Text")}
  //...code
}
export default withTranslation()(Test);
```

This usage needs a `class`, but sometimes we need to translate at a `function`, I have no ideas to make it usable and look good. :-(

web/src/Setting.js
hsluoyz: @sh1luo I agree

@WindSpiritSR plz change

hsluoyz:
@WindSpiritSR yeah, I also see this `[object object]` bug.

> These thing was wrong, because this function return a DOM object, if the text like <input placeholder={Setting.I18n("text")}/>, we will get a input box that shows [object object]

I don't understand, this is your PR, right? Why do you say this is wrong instead of fixing it? It's just like you are reviewing a PR from others.

hsluoyz:
@WindSpiritSR and plz fix:

image

PR #191, opened 2021-07-24
feat: add twitter provider

sh1luo:


sh1luo:
Twitter is different from other providers. I wrote a demo according to [the official document](https://developer.twitter.com/en/docs/authentication/api-reference/request_token), but I reported this error anyway.

image

I want to ask others opinions. You can clone this pr to run locally, observe the console, and compare with the official results given by Twitter to try to find out my problem.

thanks

PR #196, opened 2021-07-25
feat: integrate OSS into providers

Kininaru:

Username: `admin`
Password: `123456`

Add an OSS provider, then add the provider to an application. After that, users in this application can upload files via Casdoor.

When adding or editing an OSS provider, you can leave `domain` blank. Casdoor backend will generate a domain if you don't customize it.


controllers/account.go
```diff
@@ -237,6 +237,9 @@ func (c *ApiController) UploadAvatar() {
 	var resp Response
 
 	user := object.GetUser(userId)
+	organization := object.GetOrganizationByUser(user)
+	application := object.GetApplicationByOrganizationName(organization.Name)
```
hsluoyz, line 241: Plz use: `GetApplicationByUser()` to reduce number of calls.

controllers/account.go
```diff
@@ -237,6 +237,9 @@ func (c *ApiController) UploadAvatar() {
...
+	provider := application.GetOssProvider()
```
hsluoyz, line 242: Change to: `storageProvider`

object/application_item.go
```diff
@@ -42,6 +42,10 @@ func (application *Application) GetSmsProvider() *Provider {
 	return application.getProviderByCategory("SMS")
 }
 
+func (application *Application) GetOssProvider() *Provider {
```
hsluoyz, line 45: All "Oss", "OSS" should be changed to "Storage"

hsluoyz (changes requested):

hsluoyz (approved):

PR #205, opened 2021-07-28
feat: add user info options for organization edit page

WindSpiritSR:
Deployed at: https://door.leviatan.cn/
Admin: `admin`
Password: `123`

And you can login https://forum.leviatan.cn/ with:
Admin: `admin_forum`
Password: `123`


WindSpiritSR:
1. Add `public` to control user info is visible to other users
2. Add `country/region` support for bc5e748
3. Rebase `upstream/master` and support preview `Edit User` page in `normal-user` and `admin` view
4. I will squash these PR if we ready to merge

object/user.go
```diff
@@ -116,6 +116,65 @@ func GetMaskedUser(user *User) *User {
...
+			user.WeChat = ""
+			user.Facebook = ""
+			user.DingTalk = ""
+			user.Weibo = ""
+			user.Gitee = ""
+			user.LinkedIn = ""
+			break
+		}
+	}
+	return user
```
sh1luo, line 175: Too wordy, try to solve it by reflection, like [object.GetUserField()](https://github.com/casbin/casdoor/blob/master/object/user_util.go#L92).

web/src/account/AccountPage.js
```diff
@@ -18,7 +18,9 @@ import UserEditPage from "../UserEditPage";
 class AccountPage extends React.Component {
   render() {
     return (
-      <UserEditPage organizationName={this.props.account.owner} userName={this.props.account.name} account={this.props.account} />
+      <UserEditPage organizationName={this.props.account.owner}
+                    accountItems={this.props.account.organization.accountItems} userName={this.props.account.name}
+                    account={this.props.account}/>
```
sh1luo, line 23: The Code format keeps consistent, either:
```
attr_A=...
attr_B=...
attr_C=...
```
, or:
```
attr_A=... attr_B=... attr_C=...
```
It looks better this way.

sh1luo (commented):
1. Please keep the commit tidy and remove unnecessary commits.
2. I think that `AccountTable` should be considered to re-use existing components or refactor it to satisfy most scenarios, otherwise Always write `xxxTable`, the code repetition rate is a bit high, what do you think?

object/user.go
WindSpiritSR: We can use reflection to get `Name` or `Value`, but we cannot use it set value by name like `user["Id"] = ""`

object/organization_item.go
```diff
@@ -0,0 +1,58 @@
+package object
+
+type AccountItem struct {
+	Name     string `json:"name"`
+	Visible  bool   `json:"visible"`
+	Required bool   `json:"required"`
```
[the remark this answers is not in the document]
WindSpiritSR: User updates their info from `/account` needs to know which field is required

object/application.go
```diff
@@ -32,7 +32,7 @@ type Application struct {
 	EnablePassword  bool            `json:"enablePassword"`
 	EnableSignUp    bool            `json:"enableSignUp"`
 	Providers       []*ProviderItem `xorm:"varchar(10000)" json:"providers"`
-	SignupItems     []*SignupItem   `xorm:"varchar(1000)" json:"signupItems"`
+	SignupItems     []*SignupItem   `xorm:"text" json:"signupItems"`
 	OrganizationObj *Organization   `xorm:"-" json:"organizationObj"`
```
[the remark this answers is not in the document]
WindSpiritSR: We have no idea how much data it needs to storage, if we use `varchar(1000)` to storage 2000 char it will be split.

And a normal value in my dev database, this field's value is

```json
[{"name":"ID","visible":false,"required":true,"prompted":false,"rule":"Random"},{"name":"Username","visible":true,"required":true,"prompted":false,"rule":"None"},{"name":"Display name","visible":true,"required":true,"prompted":false,"rule":"None"},{"name":"Password","visible":true,"required":true,"prompted":false,"rule":"None"},{"name":"Confirm password","visible":true,"required":true,"prompted":false,"rule":"None"},{"name":"Email","visible":true,"required":true,"prompted":false,"rule":"None"},{"name":"Phone","visible":true,"required":true,"prompted":false,"rule":"None"},{"name":"Agreement","visible":true,"required":true,"prompted":false,"rule":"None"}]
```

Its length is 661 chars, we cannot make sure it won't be 1001 chars.

> (varchar use) L + 1 bytes if column values require 0 − 255 bytes, L + 2 bytes if values may require more than 255 bytes

> (text use) L + 2 bytes, where L < 2^16

https://dev.mysql.com/doc/refman/5.7/en/storage-requirements.html#data-types-storage-reqs-strings

WindSpiritSR:
> 2. I think that `AccountTable` should be considered to re-use existing components or refactor it to satisfy most scenarios, otherwise always write `xxxTable` files, the code repetition rate is a bit high, what do you think?

Each `xxxTable` process different things in `renderTable`, it can be reused but there will be multiple `renderxxxTables` in one `xxxTable`

WindSpiritSR:
@sh1luo Reformated an squashed

hsluoyz:
@Kininaru @sh1luo

hsluoyz:
@WindSpiritSR don't show Merge commits:

image

WindSpiritSR:
@hsluoyz fixed

hsluoyz:
@WindSpiritSR are all comments addressed?

WindSpiritSR:
@hsluoyz yep

object/user.go
hsluoyz: @WindSpiritSR why can't? See:

https://github.com/casbin/casdoor/blob/c632c3c30720479e37f5fff7e9efd49b5ac6dff0/object/user_util.go#L70-L90

object/organization_item.go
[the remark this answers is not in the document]
hsluoyz: @WindSpiritSR we should add this to show it's required:

image

Issue #201, opened 2021-07-26
Add country selection in both user profile and signup pages


hsluoyz:
1. User profile: https://door.casbin.com/account

![image](https://user-images.githubusercontent.com/3787410/127019664-3278a2a2-90f0-4b5d-bfe7-a3dbe2bdb053.png)

2. Signup page: 

![image](https://user-images.githubusercontent.com/3787410/127019873-4521b462-3046-498b-ac44-a8c7672f5a77.png)

We need to add all 200+ countries (better with their flag icons) like this:

![image](https://user-images.githubusercontent.com/3787410/127019970-04be39f7-4f9d-44a2-87f3-a053a93f01b8.png)

hsluoyz:
@sh1luo

sh1luo:
Doing.

PR #207, opened 2021-07-29
feat: add country/region select

sh1luo:
Fix: #201 


sh1luo:
Demo link: http://testb.rescld.cn/

PR #208, opened 2021-07-31
feat: add password Item forcibly when PasswordOn enabled

sh1luo:


sh1luo:
Demo link: http://1.13.13.178:8000/

PR #209, opened 2021-07-31
fix: sign up judgment logic error

sh1luo:


sh1luo:
This error will cause the page to be silent after `getaccount()` fails.

Demo link(error case): https://door.casbin.com/signup

hsluoyz:
Fixed by: https://github.com/casbin/casdoor/commit/e39c3b7613ee098cd54d7c0d3e51d6744a420053

PR #215, opened 2021-08-01
fix: add missing provider buttons

sh1luo:


sh1luo:
Demo link: http://testb.rescld.cn/

PR #216, opened 2021-08-01
fix: the time expired should be > time.Now().Unix()

driverzhang:
fix: i think check the time expired should be > time.Now().Unix()


hsluoyz:
@driverzhang why close?

PR #218, opened 2021-08-02
refactor: use ResponseError to response error

Kininaru:

Resolved: https://github.com/casbin/casdoor/issues/212

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -27,21 +27,13 @@ import (
 	"github.com/casbin/casdoor/util"
 )
 
-func codeToResponse(code *object.Code) *Response {
-	if code.Code == "" {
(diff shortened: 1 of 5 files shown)


controllers/auth.go
```diff
@@ -276,7 +258,7 @@ func (c *ApiController) Login() {
 				//	object.LinkMemberAccount(userId, "avatar", avatar)
 				//}
 
-				resp = c.HandleLoggedIn(application, user, &form)
+				c.HandleLoggedIn(application, user, &form)
```
hsluoyz, line 261: You didn't save the result, it will be overwritten by: `c.Data["json"] = Response{Status: "null"}`

hsluoyz:
This code is too buggy. Don't try to modify all places when it cannot be done. The OK part is so weird. I will find a day to modify it by myself.

Issue #223, opened 2021-08-05
need to show 404 page


sh1luo:
Casdoor should display 404 page when the user visits a path that does not exist.

But now, just blank, I think it's a bit unfriendly. The user may not know what is happening now causing the blank page, its own error or something else?

![image](https://user-images.githubusercontent.com/51157931/128286397-9857a068-3920-4890-8a01-b60431c03c4b.png)

hsluoyz:
@sh1luo good, plz do it.

sh1luo:
> @sh1luo good, plz do it.

ok, work on it.

PR #228, opened 2021-08-06
fix: add 404 page

sh1luo:
Fix: #223 


sh1luo:
Demo link: http://1.13.13.178:8000/foo/bar

After login with `admin`, `123`

Try to visit any path that does not exist again.

PR #270, opened 2021-08-17
fix: support server-side pagination for tokens

sh1luo:

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 
 	"github.com/casbin/casdoor/object"
+	"github.com/casbin/casdoor/util"
 )
 
 // GetTokens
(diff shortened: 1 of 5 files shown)


sh1luo:
Demo link: http://1.13.13.178:8000/tokens

web/src/TokenListPage.js
```diff
@@ -217,16 +223,27 @@ class TokenListPage extends React.Component {
       },
     ];
 
+    const paginationProps = {
+      current: this.state.page,
+      pageSize: this.state.pageSize,
+      total: this.state.total,
+      showQuickJumper: true,
+      showSizeChanger: true,
+      showTotal: () => `共${this.state.total}条`,
```
hsluoyz, line 232: i18n

web/src/TokenListPage.js
```diff
@@ -26,18 +26,24 @@ class TokenListPage extends React.Component {
     this.state = {
       classes: props,
       tokens: null,
+      total: 0,
+      page: 1,
+      pageSize: 5,
```
hsluoyz, line 31: one state is enough

controllers/token.go
```diff
@@ -28,8 +29,14 @@ import (
...
 	owner := c.Input().Get("owner")
+	limit := util.ParseInt(c.Input().Get("pageSize"))
+	page := util.ParseInt(c.Input().Get("page"))
+	if limit <= 0 || page < 1 {
+		c.ResponseError("Page or limit params error")
+		return
+	}
 
-	c.Data["json"] = object.GetTokens(owner)
+	c.ResponseOk(object.GetTokens(owner, limit, (page-1)*limit))
```
hsluoyz, line 39: ?

hsluoyz (changes requested):

controllers/token.go
sh1luo: Here need to return array of current page and total number of records. But `c.Data["json"]` only need one interface param.

web/src/TokenListPage.js
sh1luo: only save total? If that, there will be no loading effect.

sh1luo:
@hsluoyz Done, plz review.

PR #298, opened 2021-09-21
fix: signup logic when email and phone not enabled

chinggg:
`IsSignupItemEnabled` fails to do what it should do. It just check the existence of an item in `application.SignupItems`. 
But the `SignupItems` will always have all the items, even those items which are not visible or required. So I use `IsSignupItemVisible` instead.
In fact, an item is **actually enabled** only if it is visible, so `IsSignupItemEnabled` is of no use. Can we just remove it?

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -95,7 +95,7 @@ func (c *ApiController) Signup() {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


hsluoyz:
@chinggg good finding! Yes, please remove `IsSignupItemEnabled()`

PR #300, opened 2021-09-21
refactor: display resources and swagger menu  for normal users

nexmoe:
I think there have no use to display resources and swagger menu.
They should only upload thing through avatar uploading or something administers allow.
And swagger is for developer, normal users needn't see it.


hsluoyz:
Thanks!

Issue #314, opened 2021-11-08
Wrong token endpoint value caused "oauth2: server response missing access_token"


leopku:
I got "oauth2: server response missing access_token" while connecting `casdoor` using third-party oauth client.
But it's ok while changing `casdoor` to `Keycloak`.

After digging into source code, I found a wrong value of `TokenEndpoint`.
Changing `TokenEndpoint: fmt.Sprintf("%s/api/login", oidcOrigin)` to `TokenEndpoint: fmt.Sprintf("%s/api/login/oauth/access_token", oidcOrigin),` would make all thing works well.

PR #315, opened 2021-11-08
fix: Wrong token endpoint value #314

leopku:
Fix: https://github.com/casbin/casdoor/issues/314


hsluoyz:
@leopku hi, can you take a look at: https://github.com/casdoor/casdoor/issues/393 ?

PR #321, opened 2021-11-10
feat: translate the region select box

seriouszyx:
Fix: https://github.com/casbin/casdoor/issues/320

I used [i18n-iso-countries](https://github.com/michaelwittig/node-i18n-iso-countries). The results are shown below.

image

image

image

image

image

image

image

And our CDN lacks the UM icon, plz add it. The [country-flag-icons](https://catamphetamine.gitlab.io/country-flag-icons/3x2/index.html) has been provided.

image

image

Signed-off-by: “seriouszyx” <seriouszyx@foxmail.com>

PR #324, opened 2021-11-11
fix: save country code

seriouszyx:
With the update of #321, the data saved in different language interfaces are also in different languages. 
So I tend to save the country code and transfer it to the name when rendering.

image
image

Signed-off-by: “seriouszyx” <seriouszyx@foxmail.com>

### DIFF
--- a/web/src/UserListPage.js
+++ b/web/src/UserListPage.js
@@ -116,6 +116,13 @@ class UserListPage extends React.Component {
   }
 
   renderTable(users) {
+    // transfer country code to name based on selected language
+    var countries = require("i18n-iso-countries");
+    countries.registerLocale(require("i18n-iso-countries/langs/" + i18next.language + ".json"));
+    for (var index in users) {
+      users[index].region = countries.getName(users[index].region, i18next.language, {select: "official"})
+    }
+
     const columns = [
       {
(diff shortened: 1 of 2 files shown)


seriouszyx:
codes in database

image

web/src/UserListPage.js
```diff
@@ -116,6 +116,13 @@ class UserListPage extends React.Component {
   }
 
   renderTable(users) {
+    // transfer country code to name based on selected language
+    var countries = require("i18n-iso-countries");
```
hsluoyz, line 120: @seriouszyx can we use `import` instead of `require` ? Because `require` is usually Node.js's grammar. We better follow the ES6.

Same for the previous PR.

hsluoyz (changes requested):

web/src/UserListPage.js
seriouszyx: I have considered. But the next line requires a dynamically constructed path based on the selected language. It seems [ES6 doesn't support it](https://stackoverflow.com/questions/30340005/importing-modules-using-es6-syntax-and-dynamic-path).

hsluoyz: Good points!

PR #325, opened 2021-11-11
fix: panic while insert record when request uri too long

hyperzlib:
also filter "accessToken" in query.


hsluoyz:
lgtm

PR #326, opened 2021-11-11
feat: config jwt tokens in config file

hyperzlib:
Will load tokens from config item "jwtPrivateKeyFile" and "jwtPublicKeyFile".

If both files do not exist, it will generate new jwt key and save to these paths.


hsluoyz:
Replaced by the Certs page: https://door.casbin.com/certs

PR #327, opened 2021-11-11
fix: display "Continue with (User)" button of another organization

hyperzlib:
Display "Continue with (User)" button only if the currently logged-in user is a member of the client's corresponding organization.


hsluoyz:
lgtm

PR #329, opened 2021-11-12
feat: add config into .gitignore

hyperzlib:


hsluoyz:
@hyperzlib plz sign: https://cla-assistant.io/casdoor/casdoor

hsluoyz:
We don't keep a config example, as it only makes things more complicated. The `app.conf` is the only config file and can be customized by users.

PR #338, opened 2021-11-28
fix: add id_token and support auth header

ComradeProgrammer:
This PR contains 2 fixes: 
1.  add an 'id_token' field in access_token according to oidc protocol
2. supporting read client_id and client_secret from Basic Authorization Header, and return 400 instead of 200 when we cannot give out access_token.
See  <https://code.google.com/p/goauth2/issues/detail?id=31>  for why we are doing this.

This is because that despite Oauth 2.0 (rfc 6749, section 4.1.3)has mentioned that client_id and client_secret should be put into post form when using 'authorization code grant' method, (which is exactly the way we use in casdoor), however there do exist a contradictory expression in the same doc (rfc 6749, section 2.3.1) which is "The authorization server MUST support the HTTP Basic authentication scheme for authenticating clients that were issued a client password", indicating that it's ok to pass client_id and client_secret in Basic Authorization Header. For example, google uses the former way but github, reddit uses the latter way. 

Due to this confusion, it is SUGGESTED (not STIPULATED) that oauth client should try both methods, and distinguishes whether the current method successes by http response code.(for example, go oauth library try auth header first, and if non-2xx is returned, they try post form).This is the reason why we are adding 400 code into controller.

Also ,to avoid unnecessary failures, it's proper for us to support both methods, so we add support to read client_id and client_secret from Basic Authorization Header.

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -145,11 +145,28 @@ func (c *ApiController) GetOAuthCode() {
 // @Success 200 {object} object.TokenWrapper The Response object
 // @router /login/oauth/access_token [post]
 func (c *ApiController) GetOAuthToken() {
-	grantType := c.Input().Get("grant_type")
-	clientId := c.Input().Get("client_id")
-	clientSecret := c.Input().Get("client_secret")
-	code := c.Input().Get("code")
 
-	c.Data["json"] = object.GetOAuthToken(grantType, clientId, clientSecret, code)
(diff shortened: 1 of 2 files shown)


hsluoyz:
@ComradeProgrammer can you add the texts to our official docs? https://github.com/casdoor/casdoor-website

PR #348, opened 2021-12-05
fix: use env file instead of hard code

nodece:
Signed-off-by: Zixuan Liu <nodeces@gmail.com>

The env file is the most popular solution which can be used to define the dynamic configuration, so use the env file instead of hard code on the web.

The `process.env.XXX` can be replaced with a constant when building/developing.


nodece:
@hsluoyz could you take a look this?

rafaelgfirmino:
The license CLA is empty for me.
Do you know what i need to do?
image

hsluoyz:
The frontend code actually doesn't need to be configured, it natively supports nearly all scenarios. So I tend to make it less complicated.

PR #372, opened 2021-12-15
feat: support AuthnRequest in SAML

seriouszyx:

Fix: https://github.com/casbin/casdoor/issues/368

I added an option as below to determine whether to sign the request. If enabled, the certification should be uploaded to Keycloak Clients. I will update the documentation later.

image

However, there were some troubles during the development process. The gosaml2 lib seems to have some bugs about this point, and I have opened an [issue](https://github.com/russellhaering/gosaml2/issues/89). 

So if enable the option, we will use POST binding instead of Redirect binding. 

When tested with Aliyun IDaaS, the certification management function is VIP. And I can't find an option like `Client Signature Required` in Keycloak. So I suggest disable the option while using Aliyun IDaaS.

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -364,11 +364,12 @@ func (c *ApiController) Login() {
 
 func (c *ApiController) GetSamlLogin() {
 	providerId := c.Input().Get("id")
-	authURL, err := object.GenerateSamlLoginUrl(providerId)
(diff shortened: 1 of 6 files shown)


hsluoyz:
@Steve0x2a plz review

Steve0x2a (approved):
lgtm.

PR #375, opened 2021-12-16
fix: wrong sub return and docker-compose boolean value error.

Steve0x2a:
1. in get-account api, sub should reruen user's id instead of user's name.
2. according to [this document](https://docs.docker.com/compose/compose-file/compose-file-v3/#environment), any boolean values (true, false, yes, no) need to be enclosed in quotes to ensure they are not converted to True or False by the YML parser.


hsluoyz:
lgtm

PR #410, opened 2022-01-01
fix: refresh_token endpoint does not work

Steve0x2a:
1. add refresh_token to router and authz
2. according to [oidc doc](https://openid.net/specs/openid-connect-core-1_0.html#RefreshingAccessToken), adjust refresh_token endpoint return value
3. fix some refresh token bug
Signed-off-by: 0x2a <stevesough@gmail.com>

### DIFF
--- a/object/token.go
+++ b/object/token.go
@@ -17,7 +17,6 @@ package object
 import (
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/casbin/casdoor/util"
 	"xorm.io/core"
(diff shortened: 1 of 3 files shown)


object/token.go
```diff
@@ -302,57 +301,63 @@ func GetOAuthToken(grantType string, clientId string, clientSecret string, code
...
 	cert := getCertByApplication(application)
-	claims, err := ParseJwtToken(refreshToken, cert)
+	_, err = ParseJwtToken(refreshToken, cert)
 	if err != nil {
-		return &Code{
-			Message: "error: invalid refresh_token",
-			Code:    "",
-		}
-	}
-	if time.Now().Unix() > claims.ExpiresAt.Unix() {
```
Steve0x2a, line 344: ParseJwtToken will check if the token is expired

object/token.go
```diff
@@ -302,57 +301,63 @@ func GetOAuthToken(grantType string, clientId string, clientSecret string, code
...
-			Code:    "",
+		return &TokenWrapper{
+			AccessToken: fmt.Sprintf("error: %s", err.Error()),
+			TokenType:   "",
+			ExpiresIn:   0,
+			Scope:       "",
 		}
 	}
 	// generate a new token
-	user := getUser(application.Owner, token.User)
```
Steve0x2a, line 351: getUser should be passed in as the organization of the application and not the owner

hsluoyz:
@seriouszyx plz review

seriouszyx (approved):
LGTM

PR #419, opened 2022-01-13
fix: rebuild permission form

follisher:
1. Change `actions` and `resources` form to operable array
2. Change layout tag


hsluoyz:
@follisher plz fix CLA

PR #428, opened 2022-01-19
fix: Add the configuration of whether to print SQL. The default value is f…

hanpenggit:
fix: Add the configuration of whether to print SQL. The default value is false


hanpenggit:
fix: Add the configuration of whether to print SQL. The default value is false

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/429

Issue #439, opened 2022-01-22
SQL injection vulnerability in field filter


wuhan005:
The query API provides the `field` and `value` parameters to support query by field.

It inserts the user's input into the raw SQL expression which can lead to a SQL injection vulnerability.
```go
session = session.And(fmt.Sprintf("%s like ?", util.SnakeString(field)), fmt.Sprintf("%%%s%%", value))
```

And the `/api/get-organizations` is a public route that everyone can visit.

Proof of concept:
```
https://door.casbin.com/api/get-organizations?p=1&pageSize=10&value=e99nb&sortField=&sortOrder=&field=(updatexml(1,concat(0x7e,(select%20user()),0x7e),1))%20%20and%20name
```

hsluoyz:
@wuhan005 thanks for reporting! Do you have any suggestions for resolving it?

laeo:
Strict match given field is exactly one in the table, then using prepared statement to avoid SQL injection. (谷歌翻译的，将就着看吧

hsluoyz:
@seriouszyx can you fix this issue for all occurrences?

seriouszyx:
@hsluoyz I'll work on it.

wuhan005:
> @wuhan005 thanks for reporting! Do you have any suggestions for resolving it?

In my opinion, the input `field` should be filtered by an allowlist, and make sure we have set the Beego's run mode to `runmode=prod` in production to avoid leaking debug information.

wuhan005:
CVE-2022-24124 assigned.

PR #442, opened 2022-01-23
fix: fix the SQL injection vulnerability in field filter

seriouszyx:

Fix: https://github.com/casdoor/casdoor/issues/439

Solutions:
1. Use prepared statement
2. Implement the black list to filter the `field`

For the request as below:

```
http://localhost:8000/api/get-organizations?p=1&pageSize=10&value=e99nb&sortField=&sortOrder=&field=(updatexml(1,concat(0x7e,(select%20user()),0x7e),1))%20%20and%20name
```

Before:

image

After:

image

### DIFF
--- a/object/cert.go
+++ b/object/cert.go
@@ -53,10 +53,7 @@ func GetMaskedCerts(certs []*Cert) []*Cert {
 }
 
 func GetCertCount(owner, field, value string) int {
-	session := adapter.Engine.Where("owner=?", owner)
-	if field != "" && value != "" {
-		session = session.And(fmt.Sprintf("%s like ?", util.SnakeString(field)), fmt.Sprintf("%%%s%%", value))
-	}
+	session := GetSession(owner, -1, -1, field, value, "", "")
 	count, err := session.Count(&Cert{})
 	if err != nil {
 		panic(err)
--- a/object/check.go
+++ b/object/check.go
@@ -23,10 +23,14 @@ import (
 	goldap "github.com/go-ldap/ldap/v3"
 )
 
-var reWhiteSpace *regexp.Regexp
+var (
+	reWhiteSpace     *regexp.Regexp
+	reFieldWhiteList *regexp.Regexp
+)
 
 func init() {
 	reWhiteSpace, _ = regexp.Compile(`\s`)
+	reFieldWhiteList, _ = regexp.Compile(`^[A-Za-z0-9]+$`)
 }
 
 func CheckUserSignup(application *Application, organization *Organization, username string, password string, displayName string, email string, phone string, affiliation string) string {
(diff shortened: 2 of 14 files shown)


seriouszyx:
@wuhan005 Please review.

wuhan005:
It seems that the blacklist can be bypassed.

Boolean-based blind SQL injection can be used to get the value of the built-in function, the attacker can use functions like `substring` to make a brute-force attack and get each character of the string.

PoC:
```
(substring(user(),1,5) = 'hqcas') <> name
```
* Correct:
image
* Wrong: 
image

----
**I think we should use a whitelist instead of the blacklist.**

Maybe we can use `reflect.TypeOf()` to get a struct fields white list easily, here is an example: https://go.dev/play/p/84zUGphgvXR

seriouszyx:
@wuhan005 Thanks for your advice. I'll fix it.

seriouszyx:
@wuhan005 Please review the newest commit using whitelists.

object/cert.go
```diff
@@ -76,6 +83,9 @@ func GetCerts(owner string) []*Cert {
 }
 
 func GetPaginationCerts(owner string, offset, limit int, field, value, sortField, sortOrder string) []*Cert {
+	if field != "" && !certAllowList[field] {
+		panic(errors.New("abnormal field: " + field))
+	}
 	certs := []*Cert{}
```
hsluoyz, line 89: new line

object/check.go
```diff
@@ -179,3 +179,12 @@ func CheckUserPassword(organization string, username string, password string) (*
 
 	return user, ""
 }
+
+func InitAllowList(i interface{}) map[string]bool {
```
hsluoyz, line 183: init

object/check.go
```diff
@@ -179,3 +179,12 @@ func CheckUserPassword(organization string, username string, password string) (*
...
+	typ := reflect.TypeOf(i)
+	allowedList := make(map[string]bool)
+	for i := 0; i < typ.NumField(); i++ {
+		allowedList[typ.Field(i).Tag.Get("json")] = true
+	}
+	return allowedList
+}
```
hsluoyz, line 190: new line

hsluoyz (changes requested):

seriouszyx:
@wuhan005 Discussed with @hsluoyz, we think the previous implementation was a bit complicated. The `field` only consists of upper and lower case letters and numbers in Casdoor, so we can filter easily. Could you please review the latest commit to check the SQL injection vulnerability?

hsluoyz (approved):

wuhan005:
> @wuhan005 Discussed with @hsluoyz, we think the previous implementation was a bit complicated. The `field` only consists of upper and lower case letters and numbers in Casdoor, so we can filter easily. Could you please review the latest commit to check the SQL injection vulnerability?

There is no SQL injection vulnerability, nicely done.

Issue #445, opened 2022-01-24
No response at permission page


follisher:
When I use permission page, it is will be unresponsive and show a blank page, in browser console, we can find an error info
<img width="552" alt="image" src="https://user-images.githubusercontent.com/1535732/150742906-4217a7f8-0c9c-4494-a522-132d8c483df9.png">

hsluoyz:
@seriouszyx

seriouszyx:
Hi, did you deploy Casdoor by yourself or test on https://door.casbin.com/? I tested it several times locally but didn't encounter the same problem. Could you please provide more information such as the data in your databse if you deploy it by yourself?

follisher:
maybe you need testing in the initial project, when I testing in already exist project, i'm going to get nothing as well. It seems when project initial the function getTags(tags) of `Setting.js` file, the props `tags` have not value, so the method `forEach` don't know its a array object.
I fixed the issues in the pull request https://github.com/casdoor/casdoor/pull/449

PR #449, opened 2022-01-25
fix: fix permission page can't open

follisher:
Fix: https://github.com/casdoor/casdoor/issues/445

fixed in a initial project, when open premission page, the page will be crash

### DIFF
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -596,6 +596,7 @@ export function getTagColor(s) {
 
 export function getTags(tags) {
   let res = [];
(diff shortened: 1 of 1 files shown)


seriouszyx:
@follisher please fix

image

follisher:
@seriouszyx why the semantic pull request plugin always can't passed, or it is need another way to resolve it

hsluoyz:
@follisher not only PR title, but also all commit messages should be fixed:

image

seriouszyx (approved):

PR #473, opened 2022-02-05
ci: split the build workflow and fix push the docker image

nodece:
Signed-off-by: Zixuan Liu <nodeces@gmail.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push, pull_request]
+on: [ pull_request ]
 
 jobs:
   frontend:
@@ -13,7 +13,6 @@ jobs:
           node-version: '14.17.0'
       - run: yarn install && CI=false yarn run build
         working-directory: ./web
-
 
   backend:
     name: Back-end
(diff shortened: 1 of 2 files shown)


hsluoyz:
@ComradeProgrammer plz review

ComradeProgrammer (commented):
Well I do have some questions(or confusions).

0. What kind of bug have you encountered, which make you proposed this pr?
1. "if: github.repository == 'casdoor/casdoor' && github.event_name == 'push'"should be preserved I think, otherwise the CI would be triggered in your forked repository, lead to a failure in CI.
2. The original demand is that the "pushing image to remote should only be triggered when secondary version name changed", for example, the version number changing from 1.0.0 to 1.1.0 is supposed to trigger a push, however, its changing from 1.1.0 to 1.1.1 is not supposed to trigger a push. However, using semantic-release directly will case a push every time the version number is changed. Our resources on github and docker hub are limited so we cannot let so many versions consume them.

nodece:
1. The GitHub Action doesn't automatically run on the fork repo, see: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
2. We should always bump this software when the PR is merged into the master branch, and I feel that the current resources are sufficient on Github Action and DockerHub.

ComradeProgrammer:
> 1. The GitHub Action doesn't automatically run on the fork repo, see: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
> 2. We should always bump this software when the PR is merged into the master branch, and I feel that the current resources are sufficient on Github Action and DockerHub.

1. Developer may switch on the CI in order to run tests( although we don't have one yet ), so I think is useful to prohibit the pushing behavior from being proceeded under such circumstance
2. @hsluoyz plz decide whether the demand should be altered

nodece:
> Developer may switch on the CI in order to run tests( although we don't have one yet ), so I think is useful to prohibit the pushing behavior from being proceeded under such circumstance

I agreed with you.

hsluoyz:
> The original demand is that the "pushing image to remote should only be triggered when secondary version name changed", for example, the version number changing from 1.0.0 to 1.1.0 is supposed to trigger a push, however, its changing from 1.1.0 to 1.1.1 is not supposed to trigger a push. However, using semantic-release directly will case a push every time the version number is changed. Our resources on github and docker hub are limited so we cannot let so many versions consume them.

Yes. This is the result of discussion. We don't release for revision update.

PR #476, opened 2022-02-06
test: add step in gitHub action from running tests first

rafaelgfirmino:


object/syncer_user_test.go
```diff
@@ -1,38 +0,0 @@
-// Copyright 2021 The casbin Authors. All Rights Reserved.
```
hsluoyz, line 1: Can we avoid deleting these tests? Like skip these tests in CI?

hsluoyz (changes requested):

object/syncer_user_test.go
rafaelgfirmino: This test cam be removed. 
For this test running is necessary create a syncer first running this test, now is necessary create syncer manually. every person contribute with casdor need create a syncer for running test in your local machine.
This test don't have any assertion to, we can improve this point.
For not block te pipeline i decided remove this test, but I rewrite in the future.
What do you think ?

hsluoyz: @rafaelgfirmino these are more of the utility code (at least used by me) instead of "tests" supposed to be run in CI. They can be easily run in IDE like:

image

Is there a way to keep them in the code and just tell the CI to skip them?

rafaelgfirmino: I back file and add the flag skipCi and ignore this test in pipeline, I used the flag "-tags skipCi".

xlsx/xlsx_test.go
```diff
@@ -1,22 +0,0 @@
-// Copyright 2021 The casbin Authors. All Rights Reserved.
```
hsluoyz, line 1: Don't delete tests.

hsluoyz (changes requested):

hsluoyz (changes requested):

xlsx/xlsx_test.go
rafaelgfirmino: I will back wit this code but then don't run, this file  "../../tmpFiles/example" not exist for execute test.

hsluoyz: @rafaelgfirmino that's because that file only exists in my machine. I use it for test.

hsluoyz: Just don't delete any existing tests. They all exist for a reason

hsluoyz (approved):

PR #477, opened 2022-02-06
test: add tests in strings manipulation

rafaelgfirmino:
GetMinLenStr, this function was wrong, it did not always return the smallest number.
The smaller value is returned only its the first item in slice.
Can be tested with

```go
func TestGetMinLenStr(t *testing.T) {
	scenarios := []struct {
		description string
		input       []string
		expected    interface{}
	}{
		{"Should be return casbin", []string{"casdoor", "casbin"}, "casbin"},
		{"Should be return casbin", []string{"casdoor", "casbin", "casdoor_jdk"}, "casbin"},
		{"Should be return empty string", []string{"a", "", "casbin"}, ""},
		{"Should be return a", []string{"a", "casdoor", "casbin"}, "a"},//only test pass because the first works is the smaller
		{"Should be return a", []string{"casdoor", "a", "casbin"}, "a"},
		{"Should be return a", []string{"casbin", "casdoor", "a"}, "a"},
	}
	for _, scenery := range scenarios {
		t.Run(scenery.description, func(t *testing.T) {
			actual := GetMinLenStr(scenery.input...)
			assert.Equal(t, scenery.expected, actual, "The returned value not is expected")
		})
	}
	
```
CamelToSnakeCase and SnakeString form me is the same function. 
The both function not remove the spaces, example Lorem Ipsun result is lorem _ipsun I thinks this is not expected result

ReadBytesFromPath and WriteBytesToPath never is used, for me this functions not make sense exist in casdoor

### DIFF
--- a/util/string.go
+++ b/util/string.go
@@ -49,11 +49,11 @@ func ParseBool(s string) bool {
 func BoolToString(b bool) string {
 	if b {
 		return "1"
-	} else {
-		return "0"
 	}
+	return "0"
 }
 
+//CamelToSnakeCase This function transform camelcase in snakecase LoremIpsum in lorem_ipsum
 func CamelToSnakeCase(camel string) string {
 	var buf bytes.Buffer
 	for _, c := range camel {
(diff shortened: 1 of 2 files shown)


hsluoyz:
lgtm

PR #407, opened 2021-12-31
feat: support webauthn

ComradeProgrammer:
Fix: https://github.com/casdoor/casdoor/issues/399

feat: support webauthn 
demo: https://jm.casbin.com

PR #481, opened 2022-02-08
test: add tests in setting file

rafaelgfirmino:
This "GetUploadXlsx" function might return an invalid value depending on the parameter.

### DIFF
--- a/util/setting_test.go
+++ b/util/setting_test.go
@@ -1,0 +1,26 @@
+package util
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestGetUploadXlsxPath(t *testing.T) {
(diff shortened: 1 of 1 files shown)


util/setting_test.go
```diff
@@ -0,0 +1,26 @@
+package util
+
+import (
+	"github.com/stretchr/testify/assert"
```
nodece, line 4: Format this import.

rafaelgfirmino: I thinks this is a problem github, look these pull requests:
https://github.com/casdoor/casdoor/pull/407/files
https://github.com/casdoor/casdoor/pull/432/files

This format is a same

rafaelgfirmino: I didn't know about this convention.

PR #492, opened 2022-02-13
fix: fix cacert

ComradeProgrammer:
fix: fix cacert


ComradeProgrammer:
> @ComradeProgrammer the PR: #491 is already merged, is this PR stilled needed?

 very unlikely

PR #532, opened 2022-03-03
feat: add Oauth 2.0 Token Introspection(rfc7662) endpoint support

### LINKED ISSUE #425: Does it support Token Introspection api for resource server?
Does it support token introspection restful api for resource server?
Can't find related docs. search nothing.
支不支持Oauth2.0 Resource Server使用的token introspection接口？
没找到相关文档，用 Introspection 关键字在仓库内也没搜到相关内容。
refer: https://oauth.net/2/token-introspection/

LeonDevLifeLog:
fix #425 

Signed-off-by: Leon <leondevlifelog@gmail.com>

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -229,3 +229,65 @@ func (c *ApiController) TokenLogout() {
 	c.Data["json"] = wrapActionResponse(flag)
 	c.ServeJSON()
 }
+
+// IntrospectToken
+// @Title IntrospectToken
+// @Description The introspection endpoint is an OAuth 2.0 endpoint that takes a
+//  parameter representing an OAuth 2.0 token and returns a JSON document
+//  representing the meta information surrounding the
+//  token, including whether this token is currently active.
+//  This endpoint only support Basic Authorization.
+// @Param body body {object.TokenIntrospectionRequest} true "the request body"
+// @Success 200 {object} object.IntrospectionResponse The Response object
+// @router /login/oauth/introspect [post]
+func (c *ApiController) IntrospectToken() {
+	var body object.TokenIntrospectionRequest
+	err := json.Unmarshal(c.Ctx.Input.RequestBody, &body)
+	clientId, clientSecret, ok := c.Ctx.Request.BasicAuth()
+	if !ok {
+		util.LogWarning(c.Ctx, "Basic Authorization parses failed")
+		c.Data["json"] = Response{Status: "error", Msg: "Unauthorized operation"}
+		c.ServeJSON()
+		return
+	}
+	application := object.GetApplicationByClientId(clientId)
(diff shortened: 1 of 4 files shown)


hsluoyz:
@Steve0x2a plz review

controllers/token.go
```diff
@@ -229,3 +229,65 @@ func (c *ApiController) TokenLogout() {
...
+		util.LogWarning(c.Ctx, "application: %s can not find token", application.Name)
+		c.Data["json"] = &object.IntrospectionResponse{Active: false}
+		c.ServeJSON()
+		return
+	}
+	jwtToken, err := object.ParseJwtTokenByApplication(body.Token, application)
+	if err != nil || jwtToken.Valid() != nil {
+		// and token revoked case. but we not implement
+		// TODO: 2022-03-03 add token revoked check, when we implemented the Token Revocation(rfc7009) Specs.
+		// refs: https://tools.ietf.org/html/rfc7009
```
Steve0x2a, line 271: Here is to delete the token in the database based on the token? If it is, can use object.DeleteTokenByAceessToken.

Steve0x2a (commented):
The `authz.go` file may also need to be modified, otherwise this api may not be accessible. like:
`p, *, *, POST, /api/login/oauth/introspect`

controllers/token.go
LeonDevLifeLog: this is anathor specs. [OAuth 2.0 Token Revocation(rfc7009)](https://datatracker.ietf.org/doc/html/rfc7009).

we didn't implement it

LeonDevLifeLog: > this is anathor specs. [OAuth 2.0 Token Revocation(rfc7009)](https://datatracker.ietf.org/doc/html/rfc7009).
> 
> we didn't implement it

when we implement it, this place need to check the token has been revoked.

LeonDevLifeLog:
> The `authz.go` file may also need to be modified, otherwise this api may not be accessible. like: `p, *, *, POST, /api/login/oauth/introspect`

@Steve0x2a the endpoint needs protection.

## OAuth 2.0 Token Introspection (rfc7662) Page 5

> To prevent token scanning attacks, the endpoint MUST also require
>    some form of authorization to access this endpoint, such as client
>    authentication as described in OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)] or a separate
>    OAuth 2.0 access token such as the bearer token described in OAuth
>    2.0 Bearer Token Usage [[RFC6750](https://datatracker.ietf.org/doc/html/rfc6750)].  The methods of managing and
>    validating these authentication credentials are out of scope of this
>    specification.

Currently the endpoind is protected by Basic Authoritarian(ClientId:ClientSecret)

Steve0x2a (approved):

Issue #507, opened 2022-02-23
The application parameters in login url will be lost when back form Signup Page


FishMeng811:
Steps:
step 1: Redirect to Casdoor Sign In page, now parameters(ex. state/client_id) in Login Url.
step 2: Click Sign Up button into Sign Up page.
step 3: Then Click Sign In button back to Sgin In Page, now bug shows : parameters in Url lost!

hsluoyz:
@seriouszyx

PR #533, opened 2022-03-03
fix: fix missing parameters when signup

ComradeProgrammer:
fix: fix missing parameters when signup
FIX: #507


hsluoyz:
@FishMeng811 plz review

Issue #573, opened 2022-03-17
Callback URL bug


star7th:
Create a new application and prepare to use oauth2 to connect to casdoor. Fill in the callback address http://localhost:8080/server/?s=/api/extLogin/oauth2  。 The URL of the jump is like this http://localhost:8080/server/?s=/api/extLogin/oauth2?code=a004f0cd0d4110490d46&amp;state=3b09a6c8a3219ce55ba12a0e991cf1ec
This is wrong. When splicing code and state parameters, you directly use the symbol "?" To splice URLs, which will invalidate the "? S = / API / extlogin / oauth2" parameter in the callback address.
The correct way is to judge whether the callback address contains the symbol ". If so, use the symbol "&" to splice parameters

新建应用，准备使用oauth2来连接casdoor。回调地址处填写 http://localhost:8080/server/?s=/api/extLogin/oauth2 。跳转的url是这样的http://localhost:8080/server/?s=/api/extLogin/oauth2?code=a004f0cd0d4110490d46&state=3b09a6c8a3219ce55ba12a0e991cf1ec
这是错误的。你们在拼接code和state参数的时候，直接使用符号 "？" 来拼接url ，这会导致回调地址中的 "?s=/api/extLogin/oauth2" 参数失效。
正确的做法是，判断回调地址是否包含符号“？” 。如果是，则使用符号“&”来拼接参数

PR #583, opened 2022-03-19
fix: callback url param missing

seriouszyx:
Fix: #573 


seriouszyx:
@star7th plz review

PR #591, opened 2022-03-19
fix: fix missing +build

ComradeProgrammer:
fix: fix missing // +build
without this line of comment, code won't be able to be compiled with golang whose version is not greater than 1.16


hsluoyz:
lgtm, put to: https://github.com/casdoor/casdoor/commit/0841eb5c301be5937b2b3174c6da4a7f47c2c37d

Issue #581, opened 2022-03-18
Create a setting to make the endpoint /api/get-user private


hguerrerojaime:
Create a setting to make the endpoint /api/get-user private.

Have an option to make this endpoint restricted (that needs authentication in order to use it).

hsluoyz:
@seriouszyx

PR #593, opened 2022-03-19
feat: add auth require setting for accessing user info

MayankMittal1:
Fixes: #581 
added use auth for user api setting in organization page
Screenshot from 2022-03-20 02-31-18

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -87,6 +87,42 @@ func (c *ApiController) GetUser() {
 	id := c.Input().Get("id")
 	owner := c.Input().Get("owner")
 	email := c.Input().Get("email")
+	organization := object.GetOrganization(fmt.Sprintf("%s/%s", "admin", strings.Split(id, "/")[0]))
+
+	if organization.IsProfilePublic {
+		requestUserId := c.GetSessionUsername()
+		if requestUserId == "" {
+			c.ResponseError("Please login first.")
+			return
+		}
+		targetUser := object.GetUser(id)
+		if targetUser == nil {
+			c.ResponseError(fmt.Sprintf("The user: %s doesn't exist", id))
+			return
+		}
+
+		hasPermission := false
+		if strings.HasPrefix(requestUserId, "app/") {
+			hasPermission = true
+		} else {
+			requestUser := object.GetUser(requestUserId)
+			if requestUser == nil {
+				c.ResponseError("Session outdated. Please login again.")
+				return
+			}
+			if requestUser.IsGlobalAdmin {
+				hasPermission = true
+			} else if requestUserId == id {
+				hasPermission = true
+			} else if targetUser.Owner == requestUser.Owner && requestUser.IsAdmin {
+				hasPermission = true
+			}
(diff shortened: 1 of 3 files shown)


hsluoyz:
@Steve0x2a plz review

Steve0x2a:
@MayankMittal1 hi, /api/get-user and /api/get-users are two different apis, and https://github.com/casdoor/casdoor/issues/581 refers to /api/get-user.
Also, /api/get-users is protected by [authz](https://github.com/casdoor/casdoor/blob/master/authz/authz.go), I don't think /api/get-users needs another protection at the moment. Since only users of the built-in organization can use this api, and the built-in organization is seen as the global administrator, it seems unnecessary to control permissions for them.

I think it would be meaningful if it could be changed to protect /api/get-user, because currently /api/get-user only needs to know the organization name and username for anyone to get the user information.

MayankMittal1:
@Steve0x2a okay, will do the necessary changes

MayankMittal1:
@Steve0x2a merge plz

Steve0x2a (approved):
lgtm

MayankMittal1:
@hsluoyz merge please

object/organization.go
```diff
@@ -35,6 +35,7 @@ type Organization struct {
 	Tags               []string `xorm:"mediumtext" json:"tags"`
 	MasterPassword     string   `xorm:"varchar(100)" json:"masterPassword"`
 	EnableSoftDeletion bool     `json:"enableSoftDeletion"`
+	UseAuthForUserApi  bool     `json:"useAuthForUserApi"`
```
hsluoyz, line 38: Change to: `IsProfilePublic`

controllers/user.go
```diff
@@ -87,6 +87,41 @@ func (c *ApiController) GetUser() {
...
+		hasPermission := false
+		if strings.HasPrefix(requestUserId, "app/") {
+			hasPermission = true
+		} else {
+			requestUser := object.GetUser(requestUserId)
+			if requestUser == nil {
+				c.ResponseError("Session outdated. Please login again.")
+				return
+			}
+			if requestUser.IsGlobalAdmin {
```
hsluoyz, line 112: We better use Casbin authorization to control permission here

controllers/user.go
```diff
@@ -87,6 +87,41 @@ func (c *ApiController) GetUser() {
 	id := c.Input().Get("id")
 	owner := c.Input().Get("owner")
 	email := c.Input().Get("email")
+	organization := object.GetOrganization(fmt.Sprintf("%s/%s", "admin", strings.Split(id, "/")[0]))
```
hsluoyz, line 90: new line

hsluoyz (changes requested):

controllers/user.go
MayankMittal1: but we want to have this permission changable with every organisation, using casbin auth will require declaring at a global level, with no choice,
i have used similar approach as used in set-password api to verify permissions

Issue #584, opened 2022-03-19
Login limit problem


epis2048:
I entered the wrong user name and password in the login page and did not encounter the login times limit, which may be the script burst password.

I would like to have the option to lock the account for some minutes or enable captcha after exceeding the limit.

hsluoyz:
@uivid64

uivid64:
Is it better to limit the number of logins to 5 or 10? 
If the number of logins exceeds the limit, the account will be locked with the password and only the verification code will be allowed to log in.
Is this possible to meet this requirement?

T-baby:
I feel that it may be better to go through the CAPTCHA when logging in.

hsluoyz:
@epis2048 this has been finished, plz try it

![image](https://user-images.githubusercontent.com/3787410/185198925-73dc72a6-2431-4086-a6e2-fb8555d1c84e.png)

PR #600, opened 2022-03-21
feat: rate limit on login api

MayankMittal1:
Fixes: #584 
if simultaneous wrong login attempts exceed 5, backend stops checking creds for 5 minutes and gives 1 attempt once every 5 min until right creds are entered.
currently, time and number of attempts are hardcoded, but organizations could be given a choice to customize them.
Screenshot from 2022-03-21 17-27-06

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -207,10 +207,22 @@ func (c *ApiController) Login() {
 			}
 		} else {
 			password := form.Password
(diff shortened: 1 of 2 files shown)


hsluoyz:
@MayankMittal1 can you work on this issue instead? https://github.com/casdoor/casdoor/issues/604

PR #607, opened 2022-03-22
fix: add database configuration in docker runining

zcy530:
Upload readme.md


hsluoyz:
@ComradeProgrammer plz review

web/src/Conf.js
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export const ShowGithubCorner = false;
```
sagilio, line 15: These changes seem unrelated to this PR

main.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 func main() {
-	createDatabase := flag.Bool("createDatabase", false, "true if you need casdoor to create database")
+	createDatabase := flag.Bool("createDatabase", true, "true if you need casdoor to create database")
```
sagilio, line 32: Why not add this to conf or env？

README.md
```diff
@@ -179,8 +206,6 @@ And undeploy it with:
 make undeploy
 ```
 
-That's it! Try to visit http://localhost:8000/. :small_airplane:
```
sagilio, line 182: Why move this ?

README.md
```diff
@@ -160,11 +166,32 @@ dataSourceName = root:123456@tcp(db:3306)/
...
+
+
+
+Unlike the casdoor-all-in-one image, this image don't have a MySql database inside, which means you need to create a database named `casdoor`, before running docker.
+
+
+```bash
+
+# into database container
+docker exec -it casdoor_db_1 bash
```
sagilio, line 179: Container names may be different for other users, and I think most developers should know how to use the database, we need not descript this.

sagilio (changes requested):

README.md
zcy530: I think it is possible to visit http://localhost:8000/ before deployment,  so I moved it behind docker-compose up

README.md
zcy530: I am trying to solve the issue raised by this friend. And thanks for reminding me, I found the container with this name is inside the image, so I thought it was the same for everyone?
https://github.com/casbin/casnode/issues/445

sagilio: If the user does not edit the `docker-compose.yml` file, the name will indeed be this.
But if someone uses `docker stack` to run the `docker-compose.yml file, it may be another name.
I still think we only need to notify the user about the option of auto-create DB default is false.

README.md
sagilio: Ok, but description about k8s will lack this (:

hsluoyz:
This PR lacks a lot of things. I think it's not a good base for us to work on.

main.go
RyaoChengfeng: @sagilio Why not automatically create the database if it doesn't exist, but configure it here? I think the auto-created behavior would be better.

Issue #610, opened 2022-03-23
Panic when not select a speciafic provider


sagilio:
When the user adds a provider but does not select one, now the page looks like this:
![image](https://user-images.githubusercontent.com/42855245/159692522-b1674ee6-18eb-48aa-9c29-d31a5e85f0ea.png)

A nil check may be missing here, we can return an error response or ignore the nil provider item:
![image](https://user-images.githubusercontent.com/42855245/159692135-b896fe59-adba-4475-a6ca-4989f27bb543.png)

raw log:
```text
2022/03/23 19:47:34.438 [I]  (127.0.0.1: ) subOwner = built-in, subName = admin, method = GET, urlPath = /api/get-providers, obj.Owner = , obj.Name = , result = allow
2022/03/23 19:47:34.445 [C]  C:/Users/Sagilio/source/repos/OnWork/Casdoor/casdoor/object/provider_item.go:37
2022/03/23 19:47:34.449 [C]  C:/Users/Sagilio/source/repos/OnWork/Casdoor/casdoor/object/provider_item.go:41
2022/03/23 19:47:34.451 [C]  C:/Users/Sagilio/source/repos/OnWork/Casdoor/casdoor/object/application_item.go:88
2022/03/23 19:47:34.452 [C]  C:/Users/Sagilio/source/repos/OnWork/Casdoor/casdoor/object/application_item.go:105
2022/03/23 19:47:34.452 [C]  C:/Users/Sagilio/source/repos/OnWork/Casdoor/casdoor/controllers/auth.go:73
2022/03/23 19:47:34.454 [C]  C:/Users/Sagilio/source/repos/OnWork/Casdoor/casdoor/controllers/auth.go:412
2022/03/23 19:47:34.455 [C]  C:/Tools/Environment/go/go1.17.4/src/reflect/value.go:543
2022/03/23 19:47:34.456 [C]  C:/Tools/Environment/go/go1.17.4/src/reflect/value.go:339
2022/03/23 19:47:34.458 [C]  C:/Users/Sagilio/source/go/pkg/mod/github.com/astaxie/beego@v1.12.3/router.go:897
2022/03/23 19:47:34.461 [C]  C:/Tools/Environment/go/go1.17.4/src/net/http/server.go:2879
2022/03/23 19:47:34.462 [C]  C:/Tools/Environment/go/go1.17.4/src/net/http/server.go:1930
```

PR #614, opened 2022-03-23
fix: panic when not select one provider

sagilio:
Fixes: #610 

### Changes
ignore the nil provider item

### DIFF
--- a/object/provider_item.go
+++ b/object/provider_item.go
@@ -34,6 +34,9 @@ func (application *Application) GetProviderItem(providerName string) *ProviderIt
 }
 
 func (pi *ProviderItem) IsProviderVisible() bool {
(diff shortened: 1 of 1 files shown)


hsluoyz:
@sagilio why dont' change the frontend at the same time?

sagilio:
@hsluoyz Where does the front end need to be changed? 
I found two references:
1. This method will not return the provider that has not been set
https://github.com/casdoor/casdoor/blob/d0952ae908b63faea3c6938952668552ea8662f4/object/application_item.go#L104-L111
2. I think the response here is also in line with expectations：https://github.com/casdoor/casdoor/blob/d0952ae908b63faea3c6938952668552ea8662f4/controllers/auth.go#L239-L242

PR #627, opened 2022-03-28
fix: comparing hashed password with plain text password during password grant

mihado:
The code is currently comparing the plain text password provided during request with the hashed password of the user.

### DIFF
--- a/object/token.go
+++ b/object/token.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/astaxie/beego/logs"
 	"github.com/casdoor/casdoor/util"
 	"xorm.io/core"
 )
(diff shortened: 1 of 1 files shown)


object/token.go
```diff
@@ -522,7 +523,9 @@ func GetPasswordToken(application *Application, username string, password string
 	if user == nil {
 		return nil, errors.New("error: the user does not exist")
 	}
-	if user.Password != password {
+	msg := CheckPassword(user, password)
+	if msg != "" {
+		logs.Debug(fmt.Sprintf("GetPasswordToken error: %s", msg))
```
hsluoyz, line 528: no log

hsluoyz (changes requested):

object/token.go
```diff
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/astaxie/beego/logs"
```
mihado, line 25:
```suggestion
```

object/token.go
mihado, line 528:
```suggestion
```

mihado: @hsluoyz thanks. updated

hsluoyz (approved):

hsluoyz:
@seriouszyx plz review

seriouszyx (approved):

Issue #414, opened 2022-01-05
Support alipay authentication


qq690712575:
doc: [https://opendocs.alipay.com/apis/api_2/alipay.user.info.share](url)

MayankMittal1:
@hsluoyz i am not able to register for alibaba developer , as it requires email address from chinese providers, how should i proceed

hsluoyz:
@MayankMittal1 plz choose another issue to work on

PR #638, opened 2022-04-01
feat: add Alipay support as idp

halozhy:
Fix: #414
Some points need to be noted is below:
1. The `clientId` of casdoor corresponds to the app_id of Alipay, and the method to create an Alipay app can be found at https://opendocs.alipay.com/open/200/105310
2. The `clientSecret` of casdoor corresponds to the application private key you created on the Alipay open platform. In order to ensure security, Alipay needs us use the private key to generate a signature and send it with the request parameters every time we use its api. 
To geneate the application private key (`clientSecret`) , you can see https://opendocs.alipay.com/common/02kipl. Note that after the key pair is generated, it needs to be configured into the created application according to https://opendocs.alipay.com/common/02kdnc.
3. ~~Since the crypto package in golang does not support pkcs#8 format private keys, but Alipay recommends using this private key format, I use github.com/youmark/pkcs8 to process private keys (`clientSecret`)~~
The golang built-in package x509 has the API that I need. So the pkcs8 package is no longer used.
5. Since the private key that Alipay needs to save is too long, I changed the table creation statement in `object/provider.go` so that the `clientSecret` field can store longer strings

### DIFF
--- a/idp/alipay.go
+++ b/idp/alipay.go
@@ -1,0 +1,292 @@
+// Copyright 2022 The Casdoor Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package idp
+
+import (
+	"crypto"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"github.com/youmark/pkcs8"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"sort"
+	"strings"
+	"time"
+
+	"golang.org/x/oauth2"
+)
+
+type AlipayIdProvider struct {
+	Client *http.Client
+	Config *oauth2.Config
+}
+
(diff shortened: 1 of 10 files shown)


hsluoyz:
@halozhy can you provide a demo site?

hsluoyz:
@Steve0x2a plz review

halozhy:
> @halozhy can you provide a demo site?

http://1.117.101.146:7001/

Due to my low server bandwidth, the first open may be slow (30-60s)😥

idp/alipay.go
```diff
@@ -0,0 +1,292 @@
...
+
+import (
+	"crypto"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"github.com/youmark/pkcs8"
```
hsluoyz, line 25: sort

hsluoyz (changes requested):

idp/alipay.go
```diff
@@ -0,0 +1,292 @@
...
+	block, _ := pem.Decode([]byte(privateKey))
+	if block == nil {
+		panic("fail to parse privateKey")
+	}
+
+	h := sha256.New()
+	h.Write([]byte(signContent))
+	hashed := h.Sum(nil)
+
+	privateKeyRSA, err := pkcs8.ParsePKCS8PrivateKeyRSA(block.Bytes)
```
hsluoyz, line 254: Why don't use built-in API: `x509.ParsePKCS8PrivateKey` ?

hsluoyz (changes requested):

idp/alipay.go
halozhy: Sorry for not reading the documentation well, the pkcs8 package is no longer used in the latest commit [e02eec3](https://github.com/casdoor/casdoor/pull/638/commits/e02eec3bcddc602b18e8db7ec73aa9d119a3edec)

hsluoyz (approved):

PR #679, opened 2022-04-19
fix: Add footer

Nekotoxin:
See at: http://118.31.247.32:8000/


web/src/Conf.js
```diff
@@ -19,3 +19,6 @@ export const ForceLanguage = "";
 export const DefaultLanguage = "en";
 
 export const EnableExtraPages = true;
+
+export const ShowFooter = true;
+export const FooterText = "Powered by Casdoor";
```
hsluoyz, line 24: Still need to use application edit UI

PR #680, opened 2022-04-20
feat: support other idp property

akkuman:
support Adding third-party properties of other idp to user.Properties

Sometimes we want to add custom properties from third-party idp in Properties

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -265,10 +265,11 @@ func (c *ApiController) Login() {
 			return
 		}
 
-		userInfo := &idp.UserInfo{}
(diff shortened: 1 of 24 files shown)


controllers/auth.go
```diff
@@ -265,10 +265,11 @@ func (c *ApiController) Login() {
 			return
 		}
 
-		userInfo := &idp.UserInfo{}
+		userInfo := idp.UserInfoGetter(&idp.UserInfo{})
```
hsluoyz, line 268: As I said, `UserInfoGetter` is not an element way

hsluoyz (changes requested):

PR #690, opened 2022-04-21
fix: when req error, read body(nil) will panic

akkuman:
When an HTTP request error occurs
the response.Body will be nil

ioutil.ReadAll(nil) will panic


hsluoyz:
Hi @akkuman may I ask why you make PRs to the repo? Can you please join our QQ group?

PR #694, opened 2022-04-22
fix #693

### LINKED ISSUE #693: The refresh token don't read from the RequestBody
from qq group 645200447 user 735184182

bug here: https://github.com/casdoor/casdoor/blob/7b78e6026589bc2691eee0f2d34439b61ef4c4e1/controllers/token.go#L233

akkuman:
The content cannot be read from the RequestBody
when CopyRequestBody == false

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,6 +62,8 @@ func main() {
 		beego.BConfig.WebConfig.Session.SessionProviderConfig = conf.GetConfigString("redisEndpoint")
 	}
 	beego.BConfig.WebConfig.Session.SessionCookieLifeTime = 3600 * 24 * 30
+	// support multiple readings of Request.body
+	beego.BConfig.CopyRequestBody = true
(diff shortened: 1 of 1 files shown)


hsluoyz:
@akkuman we already set it:

https://github.com/casdoor/casdoor/blob/7b78e6026589bc2691eee0f2d34439b61ef4c4e1/conf/app.conf#L1-L7

akkuman:
but I think it should not display in app.conf, user could modify.
but when it is set to false, the related functionality may become unavailable.

hsluoyz:
@akkuman I don't think it's a big issue

akkuman:
I still stand by my opinion that this option should not be available to users in configuration files

akkuman:
However, I don't think it is a bug now, I just saw this problem in the QQ group, I searched the code, but I neglected the default configuration, it should not be a bug

I just don't think that configuration files should be used to control whether bugs occur or not, as that would confuse users

PR #707, opened 2022-04-25
fix: return right after error response on GetUserInfo

frankurcrazy:


hsluoyz:
@frankurcrazy good PR! Can you also tell me which bug this issue causes?

frankurcrazy:
Perhaps still getting status 200 even if GetUserInfo failed.

Issue #691, opened 2022-04-21
Need add 'use' and 'alg' in .well-known/jwks


rakuten:
I can't use casdoor log in to BookStackApp when using OIDC Mode

I found the need for 'alg' and 'use' in its code, But it's not in casdoor. How can I change it?
```
protected function filterKeys(array $keys): array
    {
        return array_filter($keys, function (array $key) {
            $alg = $key['alg'] ?? null;

            return $key['kty'] === 'RSA' && $key['use'] === 'sig' && (is_null($alg) || $alg === 'RS256');
        });
    }
```
[Link](https://github.com/BookStackApp/BookStack/blob/development/app/Auth/Access/Oidc/OidcProviderSettings.php) Line: 169

hsluoyz:
@halozhy

halozhy:
> I can't use casdoor log in to BookStackApp when using OIDC Mode
> 
> I found the need for 'alg' and 'use' in its code, But it's not in casdoor. How can I change it?
> 

You can add the following code after [object/oidc_discovery.go#L107](https://github.com/casdoor/casdoor/blob/master/object/oidc_discovery.go#L107)
```
jwk.Algorithm = "xxx"
jwk.Use = "xxx"
```

PR #708, opened 2022-04-25
fix: add 'use' and 'alg' in .well-known/jwks

halozhy:
According to #691 , now add the `alg` and `use` parameters to the response of `/.well-known/jwks`

Fix: https://github.com/casdoor/casdoor/issues/691

### DIFF
--- a/object/oidc_discovery.go
+++ b/object/oidc_discovery.go
@@ -105,7 +105,7 @@ func GetJsonWebKeySet() (jose.JSONWebKeySet, error) {
 		jwk.Key = x509Cert.PublicKey
 		jwk.Certificates = []*x509.Certificate{x509Cert}
 		jwk.KeyID = cert.Name
-		jwk.Algorithm = "RS256"
+		jwk.Algorithm = cert.CryptoAlgorithm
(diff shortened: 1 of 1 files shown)


rakuten (commented):
@halozhy Thank you very much for the change.

Using static values for these two variables can indeed solve my problem, but is it appropriate for casdoor?

hsluoyz:
@Steve0x2a is this a good fix?

Steve0x2a (approved):
lgtm

object/oidc_discovery.go
```diff
@@ -105,6 +105,8 @@ func GetJsonWebKeySet() (jose.JSONWebKeySet, error) {
 		jwk.Key = x509Cert.PublicKey
 		jwk.Certificates = []*x509.Certificate{x509Cert}
 		jwk.KeyID = cert.Name
+		jwk.Algorithm = "RS256"
```
Steve0x2a, line 108: `jwk.Algorithm = cert.CryptoAlgorithm`  might be better

rakuten (approved):

Issue #686, opened 2022-04-21
DingTalk provider value bug


akkuman:
https://github.com/casdoor/casdoor/blob/d3a7a062d3e71fb325c22ebcb0f8b4b43ccef610/object/user_util.go#L26-L42

example:
the provider.Type is DingTalk
but at `User`, it is 
```
DingTalk      string `xorm:"dingtalk varchar(100)" json:"dingtalk"\`
```

This can be a problem if the database field names are case sensitive

hsluoyz:
@ComradeProgrammer

PR #724, opened 2022-04-30
feat: DingTalk provider value case unsensitive

Resulte:
Fix: #686 
Screen Shot 2022-04-30 at 11 25 22


hsluoyz:
@Resulte 

1. fix: 
image

2. Use `feat:` instead of `fix:`

Resulte:
> @Resulte
> 
> 1. fix:
>    image
> 2. Use `feat:` instead of `fix:`

@hsluoyz all done

PR #726, opened 2022-04-30
fix: remove extra parentheses showing account numbers

zc2638:


zc2638:
WeChat02c92dfa4d7ea89a89e3f0c2bef64202

When `displayName` is not set, the extra brackets should not be displayed when logging in again

PR #727, opened 2022-04-30
fix: docker-compose up fails

greenhandatsjtu:
Hi there, I encountered `docker-compose up` failure when I try to use docker-compose to start casdoor server, indicating "Unknown database 'casdoor'".
Here are the logs:
image
This is because the server won't create `casdoor` database unless provided with command flag `--createDatabase=true`.
However, the Docker image `CMD` is `./server`, so that server won't create database, which results in `docker-compose up` failure.
https://github.com/casdoor/casdoor/blob/c05fb772241a0f58650a5ed731208303c76580d5/Dockerfile#L37
To fix this issue, we can set `entrypoint` of service casdoor as `/bin/sh -c './server --createDatabase=true'` to make the server create database first.
Here are `docker-compose up` logs after I fixed it:
image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     build:
       context: ./
       dockerfile: Dockerfile
+    entrypoint: /bin/sh -c './server --createDatabase=true'
     ports:
       - "8000:8000"
     depends_on:


hsluoyz:
@greenhandatsjtu did you read the docs? https://casdoor.org/docs/basic/try-with-docker#option-2-try-with-docker-compose

greenhandatsjtu:
> @greenhandatsjtu did you read the docs? [casdoor.org/docs/basic/try-with-docker#option-2-try-with-docker-compose](https://casdoor.org/docs/basic/try-with-docker#option-2-try-with-docker-compose)

@hsluoyz Yeah, I can see there's a toggle `RUNNING_IN_DOCKER`, and when `RUNNING_IN_DOCKER` is true, casdoor service can use db service, which is based on mysql image and doesn't have `casdoor` schema, that's why I'm adding `entrypoint` to casdoor service.

greenhandatsjtu:
@hsluoyz  Or, as `CreateDatabase()` has no effect (because it will create database only there's no target database):
https://github.com/casdoor/casdoor/blob/8e48bddf5f7c103c2975753c6050cf8ab2a8c291/object/adapter.go#L90
Shall we just remove `createDatabase` flag and call `CreateDatabase()` unconditionally?

hsluoyz:
@ComradeProgrammer plz review

ComradeProgrammer (approved):

Issue #713, opened 2022-04-27
java SDK updateUser interface cannot change the user password


zhuqiangqiang:

hsluoyz:
@zhuqiangqiang plz call this API: https://door.casdoor.com/swagger/#/Account%20API/ApiController.SetPassword

hsluoyz:
@haoliu97

haoliu97:
@zhuqiangqiang Since changing user's password is a high-risk opeartion, do it through updateUser interface is not a good idea. This PR https://github.com/casdoor/casdoor/pull/731 satisfies your requirement, but it won't be allowed. If you really want to do that, maybe you could just apply the change to your own's code.
Or you can add a setPassword method to the java SDK to call /api/set-password and make a pull request, this api requires old password of the user as a parameter, I don't know whether this is what you want.

PR #731, opened 2022-05-02
fix:java SDK updateUser interface can't change the user's password

haoliu97:
The fix have been passed a test.
image
image


haoliu97:
releated issue: https://github.com/casdoor/casdoor/issues/713

haoliu97:
Bad exper

PR #741, opened 2022-05-05
fix: origin error when not from localhost

fukco:
Fix: https://github.com/casdoor/casdoor/issues/700

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -281,8 +281,8 @@ func (c *ApiController) GetUserinfo() {
 		return
 	}
 	scope, aud := c.GetSessionOidc()
-	host := c.Ctx.Request.Host
-	resp, err := object.GetUserInfo(userId, scope, aud, host)
+	origin := fmt.Sprintf("%s://%s", c.Ctx.Input.Scheme(), c.Ctx.Request.Host)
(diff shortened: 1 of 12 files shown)


hsluoyz:
@fukco I don't suggest changing too much code than you need. It's much more than your part. And it's also breaking our original design like different frontend and backend ports. So closed here.

fukco:
My PR still compatible with your original design with two differen frontend and backend ports.

Sorry about to change too much code, I will submit a new PR only solve this issue.

hsluoyz:
@fukco plz don't PR for now. Actually I tent to not fix this one, as it may be your own env's issue.

So we will wait for other users' voice. If this is a common issue (> 5 people report it since now), then we will re-consider it.

fukco:
Okay, In my situation, I deploy it in a server, and develop in my PC or MAC.  If i use nonloopback address, OIDC discovery address always point to https.
I don't think it is a good design.

amoraitis:
I agree with @fukco, @hsluoyz non-production vpn-accessed environments do not usually have HTTPS support.

hsluoyz:
@amoraitis the code has changed a lot after 2 years. The original issue should have already been fixed. Don't append this old ticket. Create new issue if needed based on latest codebase.

PR #749, opened 2022-05-09
fix: unchanged masked client_secret/password updated to `***`

greenhandatsjtu:
For now, when editing `Organization`, master password is returned as the masked format `***`, so when clicking `Save` or `Save & Exit` without any change, master password will be updated to `***` or the hash value of `***`.

Similarly, when editing `Provider`, client secret and client secret 2 are both returned `***`, so when saving provider without any change, they will both be updated to `***`.

The solution is to check these columns before updating the record in DB, if they are `***`, then we should `Omit()` these fields.

### DIFF
--- a/object/application.go
+++ b/object/application.go
@@ -257,7 +257,11 @@ func UpdateApplication(id string, application *Application) bool {
 		providerItem.Provider = nil
 	}
 
-	affected, err := adapter.Engine.ID(core.PK{owner, name}).AllCols().Update(application)
+	session := adapter.Engine.ID(core.PK{owner, name}).AllCols()
+	if application.ClientSecret == "***" {
+		session.Omit("client_secret")
+	}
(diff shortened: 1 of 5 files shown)


zc2638:
This fix will cause an error if the user sets the password to `***`

zc2638:
The backend returns a hasPassword field?
If the frontend is judged to be `true`, it will display `***`.
This way might be better, what do you think?

greenhandatsjtu:
> This fix will cause an error if the user sets the password to `***`

> The backend returns a `hasPassword` field? If the frontend is judged to be `true`, it will display `***`. This way might be better, what do you think?

Yes, for `client secret` and `client secret 2`, it's OK (because they would never be `***`), for password, user can't set password as `***` anymore. But @hsluoyz said it's fine to just check if the column is `***`, if so then omit the column when updating.

PR #763, opened 2022-05-29
fix: use openid or unionid as username rather than nickname when logging with WeChat

### LINKED ISSUE #762: The problem of the same username when logging in by scanning the QR code on WeChat
当新的用户使用微信扫码登陆时，微信只能够提供用户的昵称、头像、unionID 等信息，但是 casdoor 会自动将昵称当作用户名传值给 name 字段，而实际上微信提供的昵称并不是唯一值，它仅能够当作显示名（ displayName 字段）来使用，这可能会使新用户在登陆时因昵称相同而无法注册的情况发生。因此 casdoor 能否调整一下与微信字段的对应关系，也就是当新用户使用微信登陆时，自动生成随机唯一的 name 值。
When a new user uses WeChat to scan the code to log in, WeChat can only provide the user's nickname, avatar, unionID and other information, but Casdoor will automatically pass the nickname as the username to the name field, but in fact the nickname provided by WeChat is not Unique value, it can only be used as a displayName, which may prevent a new user from being able to register due to the same nickname when logging in. Therefore, can casdoor adjust the corresponding relationship with the WeChat field, that is, when a new user logs in using WeChat, a random and unique name value is automatically generated.

caoshengdong:
fix: use openid or unionid as username rather than nickname when logging with WeChat
FIX #762

### DIFF
--- a/idp/wechat.go
+++ b/idp/wechat.go
@@ -185,7 +185,7 @@ func (idp *WeChatIdProvider) GetUserInfo(token *oauth2.Token) (*UserInfo, error)
 
 	userInfo := UserInfo{
 		Id:          id,
-		Username:    wechatUserInfo.Nickname,
+		Username:    id,
 		DisplayName: wechatUserInfo.Nickname,
 		AvatarUrl:   wechatUserInfo.Headimgurl,
 	}


hsluoyz:
@caoshengdong plz join the QQ group: 482622336 for discussion

Issue #705, opened 2022-04-25
Show alert when user clicks on application edit page's preview window


nomeguy:
See: https://demo.casdoor.com/applications/app-built-in

![image](https://user-images.githubusercontent.com/85475922/165013963-8837eed3-c49a-4fd9-b004-01763c3c306a.png)

hsluoyz:
@halozhy

PR #791, opened 2022-06-11
feat: show alert when user clicks on application edit page's preview wi…

ziliangyu:
Add a mask layer, cannot click

Fix: https://github.com/casdoor/casdoor/issues/705


ziliangyu:
Add a mask layer through positioning, which cannot be clicked.
T F@L6GTYZ8CSP_IIT7(QX8

PR #793, opened 2022-06-12
fix: show alert when user clicks on application edit page's preview window

ziliangyu:
Add a mask layer through positioning, which cannot be clicked.
9_(1KYUZ3SEF9XMPTC{T@}J


hsluoyz:
Don't re-create PRs

Issue #734, opened 2022-05-03
How are users' application permissions authorized?


zdsiyan:
For example, user X has access to application A, application B, application C; user Y has access to application A, application B. I have written the distinction to properties, is there any other solution？

PR #806, opened 2022-06-20
feat: implement access control using casbin

seriouszyx:
Fix: https://github.com/casdoor/casdoor/issues/734


seriouszyx:
demo link: http://101.43.193.167:8000/

user `admin/123` can login, user `zyx/123` cannot.

two permissions:

image

model supports ACL currently:

image

PR #682, opened 2022-04-20
fix: fix failure of introspection

ComradeProgrammer:
Fix: https://github.com/casdoor/casdoor/issues/696

PR #811, opened 2022-06-21
fix: optimize api matching rules

aecra:
Revert: https://github.com/casdoor/casdoor/pull/682


hsluoyz:
Not the correct way to fix, see: https://github.com/casdoor/casdoor/pull/813

PR #813, opened 2022-06-21
fix: fix method for oauth

ComradeProgrammer:
fix: fix method for oauth


hsluoyz:
@Steve0x2a @seriouszyx is this method good? Plz review

aecra:
This is not a good way to fix oauth.

For example, access_token should only accept POST request. For OPTION request, there should be a signal func to return cors response. It is bad to execute POST func for OPTION.

And it is better to add an domain field to every application to verify OPTION request.

hsluoyz:
@aecra can you make a PR?

aecra:
@hsluoyz I can solve next week.

aecra:
#826

Issue #816, opened 2022-06-22
Sha256 encryption without slat, verification failed


704237006:
My system uses sha256 to encrypt the password without salt. However, after configuring the password encryption type in the casdoor without salt, the verification is incorrect.

![image](https://user-images.githubusercontent.com/42430068/174929640-a4f31300-974a-49e8-8a31-866e5e6aa87d.png)

PR #821, opened 2022-06-22
fix: encryption without salt

seriouszyx:
Fix: https://github.com/casdoor/casdoor/issues/816

Signed-off-by: Yixiang Zhao <seriouszyx@foxmail.com>

PR #822, opened 2022-06-22
fix without salt but getSha256HexDigest twice bug

704237006:
https://github.com/casdoor/casdoor/issues/816


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/821

PR #825, opened 2022-06-23
fix: fix #784

### LINKED ISSUE #784: Signup page not loading and Social Signup buttons now showing on the signup page
- Using the  `casbin/casdoor:latest` docker image, fresh install. When social providers are added to the application, the sign-in page is showing the social login options, but the signup page is not (please refer below screenshot)

image

- Also when the "test signup page" button is clicked or the signup url of the application is pasted in the browser, it is redirecting to the casdoor home instead of showing the signup page

- On the Signin page the signup link button is also taking to the casdoor home instead of the signup page of the application. 
image

All values are in their default, no special changes have been made. Wondering what is the issue and how to make the signup page load correctly.
image

qianxi0410:


web/src/auth/SignupPage.js
```diff
@@ -499,6 +500,40 @@ class SignupPage extends React.Component {
     )
   }
 
+  renderProviderLogo(provider, application, width, margin, size) {
```
hsluoyz, line 503: Don't duplicate the code, put the shared code into Setting.js

web/src/auth/SignupPage.js
```diff
@@ -572,6 +607,13 @@ class SignupPage extends React.Component {
             {i18next.t("signup:sign in now")}
           </a>
         </Form.Item>
+        <Form.Item>
+          {
+            application.providers.filter(providerItem => this.isProviderVisible(providerItem)).map(providerItem => {
+              return this.renderProviderLogo(providerItem.provider, application, 30, 5, "small");
+            })
+          }
+        </Form.Item>
```
hsluoyz, line 616: Did you test the code if it works? Provide a demo video

hsluoyz (changes requested):

web/src/auth/SignupPage.js
qianxi0410: in application page:
图片

in test page:

图片

web/src/auth/SignupPage.js
qianxi0410: In fact, the redner function here does not need to use size, and the logic of processing saml is not clear now.

web/src/auth/SignupPage.js
hsluoyz: @qianxi0410 you need to center it

image

web/src/auth/SignupPage.js
hsluoyz: @qianxi0410 that's another issue. If you think it's an issue and want to fix it, you need to make a new PR

PR #838, opened 2022-06-30
fix: fix CORS error after sucessful OPTION

aecra:
Cross-origin requests other than preflight requests also need to add CORS response headers.


hsluoyz:
@aecra can you explain more? In what scenario it will have other CORS requests?

aecra:
Any cross-origin request would need a preflight request to confirem that this request is accepted. And the formal request also needs to add CORS response headers, otherwise the browser will not receive the response body.

PR #862, opened 2022-07-05
fix: support user-defined clientId&Secret

Abingcbc:
`AddApplication` only supports randomly generated clientId and clientSecret. So I copy its logic and support keeping user-defined clientId and clientSecret when initializing.

### DIFF
--- a/init_data.json.template
+++ b/init_data.json.template
@@ -23,6 +23,8 @@
       "cert": "",
       "enablePassword": true,
       "enableSignUp": true,
+      "clientId": "",
+      "clientSecret": "",
       "providers": [
         {
           "name": "",
--- a/object/application.go
+++ b/object/application.go
@@ -280,8 +280,12 @@ func UpdateApplication(id string, application *Application) bool {
 }
 
 func AddApplication(application *Application) bool {
-	application.ClientId = util.GenerateClientId()
-	application.ClientSecret = util.GenerateClientSecret()
+	if application.ClientId == "" {
+		application.ClientId = util.GenerateClientId()
+	}
+	if application.ClientSecret == "" {
+		application.ClientSecret = util.GenerateClientSecret()
+	}
 	for _, providerItem := range application.Providers {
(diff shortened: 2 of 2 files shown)


object/init_data.go
```diff
@@ -106,7 +106,19 @@ func initDefinedApplication(application *Application) {
 		return
 	}
 	application.CreatedTime = util.GetCurrentTime()
-	AddApplication(application)
+	// If user not defined clientId and clientSecret
+	if len(application.ClientId) == 0 || len(application.ClientSecret) == 0 {
```
hsluoyz, line 110: Change to `s == ""`

object/init_data.go
```diff
@@ -106,7 +106,19 @@ func initDefinedApplication(application *Application) {
...
+		application.ClientId = util.GenerateClientId()
+		application.ClientSecret = util.GenerateClientSecret()
+	}
+	for _, providerItem := range application.Providers {
+		providerItem.Provider = nil
```
hsluoyz, line 115: Why set to `nil` exlicitly, isn't it `nil` by default?

object/init_data.go
```diff
@@ -106,7 +106,19 @@ func initDefinedApplication(application *Application) {
...
+	}
+
+	_, err := adapter.Engine.Insert(application)
```
hsluoyz, line 118: Why not use `AddApplication(application)`?

hsluoyz (changes requested):

object/init_data.go
Abingcbc: It is the same as `AddApplication`
https://github.com/casdoor/casdoor/blob/ec296215471b1e2e35add3826f07ec5bd8165970/object/application.go#L282-L295

object/init_data.go
Abingcbc: `AddApplication` only supports randomly generated clientId and clientSecret.

hsluoyz (approved):

PR #831, opened 2022-06-27
fix: some minor bugs and make Dockerfile more productive.

atompi:
### Make Dockerfile more productive.

+ Build all-in-one image:

```
docker build --target ALLINONE -t casbin/casdoor-all-in-one:latest .
```

+ Build standard image:

```
docker build --target STANDARD -t casbin/casdoor:latest .
```

### Solve the problem that the PermissionList page jumped abnormally when clicking the permission name in the list.

+ Insert the owner into the url when jumping to the permissions details page.

### Fix incorrect SignupApplication field value in initBuiltInUser.

+ Wrong SignupApplication value causes exceptions in multiple functions for the first admin user, such as uploading avatars, user detail pages, etc.

PR #871, opened 2022-07-09
feat: fix oauth unknown authority in docker

Abingcbc:
This has been fixed in PR: https://github.com/casdoor/casdoor/pull/491 by adding `ca-certificate`. But PR https://github.com/casdoor/casdoor/pull/831 removed these as 

> There are many unnecessary shell commands in the old Dockerfile, such as install ca-certificates && update-ca-certificates. When we use debian:lates or alpine:latest as the base image, /etc/ssl/certs/ca-certificates.crt already contains the latest ca certificates, and we don't need to perform an update operation every time we build.

 However, when I test with Github OAuth, `Post "https://github.com/login/oauth/access_token": X509 certificate signed by unknown authority` still raised. I guess default certificate in debain system is out of date. After I added `ca-certificate` again, it works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,10 @@ RUN ./build.sh
 FROM alpine:latest AS STANDARD
 LABEL MAINTAINER="https://casdoor.org/"
 
+RUN sed -i 's/https/http/' /etc/apk/repositories
+RUN apk add curl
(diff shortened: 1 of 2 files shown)


Abingcbc:
@ComradeProgrammer @hsluoyz PTAL

Issue #850, opened 2022-07-03
Add frontend JS code linter


hsluoyz:
Right now we manually control the code style quality for JS code at: https://github.com/casdoor/casdoor/tree/master/web/src , but it's very tired for code review.

We need to setup a JS linter to `fully comply with our current code style`. It perhaps needs to tune the linter parameters to be close to what style we have now as much as possible.

Like:

https://github.com/casdoor/casdoor/blob/f37fd6ba87ce24cf5948b9a0ccd80647d21ac9d1/web/src/UserEditPage.js#L522-L574

hsluoyz:
@leo220yuyaodog

PR #872, opened 2022-07-09
feat(web): add lint and fix some lint error

qianxi0410:
Close: #850 

Added minimal lint checks for react projects.
Follow Antd's style design:

1. use semicolon
2. use double quotes for JSX attributes and single quotes for the rest
3. spacing is two space characters
4. A small part of code quality inspection has been added, but it can only be set to warn level at present, otherwise it will cause too many errors, it can be set to error level later, and a new and stricter lint plugin will be introduced


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/875

Issue #706, opened 2022-04-25
Unauthorized access to page is not well displayed


nomeguy:
When logged in with a normal user, but access admin pages like org list, you will see:

![image](https://user-images.githubusercontent.com/85475922/165020253-531c427b-b1df-4eec-adf1-713af8f8e194.png)

hsluoyz:
@Xeonacid

hsluoyz:
@qianxi0410

PR #891, opened 2022-07-17
fix(web): fix the bug of infinity loop animate  when unauthorized

qianxi0410:
Close: #706


qianxi0410:
I took a cursory look, there are many list pages that need to have such a check, there are a lot of repeated codes, but my front-end technology is not good enough, I don't know how to abstract it into a public function class, this pr solves the problem A small part of these problems is provided, and a template is provided for the following pr.
@hsluoyz plz r.

web/src/OrganizationListPage.js
```diff
@@ -271,7 +275,13 @@ class OrganizationListPage extends BaseListPage {
             },
             searchText: params.searchText,
             searchedColumn: params.searchedColumn,
+            isAuthenticated: true,
```
hsluoyz, line 278: isAuthenticated -> isAuthorizied

qianxi0410:
PTAL @hsluoyz

Issue #884, opened 2022-07-15
Wrong usage of getUserById  


waltcow:
When I review code of syncer_user logic, I found something maybe wrong

https://github.com/casdoor/casdoor/blob/895cdd024d52e29ea9df61a0b6ea4219c6e44ac4/object/syncer_user.go#L115-L134

https://github.com/casdoor/casdoor/blob/801302c6e72f8cf82ed563a92694c6ce329456cd/util/string.go#L87-L95

```go
owner, name := util.GetOwnerAndNameFromId(user.GetId()) 
getUserById(owner, name)  
```
when we call `getUserById` method, it should pass through user owner and id rather than owner and name 

https://github.com/casdoor/casdoor/blob/895cdd024d52e29ea9df61a0b6ea4219c6e44ac4/object/user.go#L220-L237

hsluoyz:
@waltcow can you explain a little bit more?

waltcow:
@hsluoyz I have just edit this issue description

hsluoyz:
Actually this is not a bug, won't fix

waltcow:
so misleading code

PR #920, opened 2022-07-27
fix: wrong pass through when getUserById()

imp2002:
Fix: https://github.com/casdoor/casdoor/issues/884

passed over should be id?

image


hsluoyz:
@imp2002 actually this is not a bug. Plz discuss with me before writing code to save the efforts :)

Issue #898, opened 2022-07-19
users are not inserted into the table casdoor.user when sync users from keycloak to casdoor with keycloak syncer


billyhu:
I have launched keycloak "quay.io/keycloak/keycloak:legacy", postgres and casdoor "casbin/casdoor-all-in-one" with docker image, and tried to sync users with keycloak syncer in casdoor.
I followed this doc to config keycloak syncer - https://casdoor.org/docs/syncer/Keycloak#keycloak-syncer,
When I tried to sync users with the latest image of casdoor, there's no error message now, while it seems users are still not inserted into the table casdoor.user, so we cannot see users list on the page. could you please have a check on this issue?

hsluoyz:
@seriouszyx

seriouszyx:
@billyhu I double-checked that the users can be synced in my environment. Could you provide the logs after clicking the `sync` button? My logs are shown below. 

![image](https://user-images.githubusercontent.com/33992371/180702593-7c1a7a25-6be8-4005-8b28-a1064c4c2b40.png)

postgres and keycloak container:

![image](https://user-images.githubusercontent.com/33992371/180702788-6a56cbca-5b9a-4bba-9a3a-5b435a5375cf.png)

casdoor syncer configuration:

![image](https://user-images.githubusercontent.com/33992371/180702854-31887d8e-8541-4ea8-9716-9c60e22e58de.png)

synced users:

![image](https://user-images.githubusercontent.com/33992371/180702975-6e5527f2-81f2-4e49-9587-04c85f2d18b2.png)

billyhu:
Thanks @seriouszyx . I have tried it again and confirmed that users can be synced if users name is not conflict between kc and casdoor, while users with the same name will block the whole syncing batch and no one is synced.

hsluoyz:
@billyhu what do you mean by "username conflict"? Can you explain more about it?

billyhu:
Sure, @hsluoyz . For example, if there's a user with name 'test' in keycloak and casdoor has a existing user with the same name 'test', the casdoor syncer cannot sync users.

hsluoyz:
@billyhu we should skip the duplicated users. Can you make a PR?

billyhu:
Thanks, @hsluoyz . I'm not sure what the 'PR' means?

hsluoyz:
@billyhu ok,

@seriouszyx plz work on this

billyhu:
Thanks @seriouszyx and @hsluoyz .

PR #928, opened 2022-07-30
fix: skip the duplicated users when sync users

seriouszyx:
Fix: https://github.com/casdoor/casdoor/issues/898


billyhu:
Thanks all for your quick fixing!

Issue #913, opened 2022-07-26
Need an API interface to retrieve user roles through user queries


Zerek-Cheng:

hsluoyz:
@Resulte

PR #929, opened 2022-07-30
feat: get user api return roles and permissions

Resulte:
Fix: https://github.com/casdoor/casdoor/issues/913

roles_permisions

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -112,6 +112,11 @@ func (c *ApiController) GetUser() {
 		user = object.GetUser(id)
 	}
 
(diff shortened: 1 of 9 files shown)


controllers/user.go
```diff
@@ -112,6 +112,11 @@ func (c *ApiController) GetUser() {
 		user = object.GetUser(id)
 	}
 
+	roles := object.GetRolesByUser(fmt.Sprintf("%s/%s", user.Owner, user.Name))
```
hsluoyz, line 115: `user.GetId()` is better than this

hsluoyz (changes requested):

hsluoyz (approved):

Issue #917, opened 2022-07-26
Can not use SMTP provider with MYSUBMAIL.COM at 465 port


Zerek-Cheng:
the error:
> tls: first record does not look like a TLS handshake

[the docs](https://mysubmail.com/documents/J2pHa)

FoeverA0:
![QQ图片20220730105148](https://user-images.githubusercontent.com/92803447/181867567-3dd54cdf-156e-4158-a993-d7b9d442aca3.png)
@Zerek-Cheng I found the following information in the connection you provided, it looks like you need to use the SMTP provider on port 25? hope this helps you : )

Zerek-Cheng:
@FoeverA0 
I have consulted the API provider, this interface supports 465 port requests. My hosting provider doesn't support port 25.

hsluoyz:
@Zerek-Cheng does this change work for you?

![image](https://user-images.githubusercontent.com/3787410/181905220-37955f4a-9590-4ade-b2ba-f0f509dbae15.png)

FoeverA0:
#931

FoeverA0:
Fix: https://github.com/casdoor/casdoor/issues/917

Zerek-Cheng:
> @Zerek-Cheng这个改变对你有用吗？
> 
> ![图片](https://user-images.githubusercontent.com/3787410/181905220-37955f4a-9590-4ade-b2ba-f0f509dbae15.png)

yea，thks

PR #935, opened 2022-07-30
fix: Can not use SMTP provider with MYSUBMAIL.COM at 465 port

FoeverA0:
Added an enable ssl switch.


FoeverA0:
fix: #917

Zerek-Cheng:
@FoeverA0 something wrong with your pq

PR #936, opened 2022-07-30
fix: unable to get user if profile is private

dirname:
unable to get the user by email or phone if the profile is not public.

### DIFF
--- a/object/check.go
+++ b/object/check.go
@@ -197,14 +197,18 @@ func filterField(field string) bool {
 	return reFieldWhiteList.MatchString(field)
 }
 
-func CheckUserPermission(requestUserId, userId string, strict bool) (bool, error) {
+func CheckUserPermission(requestUserId, userId, userOwner string, strict bool) (bool, error) {
 	if requestUserId == "" {
 		return false, fmt.Errorf("please login first")
 	}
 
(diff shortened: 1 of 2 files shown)


object/check.go
```diff
@@ -197,14 +197,18 @@ func filterField(field string) bool {
...
-func CheckUserPermission(requestUserId, userId string, strict bool) (bool, error) {
+func CheckUserPermission(requestUserId, userId, userOwner string, strict bool) (bool, error) {
 	if requestUserId == "" {
 		return false, fmt.Errorf("please login first")
 	}
 
-	targetUser := GetUser(userId)
-	if targetUser == nil {
-		return false, fmt.Errorf("the user: %s doesn't exist", userId)
+	if userId != "" {
```
dirname, line 205: When getting user by email or mobile phone, the `id` may be `""` at this time, which will trigger panic at [util/string.go:92](https://github.com/casdoor/casdoor/blob/master/util/string.go#L92)

dirname, line 205: When getting user by email or mobile phone, the `id` may be `""` at this time, which will trigger panic at [util/string.go:92](https://github.com/casdoor/casdoor/blob/master/util/string.go#L92)

Issue #946, opened 2022-08-01
login page css error


AubreySte:
![image](https://user-images.githubusercontent.com/75111676/182180175-cd4b71ad-d508-4e04-b526-1d9c23fa3a10.png)
When there is already an account logged in, the bottom is blocking the login button

AubreySte:
source code lastest

PR #951, opened 2022-08-03
fix: fixed footer css

Abingcbc:
Fix: https://github.com/casdoor/casdoor/issues/946

https://user-images.githubusercontent.com/34068205/183283533-bd0f57e5-bd76-4ee2-ad21-da92d0d750b4.mov

### DIFF
--- a/web/src/App.js
+++ b/web/src/App.js
@@ -668,7 +668,7 @@ class App extends Component {
   renderPage() {
     if (this.isDoorPages()) {
       return (
-        <div>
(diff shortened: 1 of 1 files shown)


hsluoyz:
@Abingcbc but we don't want the footer to be sticky, not only it takes a lot of space, but also it causes bugs like:

WFRV`0R_Y}1N9N@$(TWGH$Q

This bug also showed up in the above video (but very quickly):

image

Issue #846, opened 2022-07-02
Feature Request: Expose the casbin API to frontend for current user


KrishnaPG:
Please consider exposing the underlying casbin API in the casdoor SDK (JS).

hsluoyz:
@KrishnaPG which API do you need?

KrishnaPG:
In the front-end, it is usually required to disable some routes, buttons and menus based on the permissions of the current user.

For example, "create new record" button disabled for read-only permitted users.

Currently, using Casdoor as the authorization system, there does not seem to be anyway for the frontEnd to tell which routes/menus/buttons are permitted for the current user, unless the action is actually performed and server accepts/rejects the action.

The API required from the casdoor of underlying casbin is: know in a single call which all actions/resources on the page have access to the current user and which are not.  i.e. given a list of resources/actions return back yes/no for each of them in a single API call.

To be specific:

- the [batchEnforce](https://casbin.org/docs/en/management-api#batchenforce) API of casbin
- the [GetAllObjects](https://casbin.org/docs/en/management-api#getallobjects) - needed for enumerating all objects from the frontEnd side (useful to construct routes/buttons etc.)
- [GetAllActions](https://casbin.org/docs/en/management-api#getallactions)
- [GetAllRoles](https://casbin.org/docs/en/management-api#getallroles)

hsluoyz:
@seriouszyx

Issue #849, opened 2022-07-03
Use Casdoor to manage Casbin DB table


KrishnaPG:
The documentation says: 

> With [Casbin](https://casbin.org/) based authorization management, Casdoor supports ACL, RBAC, ABAC, RESTful accessing control models.

But nowhere any information could be found how to validate a user with RBAC or RESTful access control.

Been struggling to find out how to use the casdoor for authorization. 

1. Created the users (imported them from ldap)
2. Created permissions
3. Retrieving JWT token from front-end
4. ???? 

How to use the JWT token to validate the access permissions (i.e. if the user has roles required etc.), since the JWT does not include the roles for the user

thanhpd56:
I'm confusing about this too.

sersoong:
This is very complicated.
Maybe this is semi-manufactures.
No example.

redochenzhen:
any updates?  I'm stuck here, too.

zhuyalun888:
any updates? I'm stuck here, too.

KrishnaPG:
The new APIs seem to be close to achieving this. But not clear on the exact setup/sequence. Some relevant info can be found here for those following this thread: https://github.com/casdoor/casdoor/pull/955#issuecomment-1206368214

KrishnaPG:
This can be resolved once the TODOs in the https://github.com/casdoor/casdoor/issues/981 are done.

hsluoyz:
@KrishnaPG finally done by: https://github.com/casdoor/casdoor/pull/1317

PR #955, opened 2022-08-04
feat: expose some casbin APIs

seriouszyx:

Fix: https://github.com/casdoor/casdoor/issues/846

later provide a demo link


hsluoyz:
@KrishnaPG plz review

KrishnaPG:
Thank you @hsluoyz  

(Not a GO expert, hence) wondering if the `/api/enforce` also cater for array input (i.e. batch of requests)?

The scenario is: FrontEnd UIs need to control the appearance of buttons/menus etc on the page based on which actions/resources on the page have access to the current user and which do not. i.e. given a list of resources/actions return back yes/no for each of them in a single API call.

In that sense the [batchEnforce](https://casbin.org/docs/en/management-api#batchenforce) seems more close to the requirement. ( I could be wrong )

If the `/api/enforce` does not work with batch input, it would be great if you could kindly enable the `/api/batchEnforce` option also, in addition to the `/api/enforce`. 

While the `/api/enforce` can be called multiple times in a loop to achieve the same effect, it would incur too much round-trip costs.

KrishnaPG (approved):
I am not a GO expert - but based on the feature request the code seems to be addressing the required routes. Thank you.

seriouszyx:
@KrishnaPG I have deployed a demo site to test: http://101.43.193.167:8000/. You can log in with user `admin/123` to make any modifications.  And you can call this PR's APIs to test whether their response meets your needs. Welcome to comment! 

Tips: You should log in in and get `casdoor_session_id` from cookies. 

## BatchEnforce

request

```
curl --location --request POST 'http://101.43.193.167:8000/api/batch-enforce' \
--header 'Content-Type: text/plain' \
--header 'Cookie: casdoor_session_id=6a1b6092de0a0b7d27df7a6895617a26' \
--data-raw '[{"v0":"built-in/permission_5p7fe8", "v2":"app-built-in", "v3":"write"}, {"v2":"app-built-in", "v3":"read"}]'
```

response

```
[
    false,
    true
]
```

## GetAllObjects

request

```
curl --location --request GET 'http://101.43.193.167:8000/api/get-all-objects' \
--header 'Cookie: casdoor_session_id=6a1b6092de0a0b7d27df7a6895617a26'
```

response

```
[
    "app-built-in"
]
```


## GetAllActions

request

```
curl --location --request GET 'http://101.43.193.167:8000/api/get-all-actions' \
--header 'Cookie: casdoor_session_id=6a1b6092de0a0b7d27df7a6895617a26'
```

response

```
[
    "read"
]
```

## GetAllRoles

request

```
curl --location --request GET 'http://101.43.193.167:8000/api/get-all-roles' \
--header 'Cookie: casdoor_session_id=6a1b6092de0a0b7d27df7a6895617a26'
```

response 

```
[
    "role_8sdgoe"
]
```

KrishnaPG:
Thank you much @seriouszyx The API per se worked fine with curl, but I am kind of lost about the cookie requirement.

- Would this API work only with cookies? 

That would tie this very tightly with the front-end kind of applications making it difficult to call these API from backends. 

For example, as listed in https://github.com/casdoor/casdoor/issues/849#issue-1292219196 in a typical grant-token auth flow the front-end will finally have a JWT or token, and most of the times the API are accessed with `Authorization: Basic xxx` kind of secure headers as recommended [here](https://casdoor.org/docs/how-to-connect/oauth#how-to-use-accesstoken). 

That would allow the backends play nicely with the API: `frontEnd -> backEnd -> casdoor`

seriouszyx:
@KrishnaPG We have a [filter](https://github.com/casdoor/casdoor/blob/e94ada9ea2f03b70421fbc70ebde19911f55d98d/routers/auto_signin_filter.go#L30-L51) to parse the request param (such as access_token) and save it to the session automatically. What's more, we will add these APIs to [casdoor-js-sdk](https://github.com/casdoor/casdoor-js-sdk), and then you can use them more conveniently.

KrishnaPG:
Thank you. I am trying to figure out the right sequence of these API calls or the setup required to achieve the [scenario](https://github.com/casdoor/casdoor/issues/849): 

- front-end dynamically knowing in a single call which all actions on the page the user has access to  and which are not.

The generic flow of this API sequence is something as below. Could you please help us relate it to these new API:

1. FrontEnd logs in and gets a token (say a JWT)
2. On the front-end page there are list of actions (such as CRUD buttons, menus etc.) and it needs to know which of them are permitted to the user (so that it can enable/disable/hide those buttons/menus etc.). So the front-end makes an internal call to the backend by sending all the actions found on that page
3. The backend forwards all those actions to the casdoor `/api/batch-enforce` API along with the JWT and current application name, and gets back an array of true/false that it will return the front-end.

For example:
```
POST 'http://101.43.193.167:8000/api/batch-enforce'
--header 'Authization: Basic xyz'
--data-raw [{"v2":"my-app", "v3":"write.dataset1"}, {"v2":"my-app", "v3":"delete.dataset2"}]'
```

The front-end is asking if the `current user` has permission in `my-app` to `write` to `dataset1`, and `delete` for `dataset2`.

- It is not clear where in the CASDOOR UI do we need to configure these `write` to `dataset1` and `delete` for `dataset` etc. permissions. 
   -  It looks like the permissions in casdoor UI are tied to Roles and the `/api/batch-enforce` is not accepting roles.   
- Or, is the above flow incorrect? What is the right way to map the UI actions to the parameters (the POST request body values) of `/api/batch-enforce` API

seriouszyx:
@KrishnaPG For the requirements you said, we can use the `RBAC with domains` model. You can use this [online editor](https://casbin.org/casbin-editor/#3LPLAXESB) to test. For `/api/batch-enforce` API, the request should be like the below. 

```
POST 'http://101.43.193.167:8000/api/batch-enforce'
--header 'Authization: Basic xyz'
--data-raw [{"v2":"my-app", "v3":"dataset1", "v4":"write"}, {"v2":"my-app", "v3":"dataset2", "v4":"delete"}]'
```

However, permissions for the RBAC model is currently under development and will be completed ASAP, including the optimization of role and the support of domain. I tend to open a new PR for those features.

KrishnaPG:
Thank you much @seriouszyx It is slowly becoming clear how all these things need to be put together.

From the Casdoor UI perspective, I guess the RBAC-policy that is created in the online editor will go into the [Models](http://101.43.193.167:8000/models) section of Casdoor UI. 

And the

- `"v2":"my-app"` will be the [Applications](http://101.43.193.167:8000/applications) part of Casdoor ui

The `"v3":"dataset1"` and `"v4":"write"` it is not clear where they map to in the Casdoor UI?

- I guess the `"v4":"write" / "read"` etc. will be the [Roles](http://101.43.193.167:8000/roles) part of Casdoor UI, correct?

That leaves the `"v3":"dataset1"` where will it have to be configured in the Casdoor UI?

seriouszyx:
@KrishnaPG Most of what you describe is correct, except for the resources. The resources in Casdoor mean something that needs protecting. Currently, the type of resource only contains `Application`, which determines whether a user can log in to an application ([doc](https://casdoor.org/docs/permission/overview#permission-for-applications)). 

But In your case, the resource is `dataset`. So I think the flow should be as follows.

1. Design casbin's model according to your scenario and test it in the online editor.
2. Add the new model in the `Models` part of CASDOOR UI.
3. Add new permission in the `Permissions` part of CASDOOR UI.

image

4. Call the casbin APIs.

KrishnaPG:
Thank you much for the detailed explanation @seriouszyx The last point then remaining is: defining the resources `dataset1`, `dataset2` etc 

- Currently the [Resources](http://101.43.193.167:8000/resources) UI in Casdoor is expecting to upload a file. 

image

Unlike the other tabs in the Cadoor UI (such as Applications, Users, Roles etc.) there is no **Add New** button for the `Resources` Tab. Seems uploading is the only way.

- Usually these `dataset1`, `dataset2` etc. are known at runtime dynamically for the backend (and may not be always known statically upfront). 
- Is there a way to **create** and **manage** these **Resources programmatically** through API 
   -  there seems to be an API named `/api/add-resource` but it is not accepting any parameters (refer swagger screenshot below)
image

----------------- 
Q: 

1. Wondering how to **create** and **manage** these **Resources programmatically** on the fly?
2. The list of possible actions (namely, `Read` `Write` `Delete` etc.) are they predefined and fixed, or can they be created dynamically through API?
   -  E.g. The actions may vary based on the context (domain), such as `Print`, `Clone`, `Download`, `Lease` etc. 

It would be great if you could allow these **actions** to be **defined and managed programmatically** through API and not hardcode them to be just CRUD.


image

seriouszyx:
@KrishnaPG The resources in `Permissions` and `Resources`  are two completely different concepts. In `Permissions`, both resources and actions can be created directly in the CASDOOR UI. You can edit the permission on the [demo site](http://101.43.193.167:8000/permissions/built-in/permission_5p7fe8).

[Built-in-Organization (1).webm](https://user-images.githubusercontent.com/33992371/183257299-07d0998a-16da-4270-94b6-fb89f857ebfd.webm)

KrishnaPG:
oh...ok Great. Thank you @seriouszyx 

Then I guess it also solves the issue of: creating and managing Actions and Resources programmatically on the fly, since there seems to be `/api/add-permission` that allows both `actions[]` and `resources[]` be specified in its POST body.

I will try to create a simple `front-end/backend` demo app for beginners based on all these points we discussed here, demonstrating how to use Casdoor for dynamic authorization. (Will wait for the RBAC and other TODOs mentioned in this thread to get implemented.) 

Thank you much again for taking time to clarify all these points. Would have been almost impossible to figure them out oneself.

hsluoyz:
@KrishnaPG can you also help review this PR?

KrishnaPG (approved):
Not much of a GO expert, but based on the code and the feature request, these changes seem ok to me.

Would be great to see the unit-test case results for these API.

object/permission.go
```diff
@@ -45,13 +45,13 @@ type Permission struct {
...
 
 type PermissionRule struct {
-	PType string `xorm:"varchar(100) index not null default ''"`
-	V0    string `xorm:"varchar(100) index not null default ''"`
-	V1    string `xorm:"varchar(100) index not null default ''"`
-	V2    string `xorm:"varchar(100) index not null default ''"`
-	V3    string `xorm:"varchar(100) index not null default ''"`
-	V4    string `xorm:"varchar(100) index not null default ''"`
-	V5    string `xorm:"varchar(100) index not null default ''"`
+	PType string `xorm:"varchar(100) index not null default ''" json:"pType"`
```
hsluoyz, line 48: Now we use `Ptype` for varname and `ptype` for json.

controllers/permission.go
```diff
@@ -114,3 +114,65 @@ func (c *ApiController) DeletePermission() {
 	c.Data["json"] = wrapActionResponse(object.DeletePermission(&permission))
 	c.ServeJSON()
 }
+
+func (c *ApiController) Enforce() {
```
hsluoyz, line 118: Put all this new code to a new file: `enforcer.go`

controllers/permission.go
```diff
@@ -114,3 +114,65 @@ func (c *ApiController) DeletePermission() {
...
+	userId := c.GetSessionUsername()
+	if userId == "" {
+		c.ResponseError("Please sign in first")
```
hsluoyz, line 121: You forgot to add `return` after `ResponseError`, similar for all below

hsluoyz (changes requested):

hsluoyz (approved):

KrishnaPG:
Created a new issue https://github.com/casdoor/casdoor/issues/981 to track the pending TODOs discussed in this thread.

Issue #968, opened 2022-08-06
bug: panic occurs when creating a user using add-user API in a non-existent organization


BeSoBad:
Response:
```
<!DOCTYPE html>
<html>
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
    <title>beego application error</title>
    <style>
        html, body, body * {padding: 0; margin: 0;}
        #header {background:#ffd; border-bottom:solid 2px #A31515; padding: 20px 10px;}
        #header h2{ }
        #footer {border-top:solid 1px #aaa; padding: 5px 10px; font-size: 12px; color:green;}
        #content {padding: 5px;}
        #content .stack b{ font-size: 13px; color: red;}
        #content .stack pre{padding-left: 10px;}
        table {}
        td.t {text-align: right; padding-right: 5px; color: #888;}
    </style>
    <script type="text/javascript">
    </script>
</head>
<body>
    <div id="header">
        <h2>casdoor:runtime error: invalid memory address or nil pointer dereference</h2>
    </div>
    <div id="content">
        <table>
            <tr>
                <td class="t">Request Method: </td><td>POST</td>
            </tr>
            <tr>
                <td class="t">Request URL: </td><td>/api/add-user?id=/test-user123</td>
            </tr>
            <tr>
                <td class="t">RemoteAddr: </td><td>127.0.0.1</td>
            </tr>
        </table>
        <div class="stack">
            <b>Stack</b>
            <pre>C:/Program Files/Go/src/runtime/panic.go:844
C:/Program Files/Go/src/runtime/panic.go:220
C:/Program Files/Go/src/runtime/signal_windows.go:255
c:/Users/Artem/repos/casdoor/object/user_cred.go:34
c:/Users/Artem/repos/casdoor/object/user.go:420
c:/Users/Artem/repos/casdoor/controllers/user.go:189
C:/Program Files/Go/src/reflect/value.go:556
C:/Program Files/Go/src/reflect/value.go:339
C:/Users/Artem/go/pkg/mod/github.com/astaxie/beego@v1.12.3/router.go:897
C:/Program Files/Go/src/net/http/server.go:2916
C:/Program Files/Go/src/net/http/server.go:1966
C:/Program Files/Go/src/runtime/asm_amd64.s:1571
</pre>
        </div>
    </div>
    <div id="footer">
        <p>beego 1.12.3 (beego framework)</p>
        <p>golang version: go1.18.3</p>
    </div>
</body>
</html>
```

[...]

PR #969, opened 2022-08-06
fix: panic when creating a user in a non-existent org

BeSoBad:
If you try to create a user in a non-existent org then panic will occur. This PR fixes this.

Fix: https://github.com/casdoor/casdoor/issues/968


hsluoyz:
@BeSoBad plz fix:

image

Issue #941, opened 2022-07-31
Support signin page UI with frame border and background image


hsluoyz:
Like:

![image](https://user-images.githubusercontent.com/3787410/182035624-098d93fc-a7d6-47cd-af34-734dc5eaba82.png)

This PR: https://github.com/casdoor/casdoor/pull/978 can be used as an example.

![{D9A2CFF9-7368-D870-B319-FE813747ECA0}](https://user-images.githubusercontent.com/3787410/185942051-aba3196c-af34-49b6-9cbc-5ecca440f84a.jpg)

![{56484345-9FD8-9F61-A23B-64AC4B041720}](https://user-images.githubusercontent.com/3787410/185942062-296053ed-b8e6-4b43-8a58-4902a73ed12e.jpg)

![{01E2C15B-7517-767C-27F3-E0F98384225E}](https://user-images.githubusercontent.com/3787410/185942087-31acd82a-d66d-4922-9a18-ddc85393375b.jpg)

hsluoyz:
@leo220yuyaodog the previous two PRs have failed, you can refer to them

PR #978, opened 2022-08-07
feat: support signin page UI with frame border

qiuyuyin:
Fix: https://github.com/casdoor/casdoor/issues/941

Just add a grey background to the login page and add a frame border to the login and sign up page, next pr will add the user custom background image


fernandolguevara:
@qiuyuyin Can you move the hardcoded text (ex: made with) on this page to i18n files?

Issue #958, opened 2022-08-05
fix all frontend warnings


nomeguy:
like

![image](https://user-images.githubusercontent.com/85475922/183026945-07bffc79-5689-4d4d-b904-546d478e6ba7.png)

hsluoyz:
@joy0010 @AnjanaVakil @jakiuncle can anyone work on this?

qianxi0410:
Now, with `yarn fix`, you can show these warnings to the console. @jakiuncle

qianxi0410:
This issue should be closed by #983

hsluoyz:
Fixed by: https://github.com/casdoor/casdoor/pull/983

PR #982, opened 2022-08-07
fix: frontend warning fixes

Rohanfizz:
Hi,
In this PR, I fixed all the frontend warnings listed below with what I did to fix them. Please review.
Issue= #958 

**1. Alot of unused variables**
There were alot of unused variables, so instead of deleting each and every variable, I decided to turn off no-unused-vars from .eslintrc, as at this point I don't think anyone takes that warning seriously.

**2. Statements inside switch case.**
Statements are written inside switch case, which is discouraged according to ->
https://stackoverflow.com/questions/50752987/eslint-no-case-declaration-unexpected-lexical-declaration-in-case-block
Added scope braces => {} around every warned case.

**3. Key props not given to elements inside array.**
I gave them a "placeholderKey".

**4. componentDidMount used and is not safe** to use in react according to => https://stackoverflow.com/questions/46043832/why-componentwillmount-should-not-be-used
replaced it with componentDidMount

### DIFF
--- a/web/.eslintrc
+++ b/web/.eslintrc
@@ -55,7 +55,7 @@
     "react/react-in-jsx-scope": "off",
 
     // don't use strict mod now, otherwise there are a lot of errors in the codebase
-    "no-unused-vars": "warn",
+    "no-unused-vars": "off",
     "react/no-deprecated": "warn",
     "no-case-declarations": "warn",
     "react/jsx-key": "warn"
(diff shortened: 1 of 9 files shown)


FoeverA0:
@Rohanfizz It looks like you need to change the title of ur commit, it has to be in the form of fix:+ spaces +"ur title" to pass the Semantic PR check, I know because I just couldn't pass the Semantic PR check because of this. And remember that it must be "fix" and not "Fix". Hope this helps you : )

hsluoyz:
Sorry but the PR quality is not satisfactory, so closed here

Rohanfizz:
> 

Sure, It was my first time contributing to this repo😅
Ill make another PR and try to make it pass all the semantic checks👍

PR #986, opened 2022-08-08
fix: Can not use SMTP provider with MYSUBMAIL.COM at 465 port

FoeverA0:
Add an enable SSL switch.


FoeverA0:
Sorry to disappoint you again, but thanks for everything you have done for me :)

Issue #993, opened 2022-08-09
bug: wrong name mapping 'p_type'


brookqin:
casdoor:
https://github.com/casdoor/casdoor/blob/deed8577884501bdeeac35f43cacd39aabf9cbc5/object/permission.go#L43
`Ptype` map to `ptype`;

xorm-adapter:
https://github.com/casbin/xorm-adapter/blob/7a90dd8bfb2e70b3a768479c866162b4d2a4cb7b/adapter.go#L39
`PType` map to `p_type`;

when init database, got a warning
> [xorm] [warn]  2022/08/09 11:25:22.784737 Table permission_rule has column ptype but struct has not related field

and schema of `permission_rule` became to `ptype, v0, v1, v2, v3, v4, v5, p_type`

hsluoyz:
@brookqin `Ptype` is correct. xorm-adapter code will also be modified: https://github.com/casbin/xorm-adapter/issues/53

PR #994, opened 2022-08-09
fix: PType name mapping

brookqin:
#993


hsluoyz:
See: https://github.com/casdoor/casdoor/issues/993#issuecomment-1208892774

PR #943, opened 2022-08-01
feat: Support sub-directory

zzjin:
By adding PUBLIC_URL to relative `.`, we can access casdoor's all web pages and static asserts under relative public url.

Closes #933

Signed-off-by: zzjin <tczzjin@gmail.com>

PR #999, opened 2022-08-10
feat: delete .env to fix static file path bug

leo220yuyaodog:
Revert: https://github.com/casdoor/casdoor/pull/943 @zzjin 

Bug (white screen):

54A@J$LR0WCS{M$7RSE9Z{E

### DIFF
--- a/web/.env
+++ b/web/.env
@@ -1,1 +1,0 @@
-PUBLIC_URL=.


zzjin:
Strange, How do  you start test env?

leo220yuyaodog:
> Strange, How do you start test env?
> 
> > Revert: #943 @zzjin
> > Bug (white screen):
> > 54A@J$LR0WCS{M$7RSE9Z{E
> 
> After lookat you screenshort, it'ss seems to be a backend bug as you do not spec the `redirect_uri=undefined` and the port is 7001 not 8000?
> 
> As result, if anything frontend is wrong, it will show a error page:
> 
> image

Add the .env ,  run` yarn build`, the files in `web/build/asset-manifest.json`  will has prefix `.` , it cause except app-built-in application, the login page is white. It because the request url is nomatch with the file path in `asset-manifest.json`. So I delete the .env . It is good  if you had a better solution.

PR #1001, opened 2022-08-10
fix: support createDatabase on other db

akkuman:
Currently, only mysql is supported to create databases

The PR Added support for creating other db


hsluoyz:
@akkuman Casdoor should not create a new DB, the legacy code is remained there but we will not further add create DB code

Issue #996, opened 2022-08-09
Accelerate CI


qianxi0410:
Note that the front-end related CI is really slow, which seriously slows down the build. Note that the `yarn install` step is the cause of the serious slowdown, and this step can be accelerated using the cache.

hsluoyz:
@qianxi0410 good point, how to add the cache then?

qianxi0410:
see: [yarn-cache](https://github.com/marketplace/actions/yarn-install-cache)
@hsluoyz

hsluoyz:
@qianxi0410 good, plz make a PR to add it

PR #1004, opened 2022-08-10
chore(ci): use cache to accelerate ci

qianxi0410:
Close: #996


qianxi0410:
As you can see, the yarn install time has dropped from 90s to 20s @hsluoyz

Issue #1002, opened 2022-08-10
Require mock send verification code.


kofj:
In some case, like running casdoor in the CICD, we want mock verification code.

kofj:
I will create some PR later.

kofj:
SMS Sender PR https://github.com/casdoor/go-sms-sender/pull/11

PR #1009, opened 2022-08-12
feat: support Mock SMS provider

kofj:
1. Update go-sms-sender to v0.3.0.
2. Fix: avoid page crash if not found provider info.

Signed-off-by: 疯魔慕薇 <kfanjian@gmail.com>

Fix: https://github.com/casdoor/casdoor/issues/1002

### DIFF
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -49,6 +49,10 @@ export const OtherProviderInfo = {
       logo: `${StaticBaseUrl}/img/social_huawei.png`,
       url: "https://www.huaweicloud.com/product/msgsms.html",
     },
+    "Mock SMS": {
+      logo: `${StaticBaseUrl}/img/social_default.png`,
+      url: "",
+    },
   },
   Email: {
     "Default": {
@@ -532,7 +536,12 @@ export function getProviderLogoURL(provider) {
     }
     return `${StaticBaseUrl}/img/social_${provider.type.toLowerCase()}.png`;
   } else {
-    return OtherProviderInfo[provider.category][provider.type].logo;
+    const info = OtherProviderInfo[provider.category][provider.type];
+    // avoid crash when provider is not found
+    if (info) {
+      return info.logo;
+    }
+    return "";
   }
 }
 
--- a/web/src/auth/Provider.js
+++ b/web/src/auth/Provider.js
@@ -137,7 +137,12 @@ export function getProviderUrl(provider) {
 
     return `${urlObj.protocol}//${host}`;
   } else {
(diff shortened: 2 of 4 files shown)


kofj:
#1002

kofj:
@hsluoyz Review pls, thx.

web/src/auth/Provider.js
```diff
@@ -137,7 +137,8 @@ export function getProviderUrl(provider) {
 
     return `${urlObj.protocol}//${host}`;
   } else {
-    return Setting.OtherProviderInfo[provider.category][provider.type].url;
+    const info = Setting.OtherProviderInfo[provider.category][provider.type];
+    if (info) {return info.url;}
```
hsluoyz, line 141: @qianxi0410 why the linter doesn't report this?

web/src/Setting.js
```diff
@@ -532,7 +536,8 @@ export function getProviderLogoURL(provider) {
     }
     return `${StaticBaseUrl}/img/social_${provider.type.toLowerCase()}.png`;
   } else {
-    return OtherProviderInfo[provider.category][provider.type].logo;
+    const info = OtherProviderInfo[provider.category][provider.type];
+    if (info) {return info.logo;}
```
hsluoyz, line 540: @qianxi0410 why the linter doesn't report this?

hsluoyz (changes requested):

web/src/Setting.js
hsluoyz: @kofj you already added new item to `OtherProviderInfo`, why add `if` here?

web/src/auth/Provider.js
hsluoyz: @kofj you already added new item to `OtherProviderInfo`, why add `if` here?

web/src/Setting.js
kofj: In other cases maybe cause page crash, so we need check it first.

web/src/auth/Provider.js
kofj: In other cases maybe cause page crash, so we need check it first.

hsluoyz: Code format should be:

```js
if (info) {
  return info.url;
}
```

And make sure all paths return value

web/src/Setting.js
hsluoyz: same as above

web/src/auth/Provider.js
kofj: resolved

hsluoyz (approved):

Issue #942, opened 2022-08-01
Modify the link of  social app image url and put these images on the local server to improve the access speed.


jyxenlink:
The social app image address of the login page is https://cdn.casbin.org/img/xxx.png , the access speed is slow. I suggest to modify the link and put these images on the local server to improve the access speed.

imp2002:
I'm working on it.

hsluoyz:
We need to:

1. Wrap all system images (not user avatar images) into: `StaticBaseUrl`, which is `https://cdn.casbin.org` by default.
2. If some images are missing from the above CDN, then add it by making a PR to: https://github.com/casbin/static

hsluoyz:
@leo220yuyaodog

leo220yuyaodog:
> The social app image address of the login page is https://cdn.casbin.org/img/xxx.png , the access speed is slow. I suggest to modify the link and put these images on the local server to improve the access speed.

Your can modify the `StaticBaseUrl` in `web/src/Setting.js `

PR #1015, opened 2022-08-13
feat: use staticBaseUrl for all static resources

leo220yuyaodog:
Fix: #942


web/src/App.less
```diff
@@ -46,7 +46,7 @@
 }
 
 .language_box {
-  background: url("https://cdn.casbin.org/img/muti_language.svg");
+  background: url("./assets/icons/muti_language.svg");
```
hsluoyz, line 49: Also define `staticBaseUrl` in css

hsluoyz (changes requested):

hsluoyz (approved):

Issue #981, opened 2022-08-07
RBAC with domains + Object resource types in Permissions


KrishnaPG:
Creating this issue to track the TODOs discussed here: https://github.com/casdoor/casdoor/pull/955

- RBAC with domains model
- Resource type other than Application in the Permissions

![image](https://user-images.githubusercontent.com/4610831/183300293-fd2a58dd-e6cc-4a00-8632-bea3e2656a46.png)
![image](https://user-images.githubusercontent.com/4610831/183300318-f9bb80b0-d7d8-4f60-b06e-88421d232689.png)

These are required to make the Enforcement API work correctly. Ref: https://github.com/casdoor/casdoor/issues/849 and https://github.com/casdoor/casdoor/issues/846

hsluoyz:
@seriouszyx

PR #1020, opened 2022-08-15
feat: support RBAC with domains model and add adapter to specify the table name for policy storage

seriouszyx:

For: https://github.com/casdoor/casdoor/issues/981

### DIFF
--- a/object/permission_enforcer.go
+++ b/object/permission_enforcer.go
@@ -24,8 +24,12 @@ import (
 )
 
 func getEnforcer(permission *Permission) *casbin.Enforcer {
+	tableName := "permission_rule"
+	if len(permission.Adapter) != 0 {
+		tableName = permission.Adapter
+	}
 	tableNamePrefix := conf.GetConfigString("tableNamePrefix")
-	adapter, err := xormadapter.NewAdapterWithTableName(conf.GetConfigString("driverName"), conf.GetBeegoConfDataSourceName()+conf.GetConfigString("dbName"), "permission_rule", tableNamePrefix, true)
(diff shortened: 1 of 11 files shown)


hsluoyz:
@KrishnaPG plz review

object/permission_enforcer.go
```diff
@@ -35,7 +39,7 @@ func getEnforcer(permission *Permission) *casbin.Enforcer {
 r = sub, obj, act
 
 [policy_definition]
-p = permission, sub, obj, act
+p = sub, obj, act
```
KrishnaPG, line 42: Thank you @seriouszyx In one of the screens you have shared earlier for the RBAC + domains model, the policy looked like having 4 variables 

image

Wondering about this `p = sub, obj, act` since it seems to be lacking the domain vairable

KrishnaPG (commented):
Not much of a go expert. But regarding the policy `p = sub, obj, act` wondering if it would be alright without the domain variable in the policy definition

object/permission_enforcer.go
seriouszyx: In order to prevent the program from crashing due to the model in permission being empty, I set a default model (RBAC model) by hard coding. If using the `RBAC with domains` model, you can add a new model and assign it to permission. 

As for the `permission` column (p = permission, sub, obj, act) in the policy definition, we previously set it to distinguish between different permissions in the same database table. But now we support the `adapter` configuration to specify the table name for policy storage. So the `permission` column is unnecessary.  

After this PR is merged, I will refactor the permission documentation and explain the exposed casbin APIs (https://github.com/casdoor/casdoor-website/issues/272).

hsluoyz: @KrishnaPG

KrishnaPG (approved):
Not a Go expert. Based on the requirement, looks ok.

PR #1037, opened 2022-08-19
feat(storage): add support for min.io storage

fernandolguevara:
min.io server is compatible with aws s3 interface 🙃

for reference [go code](https://docs.min.io/docs/how-to-use-aws-sdk-for-go-with-minio-server.html)


hsluoyz:
@fernandolguevara plz also add this to docs after: https://casdoor.org/docs/provider/storage

PR #1043, opened 2022-08-20
fix(i18n): add spanish translations

fernandolguevara:
image

Unfortunately we do not have the Spanish flag in choosed IconFontSet

```js
const IconFont = createFromIconfontCN({
  scriptUrl: "//at.alicdn.com/t/font_2680620_ffij16fkwdg.js",
});
```


hsluoyz:
@Selflocking @leo220yuyaodog any thoughts about how to fix the Spanish icon? I think the IconFontSet should already have it

fernandolguevara:
@hsluoyz we can use this[ flag set](https://catamphetamine.gitlab.io/country-flag-icons/3x2/index.html)

image

hsluoyz:
@fernandolguevara 

1. Plz provider a screenshot for the new flags
2. The new flag images should be put to: https://github.com/casbin/static to avoid the URL failure in future
3. Can you add some code to allow customizing the option orders in Conf.js ?

fernandolguevara:
image

image

PR #1064, opened 2022-08-25
feat: support signin page UI with frame border and background img

joy0010:
Fix: #941 

- Achived background image and login box border in the LoginPage.
- Made the footer fixed, so that it wouldn't disappear when people drag and drop to change the size of window.


nomeguy:
Replaced by: https://github.com/casdoor/casdoor/pull/1107

PR #1076, opened 2022-08-30
fix: Add default access permission for new built-in group users

cofecatt:


hsluoyz:
@seriouszyx plz review

PR #1096, opened 2022-09-04
fix: fix bugs about 3rd-party login in cas flow

ComradeProgrammer:
fix: fix bugs about 3rd-party login in cas flow


hsluoyz:
@ComradeProgrammer fix linter:

image

PR #1136, opened 2022-09-17
fix: check whether to use proxy in build

gtn1024:
image

image


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/1149

Issue #1200, opened 2022-10-09
Fix the UI problems in pad and mobile views


hsluoyz:
https://door.casdoor.com/

Pad:

![image](https://user-images.githubusercontent.com/3787410/194757635-550e2260-e66f-45be-b202-1b7b73692295.png)

Mobile:

![image](https://user-images.githubusercontent.com/3787410/194757683-a4023a90-294a-4139-aab8-961f45a299b3.png)

![image](https://user-images.githubusercontent.com/3787410/194757697-cc719e59-8d82-43f7-8027-2c05a541e887.png)

PR #1202, opened 2022-10-09
feat: fix pad and mobile views

imchell:
Fix: https://github.com/casdoor/casdoor/issues/1200

The problem was described in #1200.

## Improved

* Proper image size in both pad view and mobile view
* Menu no longer covers avatar and language switch

pad view
mobile view

## Working On

* The layout of menu in mobile view looks like a list. My opinion is to make it collapsible. Should the design orientation be discussed in advance?

mobile view menu


imchell:
Update mobile menu style

unexpanded
expanded

Issue #1208, opened 2022-10-10
[frontend] The spin for account page should be centralized


hsluoyz:
![image](https://user-images.githubusercontent.com/3787410/194895737-3e3745f6-a9ed-4868-bbda-38fef9f255a7.png)

![image](https://user-images.githubusercontent.com/3787410/194895953-c390ed61-8a56-4e4b-825c-873aeb3f9f6c.png)

PR #1209, opened 2022-10-10
fix: center loading in account page

xxrlzzz:
Fix: https://github.com/casdoor/casdoor/issues/1208


xxrlzzz:
截屏2022-10-11 00 45 59

Issue #1213, opened 2022-10-11
[frontend] make the app list in homepage have the same height


hsluoyz:
Log into: https://door.casdoor.com/login/casbin

![image](https://user-images.githubusercontent.com/3787410/195133904-68ae6b01-d521-43bc-ab7c-09cde8522012.png)

It should be:

![CA(NU9{G)T~~FT(OK E~AH](https://user-images.githubusercontent.com/3787410/195911898-d4b2428d-12e1-4b31-9ee2-1d33d61f9fbf.png)

![4B@RJ$9}J{U_S1A_(WUVI F](https://user-images.githubusercontent.com/3787410/195911913-f81e8b83-1ece-4474-b096-c2dc38905d8c.png)

PR #1220, opened 2022-10-14
fix: frontend application list should have same height

xxrlzzz:
for #1213


xxrlzzz:
截屏2022-10-15 00 55 40

Aloento:
圣地巡礼

forever14998:
圣地巡礼

Issue #1181, opened 2022-10-03
The page switch should be faster by using frontend routing instead of full page reload


hsluoyz:
See links at: https://door.casdoor.com/login

![image](https://user-images.githubusercontent.com/3787410/193609392-3e1140c2-6caa-4c14-9b88-5a3a0af4adef.png)

PR #1227, opened 2022-10-17
fix: forget password and sign up router

imchell:
Fix: #1181 

page router

### DIFF
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -786,16 +786,16 @@ export function goToSignup(ths, application) {
   }
 }
 
-function JumpLink({url, children}) {
+function JumpLink({url, children, onClick}) {
   if (url === null) {
-    return <Link style={{float: "right"}}>{children}</Link>;
+    return <Link style={{float: "right"}} onClick={onClick} to={""}>{children}</Link>;
   }
   if (url.startsWith("/")) {
-    return <Link to={url} style={{float: "right"}}>{children}</Link>;
+    return <Link to={url} style={{float: "right"}} onClick={onClick}>{children}</Link>;
   } else if (url.startsWith("http")) {
-    return <a href={url} target="_blank" rel="noopener noreferrer" style={{float: "right"}}>{children}</a>;
+    return <a href={url} target="_blank" rel="noopener noreferrer" style={{float: "right"}} onClick={onClick}>{children}</a>;
   } else {
-    return <Link style={{float: "right"}}>{children}</Link>;
+    return <Link style={{float: "right"}} onClick={onClick} to={""}>{children}</Link>;
   }
 }
 
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -761,28 +761,6 @@ export function goToLogin(ths, application) {
   }
 }
 
-export function goToSignup(ths, application) {
-  if (application === null) {
-    return;
-  }
-
-  if (!application.enablePassword && window.location.pathname.includes("/login/oauth/authorize")) {
-    const link = window.location.href.replace("/login/oauth/authorize", "/auto-signup/oauth/authorize");
-    goToLink(link);
-    return;
-  }
-
-  if (authConfig.appName === application.name) {
-    goToLinkSoft(ths, "/signup");
-  } else {
-    if (application.signupUrl === "") {
-      goToLinkSoft(ths, `/signup/${application.name}`);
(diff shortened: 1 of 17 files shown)


hsluoyz:
@leo220yuyaodog plz review

web/src/Setting.js
```diff
@@ -797,6 +818,22 @@ export function goToForget(ths, application) {
...
+    return "/";
+  }
+
+  if (authConfig.appName === application.name) {
+    return "/forget";
+  } else {
+    if (application.forgetUrl === "") {
+      return `/forget/${application.name}`;
+    } else {
+      return "application.forgetUrl";
```
hsluoyz, line 832: `application.forgetUrl` can be other domain's URL

hsluoyz (changes requested):

leo220yuyaodog:
@imchell Actually, now from the login page fo to the signup and forget page is the SPA router. It can not invoke the page refresh.  From the signup page to login page will refresh the browser. 

When switch the pages will appears a short white pages, because react renders background image. The backgroud image url is from the react state `application`. And the `LoginPage`, `SignupPage` have their owe state  `application`, so react renders background iamge again, cause the problem. 

I have an idea that extract the background div to the father component. Then when switch the pages, the background will not render again.  Do you hava other good implenmentation metheds?

imchell:
@leo220yuyaodog Thanks for reviewing. Yes, the background page reloads while switching between `LoginPage` and `SignupPage`. I'm not a react-router pro but I don't know why background being extracted to the parent component helps. It is foreseeable that the extraction would lead to the same background image of all pages in `<Switch>`, which is not the intended effect, thus needing some work on decomposing elements in `<Switch>`.

leo220yuyaodog:
> @leo220yuyaodog Thanks for reviewing. Yes, the background page reloads while switching between `LoginPage` and `SignupPage`. I'm not a react-router pro but I don't know why background being extracted to the parent component helps. It is foreseeable that the extraction would lead to the same background image of all pages in `<Switch>`, which is not the intended effect.

I don't think about it enough. Fixing the front-end routing bug is probably the best thing to do right now, if it can resolve the problem.

web/src/Setting.js
```diff
@@ -781,6 +782,27 @@ export function goToSignup(ths, application) {
   }
 }
 
+export function renderDynamicSignUpLink(application, children) {
```
hsluoyz, line 785: All `children` to `text`

web/src/Setting.js
```diff
@@ -781,6 +782,27 @@ export function goToSignup(ths, application) {
...
+  if (!application.enablePassword && window.location.pathname.includes("/login/oauth/authorize")) {
+    const link = window.location.href.replace("/login/oauth/authorize", "/auto-signup/oauth/authorize");
+    return <a style={{float: "right"}} href={link} target="_blank" rel="noopener noreferrer">{children}</a>;
+  }
+
+  if (authConfig.appName === application.name) {
+    return <Link to={"/signup"}>{children}</Link>;
+  } else {
+    if (application.signupUrl === "") {
+      return <Link to={`/signup/${application.name}`}>{children}</Link>;
```
hsluoyz, line 799: Extract a function, use `url` as arg

web/src/Setting.js
hsluoyz, line 785: renderSignupLink

web/src/Setting.js
```diff
@@ -797,6 +819,22 @@ export function goToForget(ths, application) {
   }
 }
 
+export function renderDynamicForgetLink(application, children) {
```
hsluoyz, line 822: renderForgetLink

hsluoyz (changes requested):

web/src/auth/LoginPage.js
```diff
@@ -473,12 +471,9 @@ class LoginPage extends React.Component {
               !application.enableSignUp ? null : (
                 <>
                   {i18next.t("login:No account?")}&nbsp;
-                  <a onClick={() => {
-                    sessionStorage.setItem("signinUrl", window.location.href);
```
hsluoyz, line 477: `sessionStorage.setItem("signinUrl", window.location.href);` is missing

hsluoyz (changes requested):

hsluoyz (approved):

PR #1239, opened 2022-10-25
fix: make the app list in homepage have the same height

jakiuncle:
Fix: https://github.com/casdoor/casdoor/issues/1213


jakiuncle:
Display like this
image

jakiuncle:
https://user-images.githubusercontent.com/88994283/197724605-7df27b9b-d205-4dfd-841d-14278ee19878.mp4

PR #1265, opened 2022-11-06
feat(RoleEditPage): Fix a crash when the sample role domains is null

Dingzhaocheng:
Fix the crash when the role object domains is null in the on line demo role page
截屏2022-11-06 21 06 07


hsluoyz:
@Dingzhaocheng plz squash into 1 commit and sign CLA:

image

PR #1271, opened 2022-11-06
fix: rounded corner login page

imchell:
Before:
login before

After:
login after

### DIFF
--- a/web/src/App.less
+++ b/web/src/App.less
@@ -134,6 +134,11 @@
   box-sizing: border-box;
   margin: 0 auto;
   position: relative;
+  box-shadow: 0 0 20px rgb(0 0 0 / 20%);
+  padding: 10px 100px 20px;
+  border: 2px solid #fff;
+  border-radius: 7px;
+  background-color: rgb(255 255 255);
 }
 
 .loginBackground {
--- a/web/src/App.less
+++ b/web/src/App.less
@@ -105,14 +105,6 @@
   }
 }
 
-.forget-content {
-  padding: 10px 100px 20px;
-  border: 2px solid #fff;
-  border-radius: 7px;
-  background-color: rgb(255 255 255);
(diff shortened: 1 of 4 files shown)


hsluoyz:
@leo220yuyaodog plz review

web/src/App.less
```diff
@@ -134,6 +126,11 @@
   box-sizing: border-box;
   margin: 0 auto;
   position: relative;
+  box-shadow: 0 0 20px rgb(0 0 0 / 20%);
```
leo220yuyaodog, line 129: Please don't modify class `.login-panel`. It't will cause many other problem that incompatible with other feature about the custom login page ui. For beatifying the default login page.  As far as possible just modify the class `.login-form`

image

imchell: Looking into it.

imchell: Hi @leo220yuyaodog I've double checked but it seems I didn't modify `.login-panel`? How to reproduce your result shown in the screenshot?

leo220yuyaodog: > Hi @leo220yuyaodog I've double checked but it seems I didn't modify `.login-panel`. How to reproduce your result shown in the screenshot? What I have changed is `.login-content` only.

Custom the application UI follow the document https://casdoor.org/docs/application/ui-customization

imchell: > Custom the application UI follow the document https://casdoor.org/docs/application/ui-customization

I see the problem here. Does the style filled in Form CSS in casdoor overwrite the default style in `.login-panel`? If that is the case, can we just assign `.login-panel` a more unified default style?

hsluoyz: @imchell , @leo220yuyaodog wrote this part of code and is more familar with it. I think it's more suitable for you to create an issue instead and let him fix it.

Issue #1272, opened 2022-11-08
I can't jump to the business system after I reset my password by email


star7th:
My business system is connected to casdoor. Ordinary users log in using email. When they forget their email password, choose Email Reset Password. After resetting, they will stay on the casdoor page and cannot jump back to the business system.
The whole process is not friendly. The normal process is as follows:
The user clicks the login button in the business system to jump to the login page. He found that he didn't remember the password, so he reset it. After reset, he log in with the new password

PR #1288, opened 2022-11-12
feat: fix the issue of jumping back to the login page after resetting password

imchell:
Fix: #1272 

Original codes failed because:

1. `redirectToLoginPage()` was called in class components which is not compatible with react hooks (`useHistory()` in this case). `redirectToLoginPage()` now uses naive JS implementation of redirecting.
2. `path.join(application.homepageUrl, "login")` returns `login` instead of `/login`.

### DIFF
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -764,9 +764,9 @@ export function renderLoginLink(application, text) {
   return renderLink(url, text, null);
 }
 
-export function redirectToLoginPage(application) {
+export function redirectToLoginPage(application, history) {
   const loginLink = getLoginLink(application);
-  window.location.replace(loginLink);
+  history.push(loginLink);
 }
 
 function renderLink(url, text, onClick) {
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import React from "react";
-import {Link, useHistory} from "react-router-dom";
+import {Link} from "react-router-dom";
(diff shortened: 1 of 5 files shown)


hsluoyz:
@leo220yuyaodog plz review

web/src/Setting.js
```diff
@@ -766,8 +766,7 @@ export function renderLoginLink(application, text) {
 
 export function redirectToLoginPage(application) {
   const loginLink = getLoginLink(application);
-  const history = useHistory();
-  history.push(loginLink);
+  window.location.replace(loginLink);
```
hsluoyz, line 769: Why switch to hard-routing? Will it cause full page reload? This is not good

hsluoyz (changes requested):

hsluoyz (approved):

web/src/auth/ForgetPage.js
```diff
@@ -166,7 +167,7 @@ class ForgetPage extends React.Component {
     values.userOwner = this.state.application?.organizationObj.name;
     UserBackend.setPassword(values.userOwner, values.username, "", values?.newPassword).then(res => {
       if (res.status === "ok") {
-        Setting.redirectToLoginPage(this.state.application);
+        Setting.redirectToLoginPage(this.state.application, this.props.history);
```
leo220yuyaodog, line 170: If set password successfully, when I use the account to login. Show errer message
sign out before login. Maybe can redirect to the home page.

PR #1297, opened 2022-11-14
feat: encode redirectUri

MagicalSheep:
Some single page application use hash route mode for the router, which will make all URLs in this application with a character `#`. Currently, the Casdoor front-end sends the redirect URI from application parameters directly without any changes. It will cause the backend side cannot to get the correct redirect URI. 

This commit uses the `encodeURIComponent` function to encode the redirect URI parameters before requesting for the backend. It should help to fix the problem without other side effects.


hsluoyz:
@leo220yuyaodog plz review

PR #1325, opened 2022-11-25
fix: fail to return after flush the page

leo220yuyaodog:
Old methed just get the url path parameter when click the butten. But when the page flushed, the returnUrl will disappear, so can not return to the specified page.

### DIFF
--- a/web/src/UserEditPage.js
+++ b/web/src/UserEditPage.js
@@ -49,6 +49,7 @@ class UserEditPage extends React.Component {
       applications: [],
       mode: props.location.mode !== undefined ? props.location.mode : "edit",
       loading: true,
+      returnUrl: null,
     };
   }
 
@@ -57,6 +58,7 @@ class UserEditPage extends React.Component {
     this.getOrganizations();
     this.getApplicationsByOrganization(this.state.organizationName);
     this.getUserApplication();
+    this.setReturnUrl();
   }
 
   getUser() {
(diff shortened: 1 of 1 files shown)


web/src/UserEditPage.js
```diff
@@ -104,8 +104,9 @@ class UserEditPage extends React.Component {
 
   setReturnUrl() {
     const searchParams = new URLSearchParams(this.props.location.search);
-    const returnUrl = searchParams.get("returnUrl");
-    returnUrl !== null ? this.setState({returnUrl: returnUrl}) : null;
+    this.setState({
+      returnUrl: searchParams.get("returnUrl"),
```
leo220yuyaodog, line 108: The thirty-party app call API get-account. It will first through the frontend route of casdoor. 
- This time the component can get the `returnUrl` and setState.  
- Then the casdoor frontend call server api. It cause the `access_token` and `returnUrl` disappear.  setState will set `returnUrl = null`. 

So if the `returnUrl` is null. It should  not modify the state

hsluoyz: Is it OK now?

hsluoyz (changes requested):

hsluoyz (approved):

web/src/UserEditPage.js
leo220yuyaodog: Yes, it is ok now.

Issue #1279, opened 2022-11-09
need casbin "RBAC With Domains/Tenants" model


JhouXerox:
there's some domain/tenant permission l want to embed inside of casdoor.
I want to achieve like follow cases:
1. role A has permission in domain A not domain B.
like operation can write anything in prod/test/dev domain. but engineer can only write in dev domain, read in test/prod domain.

2. some resource can only be accessed by role C in tenant C
like if I created something, my partner in my department can also read the resource, not in other departments

I can use /api/enforce api to check if current user has the permission to do the action.

for now I've tried follow, test in image v1.153.0:

 I try to build the model with "RBAC With Domains/Tenants"
<img width="870" alt="image" src="https://user-images.githubusercontent.com/24403596/200711951-b4277b63-34e4-403a-9885-1aac3646a596.png">
with the test role
<img width="713" alt="image" src="https://user-images.githubusercontent.com/24403596/200712022-a2cbe6c7-2a49-452e-8a97-7dd0e3d40eed.png">

permission conf:
<img width="751" alt="image" src="https://user-images.githubusercontent.com/24403596/200712061-38eb15f9-cf53-4f08-acf3-31df695e29e4.png">

there's tenant info in my access_token
<img width="474" alt="image" src="https://user-images.githubusercontent.com/24403596/200712191-9b94ecfb-939c-46f4-b930-6e2ccc0738f1.png">

but when I call /api/enforce,it returns an exception "casdoor:grouping policy elements do not meet role definition"
<img width="1018" alt="image" src="https://user-images.githubusercontent.com/24403596/200712305-36a15958-6aca-4d77-854d-367467650b10.png">

imp2002:
If my guess is correct, your `g` policy in column `v2` is empty, just look like follows:
![image](https://user-images.githubusercontent.com/41513919/202721668-ffa1158c-67c2-463f-93ca-4069f90ee1ac.png)

It's not meet 
```ini
[role_definition]
g = _, _, _
```
lead to the error.

But it's not your false. I guess you **not set all the fields when you created the `permission` for the first time, and then set some fields again, but this is not updated to the database(**Bug**)**, lead to `v2` empty.

I will fix it.

PR #1333, opened 2022-11-27
feat: support RBAC With Domains/Tenants

imp2002:
Fix: #1279

- Fix bug: https://github.com/casdoor/casdoor/issues/1279#issuecomment-1320035549

- Feat: support RBAC With Domains/Tenants
enforce
data

### DIFF
--- a/object/permission.go
+++ b/object/permission.go
@@ -111,7 +111,27 @@ func GetPermission(id string) *Permission {
 	return getPermission(owner, name)
 }
 
+// VerifyPermission verify if the permission is valid
+func VerifyPermission(permission *Permission) {
+	enforcer := getEnforcer(permission)
+	enforcer.EnableAutoSave(false)
+	policies, groupingPolicies := getPolicies(permission)
+
+	if len(groupingPolicies) > 0 {
+		_, err := enforcer.AddGroupingPolicies(groupingPolicies)
+		if err != nil {
+			panic(err)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


hsluoyz:
@JhouXerox @leo220yuyaodog  plz review

JhouXerox:
image
image

JhouXerox:
@imp2002 hi,I ran through docker-compose via your branch "fix-1279". Enforce API returned an unexpected result.
is there anything I built wrong?

image
image
image

leo220yuyaodog:
Our implements has some shortcoming. When use the enforce, the table `permission_rule`  loads all the policies. 

image

After use the domain, the new model tokens's len is 4, but the old policies's len is 3. It cause the errer. 
image

hsluoyz:
@imp2002

imp2002:
> is there anything I built wrong

Some policy not belong to this permisson and not match this permisson's model has been store in the table. Because the under the current implementation, enforce will load all policies from a table, it load the error policy. A good way is to create a new table for this permission, just assign a table name in adapter feild, when create a new permisson it will create auto.


image

imp2002:
> the table `permission_rule` loads all the policies

Yes, it will load all policy. But I think under the current implementation, we should use different tables for different permissions to store policy. The specific method is to specify an adapter when creating the permission, the adapter can be seen as a table.

imp2002:
Of course, maybe we should impl store policy of different permisson and loadPolicyFilter by id column. It not been the column.

PR #1400, opened 2022-12-14
feat: fix length of policy and `[policy_define]` in `model`  inconsistent

imp2002:
Because of https://github.com/casdoor/casdoor/pull/1357, the policy length become `6`, **so in all model `[policy_definition]` should be the same length now.** 

```ini
[policy_definition]
p = sub, obj, act, "", "", permissionId
```

Fix: 
image

### DIFF
--- a/object/permission_enforcer.go
+++ b/object/permission_enforcer.go
@@ -41,7 +41,7 @@ func getEnforcer(permission *Permission) *casbin.Enforcer {
 r = sub, obj, act
 
 [policy_definition]
-p = sub, obj, act
+p = sub, obj, act, "", "", permissionId
 
 [role_definition]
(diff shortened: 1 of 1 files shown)


object/permission_enforcer.go
```diff
@@ -41,7 +41,7 @@ func getEnforcer(permission *Permission) *casbin.Enforcer {
 r = sub, obj, act
 
 [policy_definition]
-p = sub, obj, act
+p = sub, obj, act, "", "", permissionId
```
hsluoyz, line 44: Should not be empty string

hsluoyz (changes requested):

hsluoyz (approved):

PR #1442, opened 2023-01-03
feat(login): add code login limit

longit644:
We have a limit when signing-in with `username` and `password` but with `code` is not yet, which may be the script burst code.

### DIFF
--- a/object/verification.go
+++ b/object/verification.go
@@ -26,6 +26,10 @@ import (
 	"xorm.io/core"
 )
 
+const (
+	wrongCode = "wrongCode"
+)
+
 type VerificationRecord struct {
 	Owner       string `xorm:"varchar(100) notnull pk" json:"owner"`
 	Name        string `xorm:"varchar(100) notnull pk" json:"name"`
@@ -167,7 +171,7 @@ func CheckVerificationCode(dest, code, lang string) string {
 	}
 
 	if record.Code != code {
-		return "Wrong code!"
+		return wrongCode
 	}
 
 	return ""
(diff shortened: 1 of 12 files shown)


hsluoyz:
@qianxi0410 @leo220yuyaodog @imp2002 plz review

object/verification.go
```diff
@@ -26,6 +26,10 @@ import (
 	"xorm.io/core"
 )
 
+const (
+	wrongCode = "wrongCode"
```
leo220yuyaodog, line 30: Herer is `wrongCode `, it is diferent from the result that function `CheckVerificationCode(dest, code, lang)` return in line 199

object/verification.go
```diff
@@ -186,6 +190,24 @@ func DisableVerificationCode(dest string) {
 	}
 }
 
+func CheckSigninCode(user *User, dest, code, lang string) string {
+	// check the login error times
+	if msg := checkSigninErrorTimes(user, lang); msg != "" {
+		return msg
+	}
+
+	result := CheckVerificationCode(dest, code, lang)
```
leo220yuyaodog, line 199: The result is `wrong code!`, can not go to the case `wrongCode`

leo220yuyaodog (commented):
Thanks for your contribution! Please check the code, there are some problem in the pr.

object/verification.go
longit644: Thanks @leo220yuyaodog, I fixed it

hsluoyz:
@longit644 the code is a very long and random string, can it be bursted? Can you give some example about other auth providers do such restriction?

longit644:
Hi @hsluoyz, we only use 6 chars random string with `0-9` charset, so it can be bursted. I not tried with other providers yet, but I got a limit times when I using some application, such as: Slack.
Slack

PR #1462, opened 2023-01-10
feat: optimize UserEditPage & PermissionLIstPage display

SealinGp:
- fix:PermissionList may broken if text undefined
- feat: update applications when organizations changed

Signed-off-by: SealinGp <sealingp@163.com>


SealinGp:
> @SealinGp use the same committer name as github
> 
> image

done

web/src/locales/de/data.json
```diff
@@ -16,6 +16,7 @@
     "Sync policies successfully": "Sync policies successfully"
   },
   "application": {
+    "Empty": "Empty",
```
hsluoyz, line 19: This is wrong, update i18n as: https://casdoor.org/docs/contributing#how-to-update-i18n-data

web/src/UserEditPage.js
```diff
@@ -115,7 +115,17 @@ class UserEditPage extends React.Component {
     return value;
   }
 
+  getEmptyText(text) {
```
hsluoyz, line 118: provide a screenshot of the UI effect

hsluoyz (changes requested):
Don't update `yarn.lock`

web/src/locales/de/data.json
SealinGp: done

web/src/UserEditPage.js
SealinGp: image
image

web/src/locales/de/data.json
```diff
@@ -173,6 +173,7 @@
     "Edit": "Bearbeiten",
```
hsluoyz, line 173: Don't update `yarn.lock`

hsluoyz (changes requested):

web/src/UserEditPage.js
hsluoyz: Don't do this. This code is fragile, and maybe don't work for other languages.

hsluoyz (changes requested):

PR #1466, opened 2023-01-11
feat: disable role and permission update and checks when updating user

leo220yuyaodog:


controllers/user_util.go
```diff
@@ -15,13 +15,14 @@
 package controllers
 
 import (
-	"encoding/json"
+	"reflect"
```
hsluoyz, line 18: Don't use `reflect`

hsluoyz (changes requested):

hsluoyz (approved):

PR #1153, opened 2022-09-22
fix: role update sync grouping policies

tom2nonames:
role update & delete sync grouping policies.

PR #1477, opened 2023-01-14
feat: update permission rule when role updated

notdu:
Continue PR: https://github.com/casdoor/casdoor/pull/1153

### DIFF
--- a/object/permission_enforcer.go
+++ b/object/permission_enforcer.go
@@ -50,26 +50,36 @@ func getEnforcer(permission *Permission) *casbin.Enforcer {
 		panic(err)
 	}
 
+	policyFilter := xormadapter.Filter{}
+
+	if !HasRoleDefinition(m) {
+		policyFilter.Ptype = []string{"p"}
+		err = adapter.LoadFilteredPolicy(m, policyFilter)
+		if err != nil {
+			panic(err)
+		}
+	}
+
 	enforcer, err := casbin.NewEnforcer(m, adapter)
 	if err != nil {
 		panic(err)
 	}
 
 	// load Policy with a specific Permission
-	err = enforcer.LoadFilteredPolicy(xormadapter.Filter{
-		V5: []string{permission.GetId()},
-	})
+	policyFilter.V5 = []string{permission.GetId()}
+	err = enforcer.LoadFilteredPolicy(policyFilter)
 	if err != nil {
 		panic(err)
 	}
 	return enforcer
 }
 
-func getPolicies(permission *Permission) ([][]string, [][]string) {
+func getPolicies(permission *Permission) [][]string {
 	var policies [][]string
-	var groupingPolicies [][]string
+
 	permissionId := permission.GetId()
 	domainExist := len(permission.Domains) > 0
+
 	for _, user := range permission.Users {
 		for _, resource := range permission.Resources {
 			for _, action := range permission.Actions {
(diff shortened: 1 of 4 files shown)


notdu:
continue the PR #1153

hsluoyz:
@leo220yuyaodog plz review

object/permission_enforcer.go
```diff
@@ -72,86 +72,118 @@ m = g(r.sub, p.sub) && r.obj == p.obj && r.act == p.act`
...
 		roleObj := GetRole(role)
+		if roleObj == nil {
+			continue
+		}
+
 		for _, subUser := range roleObj.Users {
 			if domainExist {
 				for _, domain := range permission.Domains {
-					groupingPolicies = append(groupingPolicies, []string{subUser, domain, role, "", "", permissionId})
+					groupingPolicies = append(groupingPolicies, []string{subUser, domain, role, "", "", permissionID})
```
hsluoyz, line 129: Use `permissionId`

hsluoyz (changes requested):

hsluoyz (approved):

hsluoyz:
@notdu plz fix conflicts

image

leo220yuyaodog:
Need to do some compatiable, like these pr #1478 
If use rbac model, the table permission_rule will has g type policy 
image
But if other permission use models that do not have `[role_definition]`, it will cause when init enforce.
image

See the image, the model don't has g type
image

notdu:
Hi @hsluoyz, @leo220yuyaodog,
In Casbin, we don't check if the `assertion` is exist or not when `LoadPolicyArray`, is it ok if i create a PR to fix it.
The problem above will be solved.
Screenshot 2023-01-15 at 6 05 21 PM

hsluoyz:
@notdu if the Casbin model is not a RBAC enabled model, then all `g` rules should be ignored (not loaded into the enforcer)

notdu:
@hsluoyz by adding 
` if assertion := m[sec][key]; assertion == nil { 
	return nil
}`
to `LoadPolicyArray` 
i think we can handle it, or anything wrong with my solution?

hsluoyz:
@notdu it seems to be OK. Can you update the PR so I can see the whole code?

notdu:
@hsluoyz Nothing change on this PR, i'll create a PR on casbin, after that PR is merged, i'll update this PR

hsluoyz:
@notdu the whole change should happen in this repo. Casbin side is not needed to be changed.

leo220yuyaodog:
> @hsluoyz by adding ` if assertion := m[sec][key]; assertion == nil { return nil }` to `LoadPolicyArray` i think we can handle it, or anything wrong with my solution?

https://github.com/casdoor/casdoor/blob/d0741e37057283c67e8f13e4db44b0609f32e8ad/object/permission_enforcer.go#L53
Can add the condition to filter policy of the adapter according to the model text before use the adapter to init the enforce.
```go
if !strings.Contains(permissionModel.ModelText,"[role_definition]"){
	adapter.LoadFilteredPolicy(m, xormadapter.Filter{Ptype: []string{"p"}})
}
```

notdu:
@leo220yuyaodog, @hsluoyz 
plz review my new updates

leo220yuyaodog:
> @leo220yuyaodog, @hsluoyz plz review my new updates

lgtm

Issue #1470, opened 2023-01-12
something mistake with Casbin enforce Api  in permission_enforcer.go


RomenMark:
when I  
```
POST http://localhost:8000/api/enforce
header {Authorization: Basic Base64(clientId:clientSecret)}
Body {
  "id": "OrgName/permissionName",
  "v0": "OrgName/userName",
  "v1": "example-resource",
  "v2": "example-action"
}
```
it return 500
```
invalid policy size: expected 3, got 6, pvals: [organization_yb5iw8/role_i4ug5u 222 read   organization_yb5iw8/permission_erilvx]
```

```
// permission_model  rbac
[request_definition]
r = sub, obj, act

[policy_definition]
p = sub, obj, act

[role_definition]
g = _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = r.sub == p.sub && r.obj == p.obj && r.act == p.act

```
```
//controllers/permission_enforcer.go
func Enforce(permissionRule *PermissionRule) bool {
	permission := GetPermission(permissionRule.Id)    //&{organization_yb5iw8 permission_erilvx 2023-01-11T22:03:45+08:00 New Permission - erilvx [] [organization_yb5iw8/role_i4ug5u] [] model_lpc4m4 dev Application [222] [Read] Allow true admin admin 2023-01-11T22:07:31+08:00 Approved}
	enforcer := getEnforcer(permission)
	request := []interface{}{permissionRule.V0, permissionRule.V1, permissionRule.V2}
	if permissionRule.V3 != "" {
		request = append(request, permissionRule.V3)
	}
       // request...    organization_yb5iw8/user_oioo9w something read  
	allow, err := enforcer.Enforce(request...)   // something wrong 
	if err != nil {
		panic(err)
	}
	return allow
}
```

notdu:
hi @RomenMark,
Try to change your policy_definition from
p = sub, obj, act --> p = sub, obj, act, "", "", ""

PR #1478, opened 2023-01-14
feat: fix compatibility issue between Casbin request and model

leo220yuyaodog:
resolve: #1470

### DIFF
--- a/object/permission.go
+++ b/object/permission.go
@@ -57,6 +57,40 @@ type PermissionRule struct {
 	Id    string `xorm:"varchar(100) index not null default ''" json:"id"`
 }
 
+const (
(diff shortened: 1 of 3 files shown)


object/permission.go
```diff
@@ -57,6 +57,40 @@ type PermissionRule struct {
 	Id    string `xorm:"varchar(100) index not null default ''" json:"id"`
 }
 
+const (
```
leo220yuyaodog, line 60: Now, `permission` default uses table named "permission_rule" as adapter.  `permission` cannot select other adapter. 
The field `builtInAdapter` is reserved to distinguish the other adapter in the futhur, if casdoor implement` permisson` can select exist adapter flexible.

hsluoyz:
@RomenMark @notdu plz review

PR #1490, opened 2023-01-22
feat: support config multiple phone prefixes for organization

### LINKED ISSUE #1397: Allow changing the country code of phone number on login page
Any plan to support multi country code?

hezean:
Resolves #1397

Screenshot 2023-01-23 at 01 55 39
Screenshot 2023-01-23 at 01 56 03

### DIFF
--- a/object/organization.go
+++ b/object/organization.go
@@ -41,7 +41,7 @@ type Organization struct {
 	Favicon            string   `xorm:"varchar(100)" json:"favicon"`
 	PasswordType       string   `xorm:"varchar(100)" json:"passwordType"`
 	PasswordSalt       string   `xorm:"varchar(100)" json:"passwordSalt"`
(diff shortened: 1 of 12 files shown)


hsluoyz:
@hezean communicate with the community first before making a PR to avoid that the code didn't fulfill the requirement

hezean:
Thanks for reminding @hsluoyz, this PR proposes to change two things:
- an organization can configure several phone prefixes that are accepted for its members
- every member store a phone prefix for its phone number, for the data migration, they will use the only phone prefix of the organization before

The above changes were based on a scenario I'm facing, for example, when I'm developing an app for my school, that have international students, I may not want to create several organizations for each identity country they come from, but prefer to let them choose one prefix.

HDYT?

hsluoyz:
This issue makes sense but maybe we will not use this PR directly as it differs a lot with our thoughts. Possibly replaced by: https://github.com/casdoor/casdoor/pull/1524

PR #1491, opened 2023-01-23
Fix API panic error in "/api/enforce" and "/api/batch_enforce"

JhouXerox:
append permissionId on field r to resolve  panic error in casbin package:
invalid request size: expected 3,got 4, rvals:[...]

because of permission.go:69 GetRequest func, always return request object which lenth is at least 4.

before to merge, you can reoccurrence this phenomenon on default db(built-in) to see this case.

### DIFF
--- a/object/permission_enforcer.go
+++ b/object/permission_enforcer.go
@@ -290,8 +290,12 @@ m = g(r.sub, p.sub) && r.obj == p.obj && r.act == p.act`
 		}
 		policyDefinition = append(policyDefinition, "permissionId")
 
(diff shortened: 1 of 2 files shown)


hsluoyz:
@JhouXerox plz modify all your texts into English

JhouXerox:
@hsluoyz already done~

notdu:
Hi @JhouXerox, it's not work right when you login, `CheckAccessPermission` call `Enforce` without `permissionId`, please check it again

PR #1513, opened 2023-02-02
feat: add user password in ldap server search result

### LINKED ISSUE #1508: The cn attribute is inconsistent in ldap server search method.
Hi, 

While I test the ldap server function, it print
```<username> auth test failed: LDAP Result Code 49 "Invalid Credentials": invalid credentials ErrMsg: The user: <orgName>/<DisplayName> doesn't exist```

So I check the code in ldapserver.go, and I found line 108 and 110 assign inconsistent attribute to cn.

https://github.com/casdoor/casdoor/blob/b47baa06e1a75c9a538b8b12acf513f6968bab7d/controllers/ldapserver.go#L106-L116

It cause the error above.

### PR DESCRIPTION
**Pull request by imshenyz**
fix #1508, and add user password in ldap server search result to support authentication in some ldap client like freeradius.

### DIFF
--- a/controllers/ldapserver.go
+++ b/controllers/ldapserver.go
@@ -111,8 +111,28 @@ func handleSearch(w ldapserver.ResponseWriter, m *ldapserver.Message) {
 		e.AddAttribute("uid", message.AttributeValue(user.Name))
 		e.AddAttribute("email", message.AttributeValue(user.Email))
 		e.AddAttribute("mobile", message.AttributeValue(user.Phone))
+		e.AddAttribute("userPassword", message.AttributeValue(getUserPasswordWithType(user)))
 		// e.AddAttribute("postalAddress", message.AttributeValue(user.Address[0]))
 		w.Write(e)
 	}
 	w.Write(res)
+}
+
+// get user password with hash type prefix
+// TODO not handle salt yet
+// @return {md5}5f4dcc3b5aa765d61d8327deb882cf99
+func getUserPasswordWithType(user *object.User) string {
+	org := object.GetOrganizationByUser(user)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by hsluoyz** [comment]
@imshenyz plz fix:

image

File: controllers/ldapserver.go [hunk 1]
```diff
@@ -105,14 +105,34 @@ func handleSearch(w ldapserver.ResponseWriter, m *ldapserver.Message) {
...
+	}
+	prefix := org.PasswordType
+	if prefix == "salt" {
+		prefix = "sha256"
+	} else if prefix == "md5-salt" {
+		prefix = "md5"
+	} else if prefix == "pbkdf2-salt" {
+		prefix = "pbkdf2"
+	}
+	return fmt.Sprintf("{%s}%s", prefix, user.Password)
```
hsluoyz, line 137: Is this format correct? Will it be recognized by LDAP?

hsluoyz (changes requested):

controllers/ldapserver.go
imshenyz: HI, I refer to following pages
[OpenLDAP Password Storage](https://www.openldap.org/doc/admin24/security.html#Password%20Storage) 
[FreeRADIUS Password-With-Header](https://freeradius.org/radiusd/man/rlm_pap.txt)
and tested in FreeRADIUS for the {md5} and {sha256}.

I not sure it is correct  for {pbkdf2}, but I think it is a processing logic of the ldap client.

hsluoyz: seems great!

hsluoyz (approved):

Issue #1496, opened 2023-01-27
add provider icons and menus


nomeguy:
Add icons for Email default and Captcha default

https://door.casdoor.com/providers

![image](https://user-images.githubusercontent.com/85475922/215107618-2658f87e-dae1-4b71-b2cc-9c5cc28b898d.png)

Add missing categories (totally should be 7) in the menus:

![image](https://user-images.githubusercontent.com/85475922/215107697-e682ea85-9dea-4968-91c4-0ee3af709927.png)

![image](https://user-images.githubusercontent.com/85475922/215107784-b3026d65-be2d-4f8c-8b64-15f89c5516a5.png)

GyanPrakashRaj:
hey Maintainer,
It's a good issue to start.
Can you assign it to me?

PR #1522, opened 2023-02-05
feat: add provider icons and menus

LilithHC:
Fix: #1496


web/src/Setting.js
```diff
@@ -779,54 +783,6 @@ export function getProviderTypeOptions(category) {
         {id: "Okta", name: "Okta"},
         {id: "Douyin", name: "Douyin"},
         {id: "Line", name: "Line"},
-        {id: "Amazon", name: "Amazon"},
```
nomeguy, line 782: Don't delete these

nomeguy (changes requested):

nomeguy (approved):

PR #1536, opened 2023-02-10
feat: support refresh token on GetOAuthToken

notdu:
`To refresh an Access Token, the Client MUST authenticate to the Token Endpoint`

ref: https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -172,6 +172,7 @@ func (c *ApiController) GetOAuthCode() {
 // @router /login/oauth/access_token [post]
 func (c *ApiController) GetOAuthToken() {
 	grantType := c.Input().Get("grant_type")
+	refreshToken := c.Input().Get("refresh_token")
 	clientId := c.Input().Get("client_id")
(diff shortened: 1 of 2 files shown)


controllers/token.go
```diff
@@ -172,6 +172,7 @@ func (c *ApiController) GetOAuthCode() {
 // @router /login/oauth/access_token [post]
 func (c *ApiController) GetOAuthToken() {
 	grantType := c.Input().Get("grant_type")
+	refreshToken := c.Input().Get("refresh_token")
```
hsluoyz, line 175: Why pass the token as parameter?

hsluoyz (changes requested):

controllers/token.go
notdu: What's wrong with it? The current refresh token endpoint is the same
https://github.com/casdoor/casdoor/blob/master/controllers/token.go 
line 226

hsluoyz (approved):

Issue #1537, opened 2023-02-10
[frontend] Make edit pages' field column wider when screen width is small


hsluoyz:
All edit pages like: https://door.casdoor.com/applications/built-in/app-built-in

The left column is too small when screen is small. Should extend it to right a little bit.

![image](https://user-images.githubusercontent.com/3787410/218102042-23f2919e-01ab-4fab-afd9-99ec4fad3ca8.png)

HemangVora:
Hey can I work on this issue I think I can able to solve it

AnmolBansalDEV:
I can solve this issue, please assign it to me.

KaifKhan7393:
Please, assign it to me so I can work on it

PR #1542, opened 2023-02-12
[frontend] Make edit pages' field column wider when screen width is small #1537

Aqib-Ali08:


Aqib-Ali08:
This can be one of the solution

PR #1550, opened 2023-02-13
feat: change token ExpiresIn to second

notdu:
The wrong format of ExpiresIn makes the result of casdoorsdk.GetOAuthToken (field `expiry`) wrong, too.

`expires_in`: OPTIONAL. Expiration time of the Access Token in seconds since the response was generated.

ref: https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse

in `3.2.2.5.  Successful Authentication Response`

### DIFF
--- a/object/token.go
+++ b/object/token.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	hourMinutes          = 60
+	hourSeconds          = int(time.Hour / time.Second)
 	InvalidRequest       = "invalid_request"
 	InvalidClient        = "invalid_client"
 	InvalidGrant         = "invalid_grant"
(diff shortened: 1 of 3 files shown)


leo220yuyaodog:
https://github.com/casdoor/casdoor/blob/6beb68dccec812b6794cf7186c4ed7def0cade91/util/time.go#L34
The fucntion `IsTokenExpired()` also need to be update.   `time.Minute` => `time.Second`

hsluoyz:
@leo220yuyaodog you can directly add your comments as review

notdu:
@leo220yuyaodog updated, plz review

PR #1565, opened 2023-02-18
fix: detect access denied by response message for demo mode

imp2002:
judging modification operator by `response msg` in demoMode

### DIFF
--- a/web/src/backend/FetchFilter.js
+++ b/web/src/backend/FetchFilter.js
@@ -21,35 +21,38 @@ import * as Setting from "../Setting";
 const {confirm} = Modal;
 const {fetch: originalFetch} = window;
 
-const demoModeCallback = async(url, option) => {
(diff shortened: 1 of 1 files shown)


web/src/backend/FetchFilter.js
```diff
@@ -21,35 +21,38 @@ import * as Setting from "../Setting";
...
-        Setting.openLink(`https://demo.casdoor.com${location.path}${location.search}?username=built-in/admin&password=123`);
-      },
-      onCancel() {},
-    });
-  }
-
-  return option;
+const demoModeCallback = (res) => {
+  res.json().then(data => {
+    if (data.msg === "Unauthorized operation") {
```
hsluoyz, line 26: This string is i18n, not only English:

image

hsluoyz (changes requested):

PR #1591, opened 2023-02-27
fix: fix bug form country code init error

leo220yuyaodog:
https://user-images.githubusercontent.com/71440988/221573280-3adf07ea-199f-4e9f-ae7f-f2239254d954.mp4

### DIFF
--- a/web/src/auth/SignupPage.js
+++ b/web/src/auth/SignupPage.js
@@ -98,10 +98,6 @@ class SignupPage extends React.Component {
       } else {
         Setting.showMessage("error", `Unknown application name: ${applicationName}`);
       }
-    } else {
-      this.setState({
-        countryCode: this.getApplicationObj().organizationObj.countryCodes?.[0],
-      });
     }
   }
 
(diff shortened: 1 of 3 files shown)


web/src/auth/SignupPage.js
```diff
@@ -555,6 +547,7 @@ class SignupPage extends React.Component {
         initialValues={{
           application: application.name,
           organization: application.organization,
+          countryCode: application.organizationObj.countryCodes?.[0],
```
hsluoyz, line 550: what if `application.organizationObj.countryCodes` is empty array? Define a function to reuse code

web/src/auth/SignupPage.js
```diff
@@ -398,10 +393,7 @@ class SignupPage extends React.Component {
                 ]}
               >
                 <PhoneNumberInput
-                  showSearsh={true}
                   style={{width: "35%"}}
```
hsluoyz, line 396: Missing value and onChange?

hsluoyz (changes requested):

PR #1593, opened 2023-02-28
fix(controller): add check when deleting a user

xtlsoft:


hsluoyz:
@xtlsoft can you explain the PR?

xtlsoft:
Previously no proper permission check was done when deleting a user. I could delete an admin user with a normal user.

hsluoyz:
@xtlsoft usually normal users cannot delete users. All `built-in` org users are global admins, see: https://casdoor.org/docs/basic/core-concepts#how-does-casdoor-manage-itself

xtlsoft:
Thx. I'm so sorry for bothering you.

PR #1595, opened 2023-02-28
feat: support making configs in values.yaml

fengxsong:
remove hardcode in `template/configmap.yaml`.


hsluoyz:
@henrywangx plz review

Issue #1613, opened 2023-03-03
[bug] how to use enforce to detect whether a user has permission for resource.


fengxsong:
I'm working on a demo that use casdoor to implement the authn/authz service.

The data in database are shown below:

```
mysql> select * from permission_rule;
+-------+-------------------------+------------------+-------+----+----+-------------------------------+----+
| ptype | v0                      | v1               | v2    | v3 | v4 | v5                            | id |
+-------+-------------------------+------------------+-------+----+----+-------------------------------+----+
| p     | built-in/*              | app-built-in     | read  |    |    | built-in/permission-built-in  |    |
| p     | built-in/*              | app-built-in     | write |    |    | built-in/permission-built-in  |    |
| p     | built-in/*              | app-built-in     | admin |    |    | built-in/permission-built-in  |    |
| p     | testapp/fengxusong | todo.TodoService | read  |    |    | testapp/todo.TodoService |    |
| p     | testapp/fengxusong | todo.TodoService | write |    |    | testapp/todo.TodoService |    |
| p     | testapp/fengxusong | todo.TodoService | admin |    |    | testapp/todo.TodoService |    |
+-------+-------------------------+------------------+-------+----+----+-------------------------------+----+
mysql> select * from permission;
+--------------+---------------------+---------------------------+---------------------+-----------------------------+-------+---------+----------------+---------+---------------+----------------------+--------------------------+--------+------------+-----------+----------+---------------------------+----------+
| owner        | name                | created_time              | display_name        | users                       | roles | domains | model          | adapter | resource_type | resources            | actions                  | effect | is_enabled | submitter | approver | approve_time              | state    |
+--------------+---------------------+---------------------------+---------------------+-----------------------------+-------+---------+----------------+---------+---------------+----------------------+--------------------------+--------+------------+-----------+----------+---------------------------+----------+
| built-in     | permission-built-in | 2023-03-03T07:30:54Z      | Built-in Permission | ["built-in/*"]              | []    | []      | model-built-in |         | Application   | ["app-built-in"]     | ["Read","Write","Admin"] | Allow  |          1 |           |          |                           |          |
| testapp | todo.TodoService    | 2023-03-03T15:37:46+08:00 | todo service        | ["testapp/fengxusong"] | []    | []      | rbac           |         | Application   | ["todo.TodoService"] | ["Read","Write","Admin"] | Allow  |          1 | admin     | admin    | 2023-03-03T15:38:52+08:00 | Approved |
[...]

PR #1623, opened 2023-03-06
feat: remove ID field in enforcement

imp2002:
Fix: #1613 

Remove extra field to match current `Model`


leo220yuyaodog:
Before init the Enforce, the function `GetBuiltInModel` will add permissionId in model text. Maybe problem cause by here. I'm not it complete compatible all cases. For example, now the `[role_defination]` is not handle in here.
https://github.com/casdoor/casdoor/blob/ed052b0e6a8a121f99c4404932bcaef7c3389384/object/permission_enforcer.go#L257

leo220yuyaodog:
@imp2002 Oh, I know why cause the error. The condition `p.v4 != ""`,  But we need to add the `""` to take up the field, or the size of policy is not match the model defined. 
image

PR #1637, opened 2023-03-10
refactor: set git commit and version tag at compile time

evrins:
在 go 的编译期插入 git 提交信息，解决容器部署时 前端版本信息缺失的问题

还有 QQ 群的号码是什么，casdoor.org 上的 QQ 按钮在 macOS chrome 上没有跳转到 qq for mac


hsluoyz:
@evrins Please use English everywhere.

`build.sh` is only for Docker and Linux. But we have a lot of users who don't use Docker or Linux. Your solution won't work for them.

My general advice is, for non-trivial issues like this, please discuss with maintainers first in GitHub issues or other community channels before making PR.

Issue #1594, opened 2023-02-28
OAuth 2 access token generated in client-credentials pattern cannot be intropected


deljuven:
It seems that access token comes from client-credential mode can not be parsed correctly

hsluoyz:
@deljuven can you give more details?

deljuven:
@hsluoyz 
I call oauth/token with grant_type set as client-credentials, and then intropect API returns some error message showed application not found.
I tried to reproduce the scenario using demo.casdoor.com, but find introspect API failed today. The only thing I can get is
```
{
    "active": false
}
```

deljuven:
Sorry, I think I made some mistake. The API accepts form data only, but I pass the token with json payload.

deljuven:
When I tried to use access token created via client-credentails mode, there is the following error.
<img width="1286" alt="image" src="https://user-images.githubusercontent.com/836399/223133581-1f23525d-6e91-4f29-bde2-5db01888bb6d.png">

hsluoyz:
@deljuven can you make a PR to fix it?

deljuven:
For client-credential mode, username in token is generated with "app/" prefix by default.
<img width="1079" alt="image" src="https://user-images.githubusercontent.com/836399/224220266-28e55338-fcbc-4cec-b5ff-af51cfdcf9d0.png">

PR #1639, opened 2023-03-10
feat: fix name format in application login: GetClientCredentialsToken()

hsluoyz:
Fix: https://github.com/casdoor/casdoor/issues/1594


deljuven:
I think this pr works for client-credential mode.
But here comes another problem, it seems that access token created from client-credential mode and authorization-mode will in the same format after this pr. It seems to be hard to detect a token is issued for a normal user or for an application. Does this still meet the design?

hsluoyz:
@deljuven application can be identified by "type" property now:

image

PR #1651, opened 2023-03-13
feat: can send Aliyun test SMS now

leo220yuyaodog:


web/src/auth/LoginPage.js
```diff
@@ -427,16 +426,15 @@ class LoginPage extends React.Component {
...
+                        if (!Setting.isValidEmail(value) && !Setting.isValidPhone(value)) {
                           this.setState({validEmailOrPhone: false});
                           return Promise.reject(i18next.t("login:The input is not valid Email or Phone!"));
                         }
 
-                        if (Setting.isValidPhone(this.state.username)) {
-                          this.setState({validPhone: true});
-                        }
-                        if (Setting.isValidEmail(this.state.username)) {
+                        if (Setting.isValidEmail(value)) {
```
leo220yuyaodog, line 434: There is bug in old version. When you first type a email, then type a phone. Before send code, the code just use `validEmail` to jungle it is `email` or `phone`. Now, they  are both true and actually a phone, but use email.

hsluoyz: looks good!

Issue #1654, opened 2023-03-14
Health check http path? 


chenjpu:
Is there a monitoring check path that does not generate the cookie?
Discovery Access static resources also generates cookie information on the server.

chenjpu:
server-side callback cas api ```/p3/serviceValidate``` will also generates  useless cookie information

hsluoyz:
@chenjpu why doesn't want cookie?

chenjpu:
Health checks are all server access, do not need this kind of information, regular checks will produce different cookie.

hsluoyz:
@chenjpu you are totally right. But the beego frameworks enables cookie for APIs by default. And we don't have time to pick out which APIs don't need cookie. We will not contribute time on this. If you are willing to make a PR, it's welcome.

PR #1656, opened 2023-03-15
fix: disable cookie for static files

chenjpu:
Fix: https://github.com/casdoor/casdoor/issues/1654


hsluoyz:
@chenjpu this PR has broken all the static paths:

https://github.com/casdoor/casdoor/blob/dd939b5c7e675938d752ec6b6aa9a34e5605d009/main.go#L55-L56

As an example, the Swagger docs doesn't work any more: https://door.casdoor.com/swagger

image

It should be something like:

image

This PR will be reverted.

PR #1670, opened 2023-03-19
feat: remove unused parameter permissionId

notdu:


hsluoyz:
@notdu that arg is kept there for future use. No need to remove for now

PR #1680, opened 2023-03-23
refactor: change all c.Data["json"] = to responseOk() in Go backend

Wrapping-2000:


controllers/provider.go
```diff
@@ -103,8 +103,8 @@ func (c *ApiController) UpdateProvider() {
 		return
 	}
 
-	c.Data["json"] = wrapActionResponse(object.UpdateProvider(id, &provider))
-	c.ServeJSON()
+	response := wrapActionResponse(object.UpdateProvider(id, &provider))
```
leo220yuyaodog, line 106: Do you test the frontend can handle the response successfully? The wrapActionResponse has wraped the data. If ResponseOk rewrap it, frontend may can not handle the response.

Wrapping-2000: Now I have found this problem and I am working on it. Thank you！

UsherFall: > Now I have found this problem and I am working on it. Thank you！

Are you done with this issue？How is it going now

PR #1683, opened 2023-03-24
fix: adjust error result position

leo220yuyaodog:


web/src/App.less
```diff
@@ -125,6 +125,7 @@ img {
 .loginBackground {
   flex: auto;
   display: flex;
+  justify-content: center;
```
hsluoyz, line 128: Will this affect all entry pages, like "left", "right" won't work any more?

image

hsluoyz (changes requested):

hsluoyz (approved):

Issue #1699, opened 2023-03-31
[BUG] /api/get-ldap-users return badRequest when ldap "dn" without "ou"


lzw5399:
# Problem
when using follow ldap admin dn, (by the word, this dn is the default configuration for https://github.com/osixia/docker-openldap)
```
cn=admin,dc=example,dc=org
```

return error message
> LDAP Result Code 34 "Invalid DN Syntax": bind failed ErrMsg: please use correct Admin Name format like cn=xxx,ou=xxx,dc=example,dc=com

then I found the source code, it seems **ou** is mandatory
> https://github.com/casdoor/casdoor/blob/master/object/ldapserver.go#L25
![image](https://user-images.githubusercontent.com/45249999/229003908-10ee8107-2781-4981-a791-eef538f15c34.png)

But **ou** might be optional in dn?

FIY: 
- https://www.google.com.hk/search?q=ldap+is+%22ou%22+mandatory+for+dn&ei=7z8mZLTWII3ThwOF1LsI&ved=0ahUKEwj0idOoioX-AhWN6WEKHQXqDgEQ4dUDCA8&uact=5&oq=ldap+is+%22ou%22+mandatory+for+dn&gs_lcp=Cgxnd3Mtd2l6LXNlcnAQAzIFCAAQogQyBQgAEKIEMgUIABCiBDIFCAAQogQ6CggAEEcQ1gQQsANKBAhBGABQ4gVYrQxgsg9oAXABeACAAYEBiAHnAZIBAzAuMpgBAKABAcgBCMABAQ&sclient=gws-wiz-serp

![image](https://user-images.githubusercontent.com/45249999/229004669-88323a49-c1ff-4734-8a1b-5caa40a1751e.png)

# Reproduce step
## Organization ldap config
![image](https://user-images.githubusercontent.com/45249999/229003021-c9ed7ea0-7ff1-4942-8e28-c39703eeb055.png)

# 

![image](https://user-images.githubusercontent.com/45249999/229003305-ba54ffbf-0e43-4653-ab13-0589c0accc7f.png)

leo220yuyaodog:
@lzw5399 You use Casdoor ldapserver to test. Actually, casdoor ldapserver is used for resove some clients that only support ldapto login by casdoor. So it is not a complete ldap server that support most of the ldap server function.

PR #1705, opened 2023-04-03
feat: fix LDAP server handle filter without CN field as *

leo220yuyaodog:
fix: #1699
1. Default use built-in as ou, the `admin` is used to bind the connect. If not specific the ou, we think it is the global admin. You can see the video below. the user geted from the casdoor server depend on the reserch `Base DN` and `filter`. It is not necessary to specific the ou in `admin`. 
2. The `admin` determine the client authority. For example, the `cn=admin,ou=forum,dc=example,dc=com` means your are using the admin of forum organization. Then, you can just have the authority to get the users in forum organization. If you use `cn=admin,ou=built-in,dc=example,dc=com` or `cn=admin,dc=example,dc=com`,  they are equivalent and means you login as a global admin. Then, you can get all users from casdoor.

https://user-images.githubusercontent.com/71440988/229655168-6e719a94-b62b-4383-b7f4-8c3d54c7ba48.mp4


hsluoyz:
@leo220yuyaodog fix:

image

leo220yuyaodog:
> @leo220yuyaodog fix:
> 
> image

@hsluoyz Fixed. plz review.

PR #1714, opened 2023-04-08
fix: validate parameter and nil in func updateUser

leo220yuyaodog:


controllers/user_util.go
```diff
@@ -17,11 +17,17 @@ package controllers
...
+
 	"github.com/casdoor/casdoor/object"
 )
 
-func checkPermissionForUpdateUser(userId string, newUser object.User, c *ApiController) (bool, string) {
-	oldUser := object.GetUser(userId)
+func checkPermissionForUpdateUser(id string, newUser *object.User, c *ApiController) (bool, string) {
+	oldUser := object.GetUser(id)
+	if oldUser == nil {
+		return false, i18n.Translate(c.GetAcceptLanguage(), "check:User not found")
```
hsluoyz, line 28: Already checked before

controllers/user.go
```diff
@@ -149,7 +153,7 @@ func (c *ApiController) UpdateUser() {
 		return
 	}
 
-	if msg := object.CheckUpdateUser(object.GetUser(id), &user, c.GetAcceptLanguage()); msg != "" {
+	if msg := object.CheckUpdateUser(id, &user, c.GetAcceptLanguage()); msg != "" {
```
hsluoyz, line 156: No need to change param type

hsluoyz (changes requested):

hsluoyz (approved):

PR #1725, opened 2023-04-10
fix: Make OrganizationEditPages' field column wider when screen width is small.

Wrapping-2000:


leo220yuyaodog:
@Wrapping-2000 
1. Plz rebase master onto your branch first.
image
2. In my humble opinion, the pr makes the frontend too complex. Why not just modify the value of `span` in `Col` according the language or simple setting value of `span` more large.

PR #1763, opened 2023-04-20
fix: security verification

cnjinhy:
security verification

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -161,6 +161,7 @@ func (c *ApiController) GetOAuthCode() {
 	// Recommendation:This interface can only be called as an intranet service or intercepted using tools such as iptable/nginx
 	// Or perform session operations after user authorization
 	clientSecret := c.Input().Get("client_secret")
+
 	if clientId == "" || clientSecret == "" {
 		c.ResponseError(c.T("client_id or client_secret required"))
 		return
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -130,6 +130,7 @@ func (c *ApiController) DeleteToken() {
 // @Description get OAuth code
 // @Param   user_id     query    string  true        "The id ( owner/name ) of user"
 // @Param   client_id     query    string  true        "OAuth client id"
+// @Param   client_secret     query    string  true        "OAuth client secret"
 // @Param   response_type     query    string  true        "OAuth response type"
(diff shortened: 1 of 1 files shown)


leo220yuyaodog:
@cnjinhy Hi, thanks for your contribution. But I have some confusion. Even if you get the code,  the code does not contain any information. You need to use the code to exchange accessToken. When your call the api `/login/oauth/access_token`, the `client_secret` is required. So it seems not necessary to verify client_secret` in this api.

cnjinhy:
> @cnjinhy Hi, thanks for your contribution. But I have some confusion. Even if you get the code, the code does not contain any information. You need to use the code to exchange accessToken. When your call the api `/login/oauth/access_token`, the `client_secret` is required. So it seems not necessary to verify client_secret` in this api. Can you provide more details about the Security verification

Anyone who only knows the username and clientId can obtain the **code** of any other user through this interface, and then exchange the **code** for access to the application's callback **address_token** to log in to the application; 
In theory, this interface can only obtain code after user authorization, or be used as an internal network service for server calls, rather than being exposed in the public network

leo220yuyaodog:
@cnjinhy lgtm 
plz fix the ci 
image

cnjinhy:
how to fix the "File is not `gofumpt`-ed (gofumpt)" error? 
I use gofmt , but it not works;

leo220yuyaodog:
> gofumpt

You need to  use [gofumpt](https://github.com/mvdan/gofumpt). It is more stricter than gofmt.

cnjinhy:
why e2e-test failed？ I don‘t change the file

hsluoyz:
Fixed by: https://github.com/casdoor/casdoor/commit/ee8c2650c3e50e95f75574e6d31334a67207db90 (reverts: https://github.com/casdoor/casdoor/commit/54ef2ec09fd3c767da7582f8dd6a90c2b076a96e)

cnjinhy:
> Fixed by: [ee8c265](https://github.com/casdoor/casdoor/commit/ee8c2650c3e50e95f75574e6d31334a67207db90) (reverts: [54ef2ec](https://github.com/casdoor/casdoor/commit/54ef2ec09fd3c767da7582f8dd6a90c2b076a96e))

Remove this API, it cannot be used as a backend for microservices, and it is also not compatible downwards

Issue #1266, opened 2022-11-06
Allow organization admins to manage its own organization


hsluoyz:
Like: https://github.com/casdoor/casdoor/pull/1250

How to reproduce:

1. Log into: https://door.casdoor.com/login/casbin with admin, 123
2. Test which menu item doesn't work, fix the bug. Like the "Providers" tab doesn't work here.

![image](https://user-images.githubusercontent.com/3787410/234626381-a67f2f57-39a0-400c-a5c0-f1c1d902b240.png)

PR #1805, opened 2023-05-04
fix: modify the permissions of org admin

XDTD:
Fix: https://github.com/casdoor/casdoor/issues/1266

Fixed the issue where org admins lacked permission to operate on their organization-related attributes. The issue was caused by `subowner` and `subname` values of the user being the organization name and username respectively, while the default `objowner` and `objname` values on resources such as permission page were "admin" and null.  `Subowner` and `objowner` have different meanings, and it is not clear whether ownership of some resources is attributed to "admin" or the organization.

https://user-images.githubusercontent.com/34300181/236259158-f09f26ea-34a0-4b0e-9405-d0c3a42b4a9c.mp4

Now, permissions is confirmed only if both the user and objowner are admin, but it is still unable to differentiate between admins of different orgs. For example, org1's admin should not have permission to access org2's resources, but it's unable to differentiate between org1's admin and org2's admin. 

To address this issue, I have two approaches:

1. Adopt the RABC model.
2. Standardize `objowner` as organization, since only admin can operate on relevant resources.

Both approaches require significant code changes, but I am currently more inclined towards the second approach.


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/1822

Issue #1777, opened 2023-04-22
[frontend] add lazy load for frontend static files


hsluoyz:
How to reproduce:

1. Go to: https://door.casdoor.com/login
2. Open debug window (F12)
3. Reload page (Ctrl + F5)

4.5MB is too large for single JS:

![image](https://user-images.githubusercontent.com/3787410/235028457-f1c5e7b6-0452-4687-90f2-33d3686b27ac.png)

Similar to: https://github.com/casbin/casnode/pull/319

But the lazy load for Casnode at that time is too heavy, changed a lot of code. We should find a more light-weight way (less code modification) to do it now

PR #1809, opened 2023-05-05
feat: add lazy load for frontend static files

chxclyde:
This PR addresses issue #1777, focusing on reducing the size of the main JS file. By implementing React lazy loading and lazy loading for i18next translation files, the JS file size has been reduced significantly from 4.5MB to 2.0MB.

**Changes Made:**

1. **React Lazy Load**: I applied a similar method as in [casbin/casnode#319](https://github.com/casbin/casnode/pull/319). My approach involves using one large Suspense component to wrap all possible lazy load components, resulting in fewer code changes.

2. **Lazy Load i18next Translation Files**: Following the method in [i18next/Lazy load in memory translations](https://www.i18next.com/how-to/add-or-load-translations#lazy-load-in-memory-translations), I split the translation file `data.json` under `casdoor/web/src/locales` according to namespace. This allowed us to apply the i18next backend for lazy loading, which further reduced the bundle size by approximately 0.5MB.

**Future Optimization Opportunities:**

The bundle analysis using [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) shows that [i18n-iso-countries](https://www.npmjs.com/package/i18n-iso-countries) takes up a considerable size in our bundle. Given that we're only using a small portion of this package's functionality, there might be potential to further reduce our bundle size by splitting that functionality or finding a lighter alternative package.

unnamed

Looking forward to your feedback!


thanhtinhpas1:
@chxclyde It'll be better if we have bundle splitting. It means that whenever user accesses the page, it will load bundle chunks eventually so user doesn't need to wait for too long.

chxclyde:
> 

Thanks for your feedback. I think that is a good idea. Previously I didn't see relevant implementations. Do u have any references about how we could implement that? I will look into that as well.

chxclyde:
> @chxclyde It'll be better if we have bundle splitting. It means that whenever user accesses the page, it will load bundle chunks eventually so user doesn't need to wait for too long.

Thanks for your feedback. I think that is a good idea. I haven't seen revelent methods previously, it will be helpful if you could provide some more references or documentations about this method. I will look into it as well.

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/1944

PR #1826, opened 2023-05-10
feat: AI responses support streaming

1307super:
Is an AI response that supports streaming return

### DIFF
--- a/ai/ai.go
+++ b/ai/ai.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net/http"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


hsluoyz:
@1307super please paste a video here to show that the streaming works

1307super:
> @1307super please paste a video here to show that the streaming works

Since I don't have an openai account, the content is returned by the mock openai interface


https://github.com/casdoor/casdoor/assets/35326304/6f6589af-e22c-4e86-b8a8-3157be413b76

hsluoyz:
@1307super can you add this mock openai interface to the code as a new "type" option of provider? The mock AI provider is useful to the project. Moreover, I cannot see the "streaming" with the real openai API. So I also need to test your mock openai interface.

1307super:
> @1307super the video looks good. Can you add this mock openai interface to the code as a new "type" option of provider? The mock AI provider is useful to the project. Moreover, I cannot see the "streaming" with the real openai API. So I also need to test your mock openai interface.

In fact, the data I simulated was only fixed in the interface and not called the API

Here is the code

```golang
flusher, ok := writer.(http.Flusher)
if !ok {
  return fmt.Errorf("writer does not implement http.Flusher")
}
text := "mock data"
for _, v := range text {
  if _, err := fmt.Fprintf(writer, "data: %s\n\n", string(v)); err != nil {
	  return err
  }
  flusher.Flush()
  time.Sleep(200 * time.Millisecond)
}
```

PR #1841, opened 2023-05-15
feat: User should have PasswordType like Organization

### LINKED ISSUE #1840: [backend] User should have PasswordType like Organization is
# What we have now
1. Lets suppose you have an organization that has some specific password type (plain, salt, etc.).
2. Now you change it (password type)
3. Nowdays after doing so you cannot login after changing Organization's PasswordType because CredManager created using new organization PasswordType which is different from old one

# What should we do

1. We have to use old password type settings (i.e. created when user created) because we dont change all users' passwords after changing organzation's password type
2. This password type settings should be kept alongside each user (add new column to user table or keep it just inside password hash like Apache passwords https://httpd.apache.org/docs/2.4/misc/password_encryptions.html)
3. This user's setttings will be changed later when user decided to change his (her) pasword, and in this case organization password type will be used for it
4. Newly created users should inherit organization's password type

aegoroff:
Fix: #1840

### DIFF
--- a/object/check.go
+++ b/object/check.go
@@ -175,7 +175,7 @@ func CheckPassword(user *User, password string, lang string, options ...bool) st
 		return i18n.Translate(lang, "check:Organization does not exist")
 	}
 
-	credManager := cred.GetCredManager(organization.PasswordType)
+	credManager := cred.GetCredManager(user.PasswordType)
 	if credManager != nil {
 		if organization.MasterPassword != "" {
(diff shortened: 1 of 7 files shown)


hsluoyz:
@leo220yuyaodog plz review

hsluoyz:
@aegoroff plz fix CI:

image

aegoroff:
@hsluoyz made gofmt and pushed fixed version

PR #1907, opened 2023-05-30
fix: Get logger configuration from app.conf file

zabbius:


conf/conf.go
```diff
@@ -140,3 +140,7 @@ func GetConfigRealDataSourceName(driverName string) string {
 	}
 	return dataSourceName
 }
+
+func GetConfigLogs() string {
```
hsluoyz, line 144: just use GetConfigString

hsluoyz (changes requested):

conf/conf.go
zabbius: done

PR #1939, opened 2023-06-06
feat: add lazy loading page

907997375:
fix: #1777
add lazy loading pages to reduce the file size loaded on the first screen
before
The first screen must load 4.4M

after
The first screen only needs to load 2.6M


hsluoyz:
@907997375 why close?

907997375:
I have a better idea, using webpack splitChunks reduce more main.js size.

907997375:
Split the npm package into main.js code, merge the common code,and package it into one file, reducing main.js to less than 1M
The first screen only takes about 2M
I will commit a PR Tomorrow.

PR #1944, opened 2023-06-07
feat: add lazy load page and split npm packages by webpack

907997375:
fix: https://github.com/casdoor/casdoor/issues/1777
This PR is to reduce the first screen loading time by split main.js and lazy loading
1.add lazy loading on demand loading page
2.add webpck configuration to reduing main.js size
before
must load more 4M size

after
only load about 2M size

### DIFF
--- a/web/craco.config.js
+++ b/web/craco.config.js
@@ -1,6 +1,41 @@
 const CracoLessPlugin = require("craco-less");
+const CompressionPlugin = require('compression-webpack-plugin')
 
 module.exports = {
+  webpack: {
+    configure: (webpackConfig) => {
+      if(webpackConfig.mode === "production") {
+        webpackConfig.devtool = false;
+        webpackConfig.optimization = {
+          splitChunks: {
+            minSize: 30,
+            cacheGroups: {
+              default: {
+                name: 'common',
+                chunks: 'initial',
+                minChunks: 2,
+                priority: -20
+              },
+              vendors: {
+                test: /[\\/]node_modules[\\/]/,
+                name: 'vendor',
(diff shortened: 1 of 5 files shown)


hsluoyz:
@907997375 only change webpack, don't change other source files

907997375:
lazy load must change Import method.The original method cannot be loaded async.
craco have been webpapck config .no one is using it and it has always been configured by default.

907997375:
@hsluoyz i'm not sure which file it is
craco.config.js is to modify the webpack configuration in the create-react-app, for modifying webapck without yarn ejet.

hsluoyz:
@leo220yuyaodog plz review

leo220yuyaodog:
@hsluoyz Now without lazy load:
image
Lazy load:
image

The test data is in the same restricted network conditions. It shows that lazy loading does improve the first-screen loading of the landing page.

hsluoyz:
@907997375 can it be smaller? the first JS is less than 1MB would be good

907997375:
The max is an NPM package, reducing it to less than 1m requires the use of CDN

907997375:
@hsluoyz This is opening gzip in my local nginx environment.only less than 1m!
A9K9AAAJ7ZY1J89J3$@TLI9

hardywu:
The problem is getting worse in new versions.  The v1.515.0 has a huge main.js file (7.2 MB),  minutes loading time on my small server. 

image

The latest v1.547.0 has a main.js of 8.0 MB.

hsluoyz:
@hardywu enable gzip by default, it will reduce to 2 MB

Issue #1951, opened 2023-06-09
[frontend] show 404 pages for all non-existent objects in object edit pages


hsluoyz:
Like: 

- https://door.casdoor.com/organizations/xxxx
- https://door.casdoor.com/permissions/casbin/xxx
- https://door.casdoor.com/applications/casbin/xxx

The correct behavior is showing 404 not found error

![image](https://github.com/casdoor/casdoor/assets/3787410/2a7542c7-6b41-4f69-8ff9-a7f3938fe856)

hsluoyz:
Fixed by: https://github.com/casdoor/casdoor/commit/86547f0642a2c4175bd8a342afc57b6eae66091c

PR #1954, opened 2023-06-09
fix：add show 404 pages for all non-existent object in object edit pages

907997375:
fix：#1951
existent whether the data exists when get data.it will show 404 pages if non-existent 
before

after
https://github.com/casdoor/casdoor/assets/82175017/957edb2d-c411-47b0-b12b-9c8075296cc1


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/1955

PR #1955, opened 2023-06-09
feat: add 404 page when non-existent data

907997375:
fix：#1951
existent whether the data exists when get data.it will show 404 pages if non-existent
before

after
https://github.com/casdoor/casdoor/assets/82175017/9065c97b-1178-4e6a-9cd8-6c58ed799bdb

### DIFF
--- a/web/src/ApplicationEditPage.js
+++ b/web/src/ApplicationEditPage.js
@@ -121,7 +121,9 @@ class ApplicationEditPage extends React.Component {
       .then((application) => {
         if (application === null) {
           this.props.history.push("/404");
+          return;
         }
+        
         if (application.grantTypes === null || application.grantTypes === undefined || application.grantTypes.length === 0) {
           application.grantTypes = ["authorization_code"];
         }
--- a/web/src/ApplicationEditPage.js
+++ b/web/src/ApplicationEditPage.js
@@ -119,6 +119,9 @@ class ApplicationEditPage extends React.Component {
   getApplication() {
     ApplicationBackend.getApplication("admin", this.state.applicationName)
       .then((application) => {
+        if (application === null) {
+          this.props.history.push("/404");
+        }
         if (application.grantTypes === null || application.grantTypes === undefined || application.grantTypes.length === 0) {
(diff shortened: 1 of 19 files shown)


web/src/ApplicationEditPage.js
```diff
@@ -119,14 +119,17 @@ class ApplicationEditPage extends React.Component {
   getApplication() {
     ApplicationBackend.getApplication("admin", this.state.applicationName)
       .then((application) => {
-        if (application.grantTypes === null || application.grantTypes === undefined || application.grantTypes.length === 0) {
-          application.grantTypes = ["authorization_code"];
+        if (application === undefined || application === null) {
```
hsluoyz, line 122: Is `application === undefined` needed? If not, remove it

web/src/ApplicationEditPage.js
```diff
@@ -119,14 +119,17 @@ class ApplicationEditPage extends React.Component {
...
+          this.props.history.push("/404");
+        } else {
```
hsluoyz, line 124: Just add the following code (with a trailing new line), use `return` in if, don't add `else` and change other code

```js
if (application === null) {
  this.props.history.push("/404");
  return;
}
```

web/src/ApplicationEditPage.js
```diff
@@ -119,14 +119,17 @@ class ApplicationEditPage extends React.Component {
   getApplication() {
     ApplicationBackend.getApplication("admin", this.state.applicationName)
       .then((application) => {
-        if (application.grantTypes === null || application.grantTypes === undefined || application.grantTypes.length === 0) {
-          application.grantTypes = ["authorization_code"];
+        if (application === undefined || application === null) {
+          this.props.history.push("/404");
```
hsluoyz, line 123: Don't change URL, just show 404 under the current URL like: https://door.casdoor.com/aaa

image

hsluoyz (changes requested):

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/commit/86547f0642a2c4175bd8a342afc57b6eae66091c

PR #1970, opened 2023-06-13
feat: Stop using static salt to hash passwords and start using dynamic one fixes #1919

### LINKED ISSUE #1919: Stop using static salt to hash passwords and start using dynamic one
### Problem
Now to hash passwords using `Sha256SaltCredManager,` `Md5UserSaltCredManager `and `Pbkdf2SaltCredManager` static salt value are used. Aside from that there are some inconsistency among `Sha256SaltCredManager` and `Md5UserSaltCredManager` where `Sha256SaltCredManager` uses organizationSalt to hash password, but `Md5UserSaltCredManager` uses userSalt. i.e.
```
func (cm *Sha256SaltCredManager) GetHashedPassword(password string, userSalt string, organizationSalt string) string {
	res := getSha256HexDigest(password)
	if organizationSalt != "" {
		res = getSha256HexDigest(res + organizationSalt)
	}
	return res
}
```
and
```
func (cm *Md5UserSaltCredManager) GetHashedPassword(password string, userSalt string, organizationSalt string) string {
	res := getMd5HexDigest(password)
	if userSalt != "" {
		res = getMd5HexDigest(res + userSalt)
	}
	return res
}
```
Move on. Because `Sha256SaltCredManager` uses single salt value for all organization's users it provokes security hole because all users with the same password will have the same hash and it'll make possible dictionary or rainbow tables attacks.

### Possible solution
It would be nice to have random salt (only user salt, organization salt must be removed) which changed each time user change his(her) password.

Additional information in Wikipedia https://en.wikipedia.org/wiki/Salt_(cryptography)

aegoroff:
feat: Stop using static salt to hash passwords and start using dynamic one

fixes #1919


hsluoyz:
This is not backward-compatible. Code is complex. Don't see a strong requirement for this. Closed for now.

PR #1978, opened 2023-06-16
feat: get all role/permission of an user

notdu:
The current function doesn't get all permissions and roles of an user, there are still related roles and related permissions (with sub-roles supported)

### DIFF
--- a/object/permission.go
+++ b/object/permission.go
@@ -264,18 +264,48 @@ func DeletePermission(permission *Permission) (bool, error) {
 	return affected != 0, nil
 }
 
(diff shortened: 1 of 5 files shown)


ISulimanov:
Looks likes we faced with same issue. 
If we use RBAC model where permissions belong to role - 

Not clear how to get information about all users permissions. 

Before this changes:
https://github.com/casdoor/casdoor/commit/776a682fae5a9f8f749f2dcf2080cd470c585ae0

I thought it could be resolved by batch-enforce 

but now not clear how to do it in proper way (multiple enforce requests looks like not a best idea)

PR #1986, opened 2023-06-18
fix: set S3ForcePathStyle to true by default

XDTD:


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/1988

PR #1997, opened 2023-06-20
feat: optimize batch-enforce

notdu:
Everytime load enforcer is really time-consuming, instead, we only need load enforcer one time with permissions have the same model and adapter, add list permissionIds as the filter when load enforcer.

### DIFF
--- a/controllers/enforcer.go
+++ b/controllers/enforcer.go
@@ -135,8 +135,20 @@ func (c *ApiController) BatchEnforce() {
 	}
 
 	res := [][]bool{}
+	listPermissionIdMap := map[string][]string{}
+
 	for _, permission := range permissions {
-		enforceResult, err := object.BatchEnforce(permission.GetId(), &requests)
+		key := permission.Model + permission.Adapter
+		permissionIds, ok := listPermissionIdMap[key]
+		if !ok {
+			listPermissionIdMap[key] = []string{permission.GetId()}
+		} else {
+			listPermissionIdMap[key] = append(permissionIds, permission.GetId())
+		}
+	}
+
+	for _, permissionIds := range listPermissionIdMap {
+		enforceResult, err := object.BatchEnforce(permissionIds[0], &requests, permissionIds...)
 		if err != nil {
 			c.ResponseError(err.Error())
 			return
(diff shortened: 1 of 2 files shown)


controllers/enforcer.go
```diff
@@ -135,15 +135,27 @@ func (c *ApiController) BatchEnforce() {
 	}
 
 	res := [][]bool{}
+	listPermissionIdMap := map[string][]string{}
+
 	for _, permission := range permissions {
-		enforceResult, err := object.BatchEnforce(permission.GetId(), &requests)
+		permissionIds, ok := listPermissionIdMap[permission.Adapter]
+		if !ok {
+			listPermissionIdMap[permission.Adapter] = []string{permission.GetId()}
```
hsluoyz, line 143: Why only use adapter as map key? Why not include model too? If model is different, will the enforcer work the same?

object/permission_enforcer.go
```diff
@@ -251,7 +256,7 @@ func Enforce(permissionId string, request *CasbinRequest) (bool, error) {
 	return enforcer.Enforce(*request...)
 }
 
-func BatchEnforce(permissionId string, requests *[]CasbinRequest) ([]bool, error) {
+func BatchEnforce(permissionId string, requests *[]CasbinRequest, permissionIds ...string) ([]bool, error) {
```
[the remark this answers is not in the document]
notdu: for get all related permission_rule with permissionIds when in it enforcer, you can see the implement in getEnforcer

controllers/enforcer.go
notdu: Because we get these permissions by the modelId, so it's surely the same model

object/permission_enforcer.go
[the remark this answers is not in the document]
notdu: Every time init the enforcer is really time-consuming if our system large, especially we have many permissions. We have to optimize the batch enforce process by utilize the permission_rule filter when init enforce, just init the enforcer one time

controllers/enforcer.go
hsluoyz: But `modelId` is only one of all the ways, it's optional. You need to make sure other ways work too

notdu: @hsluoyz okay, fixed it

notdu:
hi @hsluoyz, please review it

hsluoyz (approved):

PR #1998, opened 2023-06-20
fix: fix user information sync issue

### LINKED ISSUE #1933: When using synchronizer, user updates cannot be synchronized. Why?
For example, when I updated the user password, the password was not synchronized to casdoor. I saw the updated log, but the data in the database did not change.
image
image

XDTD:
fix #1933


hsluoyz:
@EthanY92 plz review

PR #2005, opened 2023-06-21
feat: optimize chunk

thanhtinhpas1:
Fix: https://github.com/casdoor/casdoor/issues/1777

Relative this issue [1777](https://github.com/casdoor/casdoor/issues/1777) this PR will lazy load many components in Home page, split chunks, tree shaking, and separate chunks which are used in Home out of Login page.

**First loading**:
normal mode, total first loading ~3.1MB:
image

gzip mode, total first loading ~800KiB:
image

The structure of chunks right now will be:
- vendors.[hashcontent:8].js: all things related to node_modules like react, react-router-dom
- main.[hashcontent:8].js: all things related to components
So with vendors, after first loading will never be re-load again.

```javscript
    configure: (webpackConfig, {env}) => {
      if (env === "production") {
        webpackConfig.devtool = false;
        // Split chunks to separate vendor dependencies from application code
        webpackConfig.optimization.splitChunks = {
          cacheGroups: {
            vendors: {
              test: /[\\/]node_modules[\\/]/,
              name: "vendors",
              priority: -10,
              enforce: true,
              maxSize: 1000000, // 1MB
              minSize: 0, // Set minimum size to 0,
              chunks: "initial", // Split initial chunks
            },
            antdTokenPreviewer: {
              test: /[\\/]node_modules[\\/](antd-token-previewer)[\\/]/,
              name: "antd-token-previewer",
              priority: -5,
              enforce: true,
              chunks: "all",
            },
            codeMirror: {
              test: /[\\/]node_modules[\\/](codemirror)[\\/]/,
              name: "codemirror",
              priority: -5,
              enforce: true,
              chunks: "all",
            },
          },
        };

        // Enable tree shaking
        webpackConfig.optimization.usedExports = true;

        webpackConfig.plugins.push(new CompressionPlugin({
          algorithm: "gzip",
          test: /\.js$|\.css$|\.html$/,
          threshold: 10240,
          minRatio: 0.8
        }))
      }

      return webpackConfig;
    }
```

Some big chunks like` antd-token-previewer`, `codemirror` also be splitted, just loading when needed as example:
image

Grapth:
image


thanhtinhpas1:
Please take a look @hsluoyz

hsluoyz:
@thanhtinhpas1 so for the first load of login page, the largest file is 741KB now?

thanhtinhpas1:
> @thanhtinhpas1 so for the first load of login page, the largest file is 741KB now?

Yes it is.

> Do you have a demo site to let me try?
Unfortunately no, I can but it might take time, can you just try in your local.

907997375:
@thanhtinhpas1 Show the packaged JS file. There are no changes after packaging, but the cacheGroup has added more sizes instead. The default file now has compressed code and does not require a terser-webpack-plugin

907997375:
@thanhtinhpas1 The more files separated, the better. Because there are other NPM packages that require them, the size has been increased. You can personally compare the packaged files, and TerserPlugin is unnecessary. By default, it has been compressed, so there is no need to compress it twice. You can deploy it locally, and I guarantee it will require more than 2m

thanhtinhpas1:
> @thanhtinhpas1 Because there are other NPM packages that require them, the size has been increased. You can personally compare the packaged files, and TerserPlugin is unnecessary. By default, it has been compressed, so there is no need to compress it twice. You can run it local, and I guarantee it will require more than 2m

image

Here is the block code of `craco`, the `webpackConfig` was modified by `webpackConfig` function inside `craco.config.js` and using [webpack-merge](https://www.npmjs.com/package/webpack-merge) lib to merge it with the default one. So the final we only have one `webpackConfig` instance, so the point `compress it twice` was not right. But maybe TerserPlugin was redundant as you said.

And the point `increase size` was right, but for fast loading, we need to split the bundle into many chunks, we need to provide information for each chunk so they can concat into a bundle easily. We totally disable it but we will face risks if do it.
image

About 700KiB I said above, it was in gzip mode, who using Casdoor can enable it to reduce bandwidth, but as an open-source, we need just to give them an option, not forcing. The real size currently is 2MB as you said, I will split it smaller.

After update:
image

thanhtinhpas1:
@hsluoyz I updated PR based on a suggestion from @907997375. Currently, the biggest chunk when first loading is 1MB. Go into Home Page or deeper, users will load more chunks if needed. But their size will always be less than 250 KiB.

Please take a look. Thanks.

web/package.json
```diff
@@ -42,7 +42,7 @@
   },
   "scripts": {
     "start": "cross-env PORT=7001 craco start",
-    "build": "craco build",
+    "build": "NODE_ENV=production craco build",
```
hsluoyz, line 45: 1. Why add this? (debug v.s. prod?)
2. Does this work for Windows too?

web/craco.config.js
```diff
@@ -1,6 +1,48 @@
 const CracoLessPlugin = require("craco-less");
 
 module.exports = {
+  webpack: {
```
hsluoyz, line 4: Does this modify any existing work flow?

Do people need to enable gzip manually?

hsluoyz (changes requested):

web/package.json
thanhtinhpas1: `craco` uses `NODE_ENV` to recognize environments, but it is in production mode when NODE_ENV is not set,  so I think we can remove it.

thanhtinhpas1: It's redundant.

web/craco.config.js
thanhtinhpas1: This webpack configuration will be merged with the default webpack by using [webpack-merge](https://www.npmjs.com/package/webpack-merge). So we just modify the optimization phase, the rest will use the default of `craco`.

thanhtinhpas1: About gzip, we need support from browsers to decompress, but with the size of our chunks, I think we are still fine.

But if we still want it, we must add this plugin [compression-webpack-plugin](https://webpack.js.org/plugins/compression-webpack-plugin/) into webpack config.

hsluoyz:
@thanhtinhpas1 how does this PR change the CDN deploy flow? https://casdoor.org/docs/deployment/deploy-cdn

thanhtinhpas1:
> @thanhtinhpas1 how does this PR change the CDN deploy flow? https://casdoor.org/docs/deployment/deploy-cdn

Nothing changes. CDN deploy just push all files in `css` and `js` folders to CDN and replace path `/static/` of scripts to domain of CDN by using below code:
```golang
html = strings.Replace(html, "\"/static/", fmt.Sprintf("\"%s", domainPath), -1)
```

image

907997375:
@thanhtinhpas1
To optimize based on actual projects,Some configurations are not optimal solutions
1.Failure to achieve optimal on-demand splitting of chunks during first loading.You only forcibly split the vendor into multiple chunks.Some functions with different user permissions are not used at all
2.“defer” src=”path“，path is not immutable
This is the result of me and you enabling gizp mode.No matter how splitchunks, you will need more resources to load for the first loading.
My result:
9e12266ca97e2efc8d87a73681aa566
Your result:
52d2927e8e83784889191e96c2de15c

thanhtinhpas1:
> @thanhtinhpas1 To optimize based on actual projects,Some configurations are not optimal solutions 1.Failure to achieve optimal on-demand splitting of chunks during first loading.You only forcibly split the vendor into multiple chunks.Some functions with different user permissions are not used at all 2.“defer” src=”path“，path is not immutable This is the result of me and you enabling gizp mode.No matter how splitchunks, you will need more resources to load for the first loading. 

Thanks for your suggestions. Although don't know how your configuration work, I suppose that the reason is the `minSize > 500 KiB` limitation can lead to many chunks being split, and it increases the size. So I will remove it, hence we just have one limitation `maxSize <= 1MB`.

Note that the main point here I think not the size of bundles, that depends on what kind of bundle users need to load. e.g vendors is something like `react, react-dom` they are required.

Result after remove `minSize`:
total about ~3.1MB

image

Result after remove `minSize` in **Gzip**:
total about ~800KiB
image

**Bundle analyze**:
image

thanhtinhpas1:
@hsluoyz I updated based on a suggestion from @907997375. The total size already updates in description.
Also casdoor backend has the option to enable gzip mode:
`enableGzip = true`
I enabled as default. Pls take a look.

907997375:
@thanhtinhpas1 
1.This is not caused by 'minsize', but chunks: 'initial'.
2.Although the file size of excessive split chunks has decreased, it has not taken into account that the server still needs to carry more request connections. The first load still requires the same size requests, and more requests need to be made for very small chunks. The optimization speed brought by very small chunks is not as good as the side effect of increasing network requests. This is my previous PR#1944, but your current PR is becoming more and more similar. I hope you have better ideas and don't stick to split chunks.

thanhtinhpas1:
> @thanhtinhpas1 1.This is not caused by 'minsize', but chunks: 'initial'. 2.Although the file size of excessive split chunks has decreased, it has not taken into account that the server still needs to carry more request connections. The first load still requires the same size requests, and more requests need to be made for very small chunks. The optimization speed brought by very small chunks is not as good as the side effect of increasing network requests. This is my previous PR#1944, but your current PR is becoming more and more similar. I hope you have better ideas and don't stick to split chunks.

Learn many things from your comments. Thanks for your support and sorry for duplicate with your PR :cry:

907997375:
@thanhtinhpas1 
I also got some inspiration from your PR, but webpack has many default optimization configurations, so we don't need to go further. Minsize does not require an extreme setting of 0, which can result in too many requests increasing the server's load. Chunksplit is already very small now, and I hope you have more good ideas in other aspects

thanhtinhpas1:
> @thanhtinhpas1 I also got some inspiration from your PR, but webpack has many default optimization configurations, so we don't need to go further. Minsize does not require an extreme setting of 0, which can result in too many requests increasing the server's load. Chunksplit is already very small now, and I hope you have more good ideas in other aspects

Sure, `minsize` does not require to equal 0. After increase I saw the number of request also reduce. Thanks @hsluoyz

thanhtinhpas1:
@hsluoyz Pls take a look.

907997375:
@thanhtinhpas1 
This is the result of your PR and my previous PR
My:
5a183283a6be55ecdc5923964cef754
Your:
e6a5f911a1d0bcd64590ad9a135094a
What we need is the optimal solution. Your PR is the same as the one I modified earlier, except for the default option of webpack. This is unnecessary. Do you have any other aspects to optimize.

PR #2012, opened 2023-06-23
feat: handle error when permission not found

notdu:
The current code not return error when permission not found, which cause panic when getEnforcer with a nil permission

### DIFF
--- a/controllers/enforcer.go
+++ b/controllers/enforcer.go
@@ -44,14 +44,26 @@ func (c *ApiController) Enforce() {
 	}
 
 	if permissionId != "" {
-		enforceResult, err := object.Enforce(permissionId, &request)
+		permission, err := object.GetPermission(permissionId)
 		if err != nil {
 			c.ResponseError(err.Error())
 			return
 		}
 
 		res := []bool{}
-		res = append(res, enforceResult)
+
+		if permission == nil {
+			res = append(res, false)
+		} else {
+			enforceResult, err := object.Enforce(permission, &request)
+			if err != nil {
+				c.ResponseError(err.Error())
+				return
+			}
+
+			res = append(res, enforceResult)
+		}
+
 		c.ResponseOk(res)
 		return
 	}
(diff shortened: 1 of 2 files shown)


object/permission.go
```diff
@@ -107,7 +108,7 @@ func getPermission(owner string, name string) (*Permission, error) {
 	if existed {
 		return &permission, nil
 	} else {
-		return nil, nil
+		return nil, errors.New("permission not found")
```
hsluoyz, line 111: Doesn't exist is not an error. See all other objects. It will break a lot of things. Modify the logic more outside if needed

hsluoyz (changes requested):

object/permission.go
notdu: so it should be return false when not found a permission?

hsluoyz: Do you search all occurences of such cases? The semantics now is that "if something doesn't exist, it doesn't need to be an error, maybe just doesn't exist".

image

If you change that semantics in one place, then you should change all occurences. And changing all without adapting other parts of code like frontend will break the system, the UI.

hsluoyz: If you are only fixing for your encountered issue at Enforce() or BatchEnforce() API in controllers, then just move your modification to controllers, don't change in object package.

notdu: @hsluoyz please review

controllers/enforcer.go
```diff
@@ -79,7 +91,13 @@ func (c *ApiController) Enforce() {
 
 	listPermissionIdMap := object.GroupPermissionsByModelAdapter(permissions)
 	for _, permissionIds := range listPermissionIdMap {
-		enforceResult, err := object.Enforce(permissionIds[0], &request, permissionIds...)
+		firstPermission, err := object.GetPermission(permissionIds[0])
```
hsluoyz, line 94: firstPermission is nil?

controllers/enforcer.go
```diff
@@ -141,7 +177,13 @@ func (c *ApiController) BatchEnforce() {
 
 	listPermissionIdMap := object.GroupPermissionsByModelAdapter(permissions)
 	for _, permissionIds := range listPermissionIdMap {
-		enforceResult, err := object.BatchEnforce(permissionIds[0], &requests, permissionIds...)
+		firstPermission, err := object.GetPermission(permissionIds[0])
```
hsluoyz, line 180: firstPermission is nil?

hsluoyz (changes requested):

controllers/enforcer.go
notdu: it cannot be nil because we get the `permissionIds` it from existed permissions

hsluoyz (approved):

PR #2030, opened 2023-06-29
fix: refactor the idp and regex code

leo220yuyaodog:


util/validation.go
```diff
@@ -70,3 +76,7 @@ func GetCountryCode(prefix string, phone string) (string, error) {
 
 	return countryCode, nil
 }
+
+func FilterField(field string) bool {
```
hsluoyz, line 80: Why old func is not removed? 
image

util/validation.go
```diff
@@ -22,10 +22,16 @@ import (
...
-var rePhone *regexp.Regexp
+var (
+	rePhone          *regexp.Regexp
+	ReWhiteSpace     *regexp.Regexp
+	reFieldWhiteList *regexp.Regexp
+)
 
 func init() {
 	rePhone, _ = regexp.Compile(`(\d{3})\d*(\d{4})`)
+	ReWhiteSpace, _ = regexp.Compile(`\s`)
```
hsluoyz, line 33: ReWhiteSpace is not used

hsluoyz (changes requested):

PR #2042, opened 2023-07-05
fix: Removed Masked of User of inside webhooks

gtjadsonsantos:


object/record.go
```diff
@@ -161,7 +161,7 @@ func SendWebhooks(record *Record) error {
 
 		if matched {
 			if webhook.IsUserExtended {
-				user, err := GetMaskedUser(getUser(record.Organization, record.User))
+				user, err := getUser(record.Organization, record.User)
```
hsluoyz, line 164: Why change this? What fields do you want to expose? Will this cause password exposure?

hsluoyz (changes requested):

PR #2043, opened 2023-07-06
fix: /api/upload-resource panics when parsing file_type

WinterOfBit:
Mentioned on [Discord](https://discord.com/channels/1022748306096537660/1065201965883199518/1126577023150723122)

Encountered when working on [Casibase#3](https://github.com/casbin/casibase/issues/3)


util/string.go
```diff
@@ -304,3 +305,18 @@ func ParseIdToString(input interface{}) (string, error) {
 		return "", fmt.Errorf("unsupported id type: %T", input)
 	}
 }
+
+func ParseFileType(contentType, fileName string) string {
+	if contentType == "" || !strings.Contains(contentType, "/") {
```
hsluoyz, line 310: Use the original code, don't re-implement it

hsluoyz (changes requested):

util/string.go
WinterOfBit: @hsluoyz previous commit reverted. Now use appending "/" to avoid panics and try to minimize changes.

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/2046

PR #2069, opened 2023-07-12
feat: ability to specify CA certificates for LDAP

ovsinc:
LDAP servers are often hosted on an organization's internal network. And access to the LDAP server is open only via the LDAP protocol over SSL (LDAPS).

In this case, there may be a problem of the LDAP client not trusting the server. The X.509 certificate trust chain will not be built.

You need to specify trusted CA certificates manually.

This PR offers a solution to this problem.


hsluoyz:
@ovsinc store cert in the "Certs" tab: https://door.casdoor.com/certs , see how applications and payment providers use it

ovsinc:
Did I understand you correctly, you suggest to bind to the "cert" object instead of the CA certificate input field?
But as far as I can see, the "cert" object is intended for security settings: jwt, generation of the rsa key and a derived server certificate. This multitude of possibilities and fields can confuse the administrator. 
The CA certificate is downloaded by the administrator, not generated. And most likely this CA certificate will be used only on the LDAP configuration form.
I can suggest and implement an alternative - add a new form "CA cert". It will implement the ability to download CA certificates.
@hsluoyz

hsluoyz:
@ovsinc 

1. All different kinds of certs go to the "Certs" tab.
2. The cert content can be large, so will not be carried in a provider get or list API call.
3. The cert can be shared by multiple LDAP settings, so it's still useful for separate storage

ovsinc:
@hsluoyz Thanks for the reply.
I agree with your arguments. But, it still seems to me that the "certs" form is overloaded with objects of different purposes.
Do you mind if I rework the "certs" form in the next PR? For example, the form can be divided into Tabs by purpose: server certificate and private key; jwt settings; CA certificates.
In this form, the use of the "certs" object looks convenient.

hsluoyz:
@ovsinc you can define a new scope or type in: https://door.casdoor.com/certs . Different options can also control which UI elements to be visible in the cert edit page

image

ovsinc:
Excellent. I will take advantage of your suggestion: I will add a new scope and UI behavior on the "certs" form.
When it's ready, I'll write.
@hsluoyz

object/cert.go
```diff
@@ -37,6 +87,8 @@ type Cert struct {
 	PrivateKey             string `xorm:"mediumtext" json:"privateKey"`
 	AuthorityPublicKey     string `xorm:"mediumtext" json:"authorityPublicKey"`
 	AuthorityRootPublicKey string `xorm:"mediumtext" json:"authorityRootPublicKey"`
+
+	CACertificate string `xorm:"mediumtext" json:"caCertificate"`
```
hsluoyz, line 91: Why not use `Certificate`?

ovsinc:
I have added CA certificate support in `LDAP`. Added a CACert certificate scope.
image

On the `LDAP` form, a certificate can only be used with a CACert scope.
image

I had to expand the scope of work a little.

Only a certificate with a JWT scope can be used on the `Application` form. Added a scope filter.
image

In addition, I added backend validation on the `Certs` form:
* the certificate and the RSA private key are validated;
* the CA certificate is validated.

Screenshot_20230714_183548

@hsluoyz

hsluoyz (changes requested):
Your changes are too large. Just store the data in "Certs" tab and use it in your scenario, control the influcence inside the LDAP CA stuff.

"CACert" -> "LDAP", CACert is not a good name

Don't do extra validation/checking, as the code is already too complicated, so much maintain burden

ovsinc:
I can't answer your comment in object/cert.go
I'll answer here.
You can switch the scope to JWT and then the certificate and key will be incompatible.
This can lead to errors that are difficult to detect.

object/cert.go
ovsinc: You can switch the "scope" to JWT and then the certificate and key will be incompatible.
This can lead to errors that are difficult to detect.

hsluoyz: That's the wrong scope's problem, not a reason to add a duplicated field

yamaritta: @hsluoyz but the ca certificate is used to verify ssl connection, not to sign data. If you are using  certificate scope to set up ssl ldap fill unused fields. It will not be obvious to the user

Issue #2055, opened 2023-07-11
request a more detailed description of ldap in swagger


yamaritta:
Hello!
The swagger documentation does not contain a description of the parameters and results for ldap methods.
Please add this
![image](https://github.com/casdoor/casdoor/assets/41572091/c2a3ce06-a527-4e20-a8b0-8d67d05737da)

PR #2094, opened 2023-07-19
fix: provide detailed description of ldap in swagger

uestc-wxy:
Fix: https://github.com/casdoor/casdoor/issues/2055

As the swagger documentation does not contain a description of the parameters and results for ldap methods,
I add the details.For example:
```go
// GetLdapUsers
// @Title GetLdapser
// @Tag Account API
// @Description get ldap users
// Param	id	string	true	"id"
// @Success 200 {object} LdapResp The Response object
// @router /get-ldap-users [get]
func (c *ApiController) GetLdapUsers() {
         ......
}
```

### DIFF
--- a/controllers/ldap.go
+++ b/controllers/ldap.go
@@ -38,8 +38,11 @@ type LdapSyncResp struct {
 }
 
 // GetLdapUsers
-// @Tag Account API
 // @Title GetLdapser
+// @Tag Account API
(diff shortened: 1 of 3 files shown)


ISulimanov:
looks like related to https://github.com/casdoor/casdoor/issues/2055

Issue #1833, opened 2023-05-12
Force user to change password after login


ISulimanov:
As administrator I wont to be able force specific user to change password.

As a option it could be boolean prop in user like - passwordChangesRequired. 
If true - than after first login - user has to change his password. 

Is there kind of feature in roadmap? 
Or it should be implemented on client side ?

hsluoyz:
@ISulimanov sounds OK. Can you make a PR to add it?

ISulimanov:
sure, we will send PR in 1-2 weeks

PR #2096, opened 2023-07-19
feat: force user to change password

ISulimanov:
fix: https://github.com/casdoor/casdoor/issues/1833

https://github.com/casdoor/casdoor/assets/7904096/1be405e6-ba3a-46f3-9948-f1b19095faf9

### Few words about solution:

There are differences in implementation when compared with Keycloak: 
- We allow to set `changePasswordRequired` without without changing the user's password. It gives more flexibility.  
- We downgrade user's roles and permissions when  `changePasswordRequired: true`.  It's more secure from Casdoor admin perspective and good for machine-to-machine scenario. 

We provide migration for new account item. Can be removed if not really needed. 

Tested:

- [x] Global admins 
- [x] Org admins
- [x] Normal users

This functionality is really crucial for us and we looking forward on positive feedback :)

Thx.


ISulimanov:
@halozhy all All committers have signed the CLA.
plz review

hsluoyz:
@ISulimanov this code adds a lot of complexity to the code base for such a small feature. Maybe 20% complexity up for 0.3% more features

Think it in another way, you can also choose to fork it and have your complex features in the fork. If you find it's difficult to rebase on top of upstream, then you know this code change is big and it's also a burden to this repo's maintainers.

Issue #1641, opened 2023-03-11
[refactor] change all `c.Data["json"] =` to `responseOk()` in Go backend


hsluoyz:
![image](https://user-images.githubusercontent.com/3787410/224469288-9f03db70-ee9f-4704-a107-4ffe45546519.png)

hsluoyz:
@thanhtinhpas1 hi, can you also work on this as the next step of: https://github.com/casdoor/casdoor/pull/1993 ?

thanhtinhpas1:
> @thanhtinhpas1 hi, can you also work on this as the next step of: https://github.com/casdoor/casdoor/pull/1993 ?

Sure, I'm just busy yesterday, will start tomorrow.

PR #2036, opened 2023-07-03
feat: replace response json by responseOk

thanhtinhpas1:
Fix issue [1641](https://github.com/casdoor/casdoor/issues/1641)

PR #1836, opened 2023-05-14
feat: return most backend API errors to frontend

yehong-z:
Fix: https://github.com/casdoor/casdoor/issues/1831

PR #2098, opened 2023-07-20
feat: refactor the code to use responseOk and add response filter

leo220yuyaodog:
fix: #1641
replace: https://github.com/casdoor/casdoor/pull/2036

1. For all GET request, use responseOk to return format response to frontend. And  use `res.data` to replace `data`
2. Imporve  fetch filter. For request urls prefixed with `get-`, handle error and 404 in the filter.
3. Transfer fetch data `res => res.json()` in the filter
4. Modified the function about captcha that does not match the current code style.
5. Add `code` in the Response, for the frontend judge the denyRequest status easily .


controllers/application.go
```diff
@@ -71,7 +68,7 @@ func (c *ApiController) GetApplications() {
 			return
 		}
 
-		applications := object.GetMaskedApplications(app, userId)
+		applications, _ := object.GetMaskedApplications(app)
```
hsluoyz, line 71: handle error

controllers/application.go
```diff
@@ -175,7 +175,7 @@ func (c *ApiController) GetOrganizationApplications() {
...
```
hsluoyz, line 178: handle error

controllers/auth.go
```diff
@@ -194,7 +194,7 @@ func (c *ApiController) GetApplicationLogin() {
 		return
 	}
 
-	application = object.GetMaskedApplication(application, "")
+	application, _ = object.GetMaskedApplication(application)
```
hsluoyz, line 197: handle error

controllers/organization.go
```diff
@@ -191,7 +189,7 @@ func (c *ApiController) GetDefaultApplication() {
 		return
 	}
 
-	maskedApplication := object.GetMaskedApplication(application, userId)
+	maskedApplication, _ := object.GetMaskedApplication(application)
```
hsluoyz, line 192: handle error

web/src/App.js
```diff
@@ -16,14 +16,15 @@ import React, {Component} from "react";
 import "./App.less";
 import {Helmet} from "react-helmet";
 import EnableMfaNotification from "./common/notifaction/EnableMfaNotification";
+import NotFindResult from "./common/result/NotFindResult";
```
hsluoyz, line 19: NotFoundResult

web/src/common/result/NotFindResult.js
```diff
@@ -0,0 +1,23 @@
+import {Button, Result} from "antd";
```
hsluoyz, line 1: Apache header

web/src/common/result/UnauthorizedResult.js
```diff
@@ -0,0 +1,14 @@
+import {Button, Result} from "antd";
```
hsluoyz, line 1: Apache header

hsluoyz (changes requested):

controllers/application.go
leo220yuyaodog: The GetMask***() after  #1836, use ...error as the last parameter. If you don't pass the err when call the function, it will not return error that not equal nil. I'm not sure if it is a good design. But in fact, now similar uses are redundant and repetitive handling err.

hsluoyz (approved):

Issue #2102, opened 2023-07-21
The 'effect' of the 'permission' is not taking effect after being set.


aosolao:
![Snipaste_2023-07-21_14-56-01](https://github.com/casdoor/casdoor/assets/30789071/62fa0d28-1bf0-4d27-8e9b-c744005a1f70)
![Snipaste_2023-07-21_15-01-00](https://github.com/casdoor/casdoor/assets/30789071/f11b2716-c31f-4743-9a8a-0f3be4bf557b)

#2092 

The problem has still not been resolved.

1. When specific "Sub roles" are configured in "permissions" and "Sub users" is left empty, the "v3" field will still not write "allow" or "deny".
   
2. When specific "Sub users" are configured in "permissions", the "v3" field will be written, but the "Allow" and "Deny" written in the field will be capitalized and will not take effect. Only when I change the first letter to lowercase "allow" or "deny" will it take effect.

Version:  v1.367.0

@hsluoyz

imp2002:
> 2\. Only when I change the first letter to lowercase "allow" or "deny" will it take effect.

Could you post your `casbin model`? Theoretically speaking, the `action` in the `policy` is the same as in the `model`.

aosolao:
> > 2. Only when I change the first letter to lowercase "allow" or "deny" will it take effect.
> 
> Could you post your `casbin model`? Theoretically speaking, the `eft` in the `policy` is the same as in the `model`.
@imp2002 

Model text:
```
[request_definition]
r = sub, obj, act

[policy_definition]
p = sub, obj, act, eft

[role_definition]
g = _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub) && r.obj == p.obj && r.act == p.act
```


If I try to change 'allow' to 'Allow' in p.eft == allow, it will result in an error message stating 'unsupported effect'.

imp2002:
can try use `allow`? Maybe the implementation of eft needs to be improved, waiting for discussion.

aosolao:
> can try use `allow`? Maybe the implementation of eft needs to be improved, waiting for discussion.

'allow' is not the issue. The problem I encountered is with the incorrect database operation on the Casdoor web page.

PR #2103, opened 2023-07-21
feat: fix `Effect` in Casbin rule

imp2002:
Fix: https://github.com/casdoor/casdoor/issues/2102


hsluoyz:
The code should not be deleted

imp2002:
@hsluoyz The logic of user adding policy and role adding policy seems to be duplicated, can't they be merged?

image

PR #2136, opened 2023-07-25
feat: prevent infinite recursion when role contain subrole

notdu:
feat: prevent infinite recursion when role contain subrole and vice versa. when both of them are still not visited, it will cause infinite recursion


hsluoyz:
@notdu can you explain the code? Why set to `false`?

Issue #2020, opened 2023-06-27
[question] Casbin group policy with domains


hound672:
Hello.
We'd like to implement groups management in our application which similar to Gitlab groups.
There are users, groups and links between them.

You can specify role for user in the group.
For example, user_1 belongs to group_1 with role admin,
user_2 belongs to group_2 with role admin and group_1 with role auditor.

For now, we're using RBAC with domains model.

Model:
```
[request_definition]
r = sub, dom, obj, act

[policy_definition]
p = sub, dom, obj, act

[role_definition]
g = _, _, _

[policy_effect]
e = some(where (p.eft == allow))

[matchers]
m = g(r.sub, p.sub, r.dom) && r.dom == p.dom && r.obj == p.obj && r.act == p.act
```

Policy and role definition we make with casbin editor:
```
p, admin, group_1, view_report, read
p, admin, group_1, create_report, read
p, admin, group_1, share_report, read
p, auditor, group_1, view_report, read

p, admin, group_2, view_report, read
p, admin, group_2, create_report, read
p, admin, group_2, share_report, read
p, auditor, group_2, view_report, read

g, user_1, admin, group_1
g, user_2, auditor, group_1
g, user_3, auditor, group_1

g, user_1, auditor, group_2
g, user_2, admin, group_2
```

Enforse works as expected:

```
user_1, group_1, view_report, read // true since user_1 belongs to domain group_1 with admin role (and admin role has view_report, read)
user_2, group_1, view_report, read // true as well
user_1, group_1, create_report, read // true 
user_2, group_1, create_report, read // false since there user_2 belongs as auditor to domain group_1 (add auditor doesn't have create_report, read )
```

And for `group_2`:
```
user_1, group_2, view_report, read // true
user_2, group_2, view_report, read // true as well
user_1, group_2, create_report, read // false 
user_2, group_2, create_report, read // true
```

Here is the link to casbin editor with this policy: https://editor.casbin.org/#ATL44EZVS

How we can mapping this policy rules to casdoor?
Should we duplicate each role for every domain? And after that we duplicate each permission for every domain's role, do we?

Otherwise it doesn't work as expected.
For example:
1. I create two roles: admin, auditor by casdoor and pass our domains (group_1, group_2)
2. I create three permissions: view_report, create_report, share_report. Link them to roles.

After that we have in the `permission rule` table:
```
ptype,v0,v1,v2,v3,v4,v5
p,built-in/admin,group_1,view_report,"","",built-in/share_report
[...]

PR #2178, opened 2023-08-05
feat: use the casbin model to store relationships between users and groups

leo220yuyaodog:
The pr use casbin store the relationship tetween users and groups, replace the old query statment that using `like` to query users in group.  Note that this commit now does not consider migrating existing user and group relationships into casbin, which would result in users not being found in the organization's group management. I'm not sure if I need to add a migration file before the pr be merged.

The `GroupManager` is not being used now. However, it may then be used to store group hierarchy to determine the inherited permissions of all groups above the user.

Next step, improve the permisson page, support UI to add permission for a group. Because of the implement of using casbin manager user in Casdoor, then we can support the more convenient API, such as `HasPermissionForUser()` to ckeck if a user(group) has permision. It will be more friendly than using Enforce API directly.

Fix: https://github.com/casdoor/casdoor/issues/2020

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -90,7 +90,7 @@ func (c *ApiController) GetUsers() {
 
 	if limit == "" || page == "" {
 		if groupName != "" {
-			maskedUsers, err := object.GetMaskedUsers(object.GetGroupUsers(groupName))
+			maskedUsers, err := object.GetMaskedUsers(object.GetGroupUsers(util.GetId(owner, groupName)))
 			if err != nil {
 				c.ResponseError(err.Error())
 				return
@@ -567,6 +567,22 @@ func (c *ApiController) RemoveUserFromGroup() {
 	name := c.Ctx.Request.Form.Get("name")
 	groupName := c.Ctx.Request.Form.Get("groupName")
 
-	c.Data["json"] = wrapActionResponse(object.RemoveUserFromGroup(owner, name, util.GetId(owner, groupName)))
-	c.ServeJSON()
+	organization, err := object.GetOrganization(util.GetId("admin", owner))
+	if err != nil {
+		return
+	}
+	item := object.GetAccountItemByName("Groups", organization)
+	res, msg := object.CheckAccountItemModifyRule(item, c.IsAdmin(), c.GetAcceptLanguage())
+	if !res {
+		c.ResponseError(msg)
+		return
+	}
+
+	affected, err := object.DeleteGroupForUser(util.GetId(owner, name), groupName)
+	if err != nil {
+		c.ResponseError(err.Error())
+		return
+	}
+
+	c.ResponseOk(affected)
 }
--- a/object/adapter.go
+++ b/object/adapter.go
@@ -23,6 +23,7 @@ import (
 	"github.com/casdoor/casdoor/util"
 	xormadapter "github.com/casdoor/xorm-adapter/v3"
 	"github.com/xorm-io/core"
+	"github.com/xorm-io/xorm"
 )
 
 type Adapter struct {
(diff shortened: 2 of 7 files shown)


object/adapter.go
```diff
@@ -181,7 +185,8 @@ func (adapter *Adapter) initAdapter() error {
 		}
 
 		var err error
-		adapter.Adapter, err = xormadapter.NewAdapterByEngineWithTableName(NewAdapter(adapter.DatabaseType, dataSourceName, adapter.Database).Engine, adapter.getTable(), adapter.TableNamePrefix)
+		engine, err := xorm.NewEngine(adapter.DatabaseType, dataSourceName)
```
leo220yuyaodog, line 188: If use `NewAdapter()` to crete connect, it will cause `sql: connect is closed error`.  Because it use the same connect as the ormer.

object/user_enforcer.go
```diff
@@ -0,0 +1,104 @@
+package object
+
+import (
+	"github.com/casbin/casbin/v2"
+	"github.com/casbin/casbin/v2/errors"
+	"github.com/casdoor/casdoor/util"
+)
+
+type IUserGroupEnforcer interface {
```
hsluoyz, line 9: No interface

controllers/user.go
```diff
@@ -90,7 +90,7 @@ func (c *ApiController) GetUsers() {
 
 	if limit == "" || page == "" {
 		if groupName != "" {
-			maskedUsers, err := object.GetMaskedUsers(object.GetGroupUsers(groupName))
+			maskedUsers, err := object.GetMaskedUsers(object.GetGroupUsers(util.GetId(owner, groupName)))
```
hsluoyz, line 93: Isn't this already merged?

object/adapter.go
```diff
@@ -313,7 +318,7 @@ func (adapter *Adapter) builtInAdapter() bool {
 		return false
 	}
 
-	return adapter.Name == "permission-adapter-built-in" || adapter.Name == "api-adapter-built-in"
+	return adapter.Name == "user-adapter-built-in" || adapter.Name == "api-adapter-built-in"
```
hsluoyz, line 321: Isn't this already merged?

object/enforcer.go
```diff
@@ -121,6 +121,10 @@ func DeleteEnforcer(enforcer *Enforcer) (bool, error) {
 }
 
 func (enforcer *Enforcer) InitEnforcer() error {
+	if enforcer == nil {
```
hsluoyz, line 124: Isn't this already merged?

object/group.go
```diff
@@ -214,39 +214,33 @@ func ConvertToTreeData(groups []*Group, parentId string) []*Group {
...
 	users := []*User{}
+	owner, _ := util.GetOwnerAndNameFromId(groupId)
+	names, err := userEnforcer.GetUserNamesByGroupName(groupId)
+	if err != nil {
+		return nil, err
+	}
+
 	session := ormer.Engine.Table("user").
-		Where(builder.Like{"`groups`", groupId + "\""})
+		Where("owner = ?", owner).In("name", names)
```
hsluoyz, line 243: Isn't this already merged?

hsluoyz (changes requested):

object/group.go
```diff
@@ -265,7 +259,7 @@ func GetPaginationGroupUsers(groupId string, offset, limit int, field, value, so
 		session = session.Desc(fmt.Sprintf("user.%s", util.SnakeString(sortField)))
 	}
 
-	err := session.Find(&users)
+	err = session.Find(&users)
```
hsluoyz, line 262: Why change this? Is this related to the goal of this PR? If not, then discard it

hsluoyz (changes requested):

object/group.go
leo220yuyaodog: Because the added code above has defined err, use `:=` cannot compile.

controllers/user.go
leo220yuyaodog: No, has rebased

hsluoyz (approved):

PR #2202, opened 2023-08-11
Allow multiple similar providers, display by display name

hopomi:
Then log in to the page, when there are multiple similar providers, they cannot be distinguished. Using displayName is easier to distinguish.

5befe7ae5cfe6a4af1a684a3f44bd680


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/commit/9156bd426ba8927c19da344185d64ce3085d6f9a

PR #2296, opened 2023-09-02
feat: fix bug that failed to set password after changing username

Selflocking:
When we add a new member to an organization using Casdoor, Casdoor will automatically generate a member with a random username, such as "user_qvducc". When we change the username, for example, to "yunshu", an issue arises where we are unable to successfully edit the password. This is because Casdoor searches for a user based on `owner/username`, and before any changes are saved, the username in the database remains "user_qvducc". However, the frontend uses `orgName/yunshu` instead of `orgName/user_qvducc` to send the request to change the password. As a result, the backend cannot find the user and the password change fails.

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -448,6 +448,11 @@ func (c *ApiController) SetPassword() {
 	}
 
 	targetUser, err := object.GetUser(userId)
+
+	if targetUser == nil {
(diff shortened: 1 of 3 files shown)


Selflocking:
demo: 
Peek 2023-09-02 22-53

PR #2306, opened 2023-09-06
feat: set default passwordType to argon2id when add organization

wzy9607:
argon2id is currently the recommended Key Derivation Function to store password by OWASP.

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html


hsluoyz:
We don't have plan for this change right now. You can change your own code for now.

PR #2307, opened 2023-09-06
fix(cas): avoid return user's secrets at serviceValidate

wzy9607:


hsluoyz:
@wzy9607 replaced by: https://github.com/casdoor/casdoor/commit/b7cd604e5602fb66362f9ef5674a945ed7d347f5

PR #2334, opened 2023-09-15
fix: logout failed

a5272689:


a5272689:
image
it is done,but in Pending

PR #2349, opened 2023-09-20
fix: empty AzureAD tenant id

### LINKED ISSUE #2348: AzureAD broken because tenant id is empty
Trying to log in with a fully configured AzureAD provider (Client-ID, Client-Secret and Domain* supplied) produces the following error message:

AADSTS50194: Application 'xxx' is not configured as a multi-tenant application. Usage of the /common endpoint is not supported for such applications created after '10/15/2018'. Use a tenant-specific endpoint or configure the application to be multi-tenant.

Bildschirmfoto von 2023-09-14 22-13-14

After some digging I found out that the tenant id is stored in the field "Domain"

[object/provider.go](https://github.com/casdoor/casdoor/blob/3b2820cbe3b691da4195a4a42968af8f97615567/object/provider.go#L65)

but it cannot be retrieved in the relevant code as only the field "HostUrl" is being supplied

[idp/provider.go](https://github.com/casdoor/casdoor/blob/3b2820cbe3b691da4195a4a42968af8f97615567/idp/provider.go#L118)

In the following sections the field "HostUrl" is being checked, but it is always empty thus leading to the tentant id never being available to the relevant parts of the code

[idp/goth.go](https://github.com/casdoor/casdoor/blob/3b2820cbe3b691da4195a4a42968af8f97615567/idp/goth.go#L109-L112)

Therefore the "tenant id" is always "common" leading to the error message above.

___
\* Side note 1: the term "Domain" is absolutely misleading because it is not mentioned in the **[docs](https://casdoor.org/docs/provider/oauth/azureAD)** and the field is about the tenant id and not about changing the domain (login.microsoftonline.com).
\* Side note 2: the **[docs](https://casdoor.org/docs/provider/oauth/azureAD)** create the impression that a multitenant application is needed to work with Casdoor which is wrong as a single tenant application will work as well.
I would be glad to create a PR to fix this if you are interested!

fischermario:
Fixes #2348

### DIFF
--- a/object/provider.go
+++ b/object/provider.go
@@ -415,6 +415,8 @@ func FromProviderToIdpInfo(ctx *context.Context, provider *Provider) *idp.Provid
 			providerInfo.ClientId = provider.ClientId2
 			providerInfo.ClientSecret = provider.ClientSecret2
 		}
+	} else if provider.Type == "AzureAD" {
+		providerInfo.HostUrl = provider.Domain
 	}
 
 	return providerInfo


object/provider.go
```diff
@@ -397,13 +397,17 @@ func providerChangeTrigger(oldName string, newName string) error {
 }
 
 func FromProviderToIdpInfo(ctx *context.Context, provider *Provider) *idp.ProviderInfo {
+	hostUrl := provider.Host
```
hsluoyz, line 400: Just change the code for your provider type like:

```go
if provider.Type == "WeChat" {
		if ctx != nil && strings.Contains(ctx.Request.UserAgent(), "MicroMessenger") {
			providerInfo.ClientId = provider.ClientId2
			providerInfo.ClientSecret = provider.ClientSecret2
		}
	}
```

hsluoyz (changes requested):

object/provider.go
fischermario: That makes more sense. I just updated the PR. Please check!

hsluoyz (approved):

Issue #2344, opened 2023-09-19
Hope to support SMS templates with multiple variables.


mapleafgo:
Hope to support SMS templates with multiple variables, such as `Flight {0} you booked on {1} is about to leave, Please get there as soon as possible.`

mapleafgo:
@hsluoyz Hello, could you explain why you closed this issue? Is it already supported in the project?

PR #2353, opened 2023-09-21
feat: Support SMS templates with multiple variables

mapleafgo:
Fix: https://github.com/casdoor/casdoor/issues/2344
Also compatible with the previous request message format

### DIFF
--- a/controllers/service.go
+++ b/controllers/service.go
@@ -35,9 +35,10 @@ type EmailForm struct {
 }
 
 type SmsForm struct {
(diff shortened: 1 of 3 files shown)


hsluoyz:
The PR doesn't make sense. Just make things complicated without any explanation

mapleafgo:
> The PR doesn't make sense. Just make things complicated without any explanation

I think so too, but I haven't thought of a more elegant way to support this issue for the time being. Do you have any good suggestions?

PR #2369, opened 2023-09-28
feat: run RecordMessage() filter after API handling

### LINKED ISSUE #2368: Records and Webhooks occurs even if event did't happend
**Problem:** 
If some exception happens during request handling (such /api/update-user and e.t.c. ) we still will have new Record and webhook call. 

**Reason:** 
	`beego.InsertFilter("*", beego.BeforeRouter, routers.RecordMessage)`

**Proposal:** 
Change to `beego.InsertFilter("*", beego.AfterExec, routers.RecordMessage)`

**Benefits:**
- No false records / webhook calls in case of panic 
- There is option to include response status code to record  / webhook
- No possibility to race condition ( when webhook calls lead to API request before the entity is updated in the database

ISulimanov:
fix #2368

### DIFF
--- a/main.go
+++ b/main.go
@@ -59,7 +59,7 @@ func main() {
 	beego.InsertFilter("*", beego.BeforeRouter, routers.CorsFilter)
 	beego.InsertFilter("*", beego.BeforeRouter, routers.ApiFilter)
 	beego.InsertFilter("*", beego.BeforeRouter, routers.PrometheusFilter)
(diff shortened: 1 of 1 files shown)


hsluoyz:
@ISulimanov does this change work? Have you tested it?

hsluoyz:
This PR is reverted by: https://github.com/casdoor/casdoor/commit/69b144d80f927dd886bb6dc72b21980b1d67ba8e

because of: https://github.com/casdoor/casdoor/issues/2399#issuecomment-1761126190

PR #2372, opened 2023-09-28
fix: Added new property domain to use inside function getOriginFromHost

gtjadsonsantos:


hsluoyz:
The change doesn't make sense

Issue #2397, opened 2023-10-12
Error when trying to GET /api/get-user with only an email address


ppadron:
Searching for an user by email or phone on `/api/get-user` seems to work only if you also provide its owner. If you try to search by email only this error is generated:

```
casdoor:GetOwnerAndNameFromId() error, wrong token count for ID: 

/usr/local/go/src/runtime/panic.go:884
/go/src/casdoor/util/string.go:114
/go/src/casdoor/controllers/user.go:164
/usr/local/go/src/reflect/value.go:584
/usr/local/go/src/reflect/value.go:368
/go/pkg/mod/github.com/beego/beego@v1.12.12/router.go:897
/usr/local/go/src/net/http/server.go:2947
/usr/local/go/src/net/http/server.go:1991
/usr/local/go/src/runtime/asm_amd64.s:1594
```

PR #2398, opened 2023-10-12
feat: support calling get-user API by only email, phone or userId without owner

ppadron:
fix: #2397

This commit will add `GetUserByEmailOnly` and `GetUserByPhoneOnly` so they can be used by the `/api/get-user` endpoint.

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -160,35 +160,47 @@ func (c *ApiController) GetUser() {
 		id = util.GetId(userFromUserId.Owner, userFromUserId.Name)
 	}
 
-	if owner == "" {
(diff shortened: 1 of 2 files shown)


hsluoyz:
@ppadron don't add new APIs, modify the exsiting get-user to support this case

ppadron:
I did not add new APIs, I changed `/api/get-user`.

But for that I had to add two new functions on `object/user.go` to retrieve the user by email (`GetUserByEmailOnly`) and phone (`GetUserByPhoneOnly`) without the owner.

I used the naming pattern from the existing `GetUserByUserIdOnly`.

PR #2450, opened 2023-10-27
feat: fix bug that tableNamePrefix caused getRolesByUserInternal() to fail

wrongch:
If set tableNamePrefix in app.conf, while cause sql error


anyidea:
@wrongch GROUPS is a reserved keyword introduced in MySQL 8.0.

PR #2460, opened 2023-10-30
fix: If redirectUri already has parameters, concatenate the code and state parameters with &, when fastAutoSignin.

u3breeze:
1. Delete unnecessary " from Payment struct.
 2. If redirectUri already has parameters, concatenate the code and state parameters with &, when fastAutoSignin.


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/commit/2398e69012c9b2d5f13fc0546da67f8e205f0d53

PR #2462, opened 2023-10-31
fix:The displayName and detail are allowed to be modified on the product page.

u3breeze:
The displayName and detail are allowed to be modified on the product page, so do not update the product's own displayName and detail when updating the plan.


hsluoyz:
That's made for a purpose

PR #2471, opened 2023-11-02
Improve captcha security

blacksponge:
## Context

Current captcha implementation
 * leaks the secrets to the client, which is a bad practice and can lead to security issues;
 * only rely on client data to verify the captcha (secret and type are both only taken from the the user form), by using specific values (Cloudflare testing credentials for example), an attacker can easily bypass the captcha this way;
 * can be bypass just by setting `captchaType` to `none` for email / phone verification.

This PR tries to fix those issues.

## Changes

* The response to the `/api/get-captcha` request no longer fills the fields `ClientSecret`, `ClientId2` and `ClientSecret2`
* For captcha type `Aliyun Captch`, the fields `Scene` and `AppKey` are used instead of `ClientId2` and `ClientSecret2`
* The authentication form for the `/api/login` request no longer accepts `ClientSecret`, for the default captcha the `CaptchaId` field is used instead
* The captcha provider is fetched using the application ID (`captcha.GetCaptchaProviderByApplication` method with `isCurrentProvider` to `false`) for the email / phone verification and the login routes
* The captcha provider is fetched using the provider ID (`captcha.GetCaptchaProviderByApplication` method with `isCurrentProvider` to `true`) for the verify captcha route

**Note:** I have only tested the default captcha and cloudfare using the testing credentials as I don't have accounts for any of the captcha providers.

Cheers,
Gaël


blacksponge:
It should now be fixed.

blacksponge:
As I didn't receive any feedback those last two weeks, is this PR something you would consider?

Issue #2464, opened 2023-10-31
[Feature]Filter the application list to exclude apps that users cannot access.


anyidea:
Even though I've set certain apps as inaccessible for a particular user in the permissions list, they are still visible in the home page app list.

anyidea:
I believe the home page app list could include an API, something like "/api/get-user-applications," to filter the applications accessible to the user.

hsluoyz:
@anyidea no need, just use `/api/get-applications`, filter out the ones that don't have access. BTW, can you make a PR?

anyidea:
> @anyidea no need, just use `/api/get-applications`, filter out the ones that don't have access. BTW, can you make a PR?

We can apply permission filtering for non-admin users in the `GetOrganizationApplications` method.

hsluoyz:
Fixed by: https://github.com/casdoor/casdoor/commit/498cd02d49db6b48d58c651485f5a2e5045e316d

PR #2480, opened 2023-11-07
feat: filter the application list to exclude apps that users cannot access

anyidea:
fix: #2464


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/commit/498cd02d49db6b48d58c651485f5a2e5045e316d

PR #2486, opened 2023-11-08
feat: fix the UI and navigation errors on the prompt page

Songjf-ttk:
Fix: https://github.com/casbin/casdoor/issues/2417

1. Fixed the incorrect prompt page preview in the application edit page.
2. Fixed the redirection error after signup -> binding; now it redirects directly to the home page after signup and binding.


Songjf-ttk:
issue2417

Songjf-ttk:
issue2417

Issue #2478, opened 2023-11-07
[Bug]The login page path is incorrect, after the session expires


anyidea:
After the non-built-in organization user session expires, it should be redirected to the login page `/login/{organization}` instead of `/login`.

hsluoyz:
@anyidea can you make a PR?

PR #2493, opened 2023-11-13
feat: fix login page path after logout

anyidea:
fix: #2478


hsluoyz:
@anyidea this PR has bug, it causes users to log into wrong organization, will be reverted

PR #2506, opened 2023-11-19
feat: make MFA works for CAS login

Songjf-ttk:
Fix: https://github.com/casbin/casdoor/issues/2205

### DIFF
--- a/web/src/auth/LoginPage.js
+++ b/web/src/auth/LoginPage.js
@@ -361,7 +361,7 @@ class LoginPage extends React.Component {
                   <MfaAuthVerifyForm
                     mfaProps={res.data2}
                     formValues={values}
-                    Params={casParams}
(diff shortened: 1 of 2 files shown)


Songjf-ttk:
issue2205

Songjf-ttk:
In [f49f4d7fc9](https://github.com/casbin/casdoor/commit/f49f4d7fc9324a595031e83a82dee857d6c778d3), the err handling was changed. After the modification, `err` is used in **ApiController Login**, which is a local variable. The issue with the local variable `err` is that if the password is incorrect during the login process, the error cannot be communicated outward. 
```
application, err := object.GetApplication(fmt.Sprintf("admin/%s", authForm.Application))
if err != nil {
  c.ResponseError(err.Error())
  return
}
```

PR #2533, opened 2023-12-10
feat: Host Helm Chart in Github Pages

sp71:
**What**
- Host Helm Chart in Github Pages

**How**
- Followed Helm Chart [Guide](https://helm.sh/docs/topics/chart_repository/#github-pages-example) on Github Pages Hosting
- Moved the chart to docs directory as recommended by Github Helm Chart Guide

**Why**
- Easier onboarding for developers to install Casdoor via helm chart

**Tested?**
image


docs/charts/Chart.yaml
```diff
@@ -21,4 +21,4 @@ version: 0.2.0
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "1.17.0"
+appVersion: "1.471.0"
```
hsluoyz, line 24: Does it need to be the same as Casdoor version? Is it good to hard code it? What if new release comes out?

docs/charts/casdoor-0.2.0.tgz
hsluoyz, line 1: Why commit a binary file?

docs/charts/templates/NOTES.txt
hsluoyz, line 1: Why rename to docs folder?

docs/index.md
hsluoyz, line 1: Docs put to: https://casdoor.org/docs/basic/try-with-docker instead of here

hsluoyz (changes requested):

docs/charts/Chart.yaml
sp71: Helm [recommends](https://helm.sh/docs/topics/charts/#the-appversion-field) matching the application version with the same value. Regarding new releases, some automation would need to be put in place (e.g. Github Action). If you prefer otherwise, that's fine as well.
>  the appVersion field is ... a way of specifying the version of the application. For example, the drupal chart may have an appVersion: "8.2.1", indicating that the version of Drupal included in the chart (by default) is 8.2.1. This field is informational, and has no impact on chart version calculations

docs/charts/casdoor-0.2.0.tgz
sp71: Yeah this file isn't strictly required since the raw helm files are available. Will remove

docs/charts/templates/NOTES.txt
sp71: I originally presumed the folder needed to be hardcoded to docs but Github allows admins to [change](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) the directory so will revert to older naming pattern

docs/index.md
sp71: For [Github Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-github-pages-site), they require a file to be placed in the directory such as index.md

hsluoyz:
@sp71 you'd better test your code in your own repo, before making PR

image

PR #2547, opened 2023-12-17
Docker build speedup

jum:
This is an idea how to speed up docker builds to be able to support multi-arch builds properly. The check for the special GOPROXY variable used in China has been moved to the build_docker.sh script, which uses docker buildx now. By not using qemu builds but go cross builds the build time is now completely dominated by the frontend build time. The frontend is only build once and not for each arch. Further speedup may be possible in the frontend build.


build_docker.sh
```diff
@@ -1,11 +1,13 @@
-#!/bin/bash
+#!/bin/sh
+DOCKER_PREFIX=jumager/casdoor
```
hsluoyz, line 2: What's `jumager`?

hsluoyz:
We have removed `arm64` images from official builds and won't add it any more because it's too slow and the images also take up our space in DockerHub.

Anyone who wants to use Casdoor in arm64 can build their own images by themselves.

So closed here

jum:
That is a leftover I have forgotten. To test and build, I needed a docker repository to push to. Should I change that to an required argument? The official casdoor repo should not be in there.Am 16.12.2023 um 23:52 schrieb hsluoyz ***@***.***>:﻿
@hsluoyz commented on this pull request.



In build_docker.sh:
> @@ -1,11 +1,13 @@
-#!/bin/bash
+#!/bin/sh
+DOCKER_PREFIX=jumager/casdoor

What's jumager?

PR #2549, opened 2023-12-17
fix: helm add index.yaml

sp71:
**Description**
- Index.yaml was not being generated by Github Action because it was being ignored
- Adding file into git so it will be tracked
- File necessary as it informs clients where to download helm chart from
- Updated app version to match latest

### DIFF
--- a/manifests/casdoor/Chart.yaml
+++ b/manifests/casdoor/Chart.yaml
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.3.0
+version: 1.477.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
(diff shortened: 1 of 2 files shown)


hsluoyz:
@sp71 does it mean that we have to manually update these versions for each release? Notice that we use semantic-release to auto-release new versions, so this is nearly unacceptable

image

sp71:
@hsluoyz Thanks for the quick review. I've updated the PR to be the next release version. Currently it's configured to be manual as we only make changes to the helm chart version if the helm chart changes. The application version is configurable through the helm chart itself. 

We can shift to another model where every release triggers the helm chart to be updated. The only difference would the helm chart version, the helm chart code itself would usually be the same which would lead to many duplicates in docker hub. 

I am hoping the [CI](https://github.com/casbin/casdoor/blob/master/.github/workflows/helm.yml#L37) will now recognize the index.yaml file it generates as its now added to git and commit it with the correct digest field.

hsluoyz:
@sp71 

1. We don't want to update those files (index.yaml and Chart.yaml) for each different version number, even auto merge. Just don't want those commits showing up in this repo. If those files have to be dynamically updated, generate them dynamically in CI only and don't add them to this git repo.
2. We don't specify any specific version numbers in the git repo's files. Those are auto-controlled by semantic-release

hsluoyz:
Closed for now, as it seems to aim very different goal

PR #2582, opened 2024-01-06
chore: update helm version to v1.492.0

MichaelSp:
According to https://hub.docker.com/r/casbin/casdoor/tags and https://github.com/casbin/casdoor/releases latest version is `v1.492.0`

The process can be automated. See https://github.com/casbin/casdoor/pull/2583

### DIFF
--- a/manifests/casdoor/Chart.yaml
+++ b/manifests/casdoor/Chart.yaml
@@ -21,4 +21,4 @@ version: 0.3.0
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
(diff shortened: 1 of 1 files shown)


hsluoyz:
@MichaelSp this is really a hard thing to maintain. This PR should not be released right? Or the version will not be 492. And I hope this version number can be slightly later than latest. Because we can't update it that often.

MichaelSp:
@hsluoyz yes, maybe you want to add [some path filters](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) here https://github.com/casbin/casdoor/blob/master/.github/workflows/build.yml#L3C1-L3C1

PR #2590, opened 2024-01-08
fix: missing table prefix in get user group

xgenvn:
- Fix an issue with the sort field and order field missing table name prefix, affecting the get user group API.


hsluoyz:
@xgenvn fix CI:

image

PR #2591, opened 2024-01-08
feat: add LDAP signin method

HGZ-20:
Add support for LDAP login methods
Add option to control LDAP user in password login method.

Fix: https://github.com/casbin/casdoor/issues/2472

https://github.com/casbin/casdoor/assets/46803939/c82ab9fc-9650-42cf-abd1-843650b3d646


controllers/auth.go
```diff
@@ -399,10 +399,14 @@ func (c *ApiController) Login() {
 				c.ResponseError(fmt.Sprintf(c.T("auth:The application: %s does not exist"), authForm.Application))
 				return
 			}
-			if !application.IsPasswordEnabled() {
+			if authForm.SigninMethod == "password" && !application.IsPasswordEnabled() {
```
hsluoyz, line 402: Values as follows

W6{$Y55Q){%@ DVVZ){0XB

object/application.go
```diff
@@ -544,6 +554,19 @@ func (application *Application) IsPasswordEnabled() bool {
 	}
 }
 
+func (application *Application) IsPasswordWithLdapEnabled() bool {
+	if len(application.SigninMethods) == 0 {
+		return application.EnablePassword
+	} else {
+		for _, signinMethod := range application.SigninMethods {
+			if signinMethod.Name == "Password" && signinMethod.Rule == "All" {
```
hsluoyz, line 562: "None" should also work like "All"

hsluoyz (changes requested):

controllers/auth.go
HGZ-20: Done

object/application.go
HGZ-20: Because the None value only affects new users within a week, compatibility with the None option will be dropped

hsluoyz (approved):

PR #2617, opened 2024-01-16
feat: Changed username regex to allow email-like usernames

ajguidet:
Hey there! First off, a big thanks to everyone for maintaining Casdoor
The motivation of this PR is to make the validation process more generic, allowing email-like usernames. This update ensures that the username validation aligns more closely with real-world usage where dots are frequently used in usernames.


hsluoyz:
@ajguidet there's no need to change this. If you want to use Email to signup, you can either:

1. Replace "Username" with "Email" in signup items table of the application edit page
2. Use "Verification code" instead of "Password" for signin methods table of the application edit page

For username, we don't want to mess it with Email. Username is username, Email is Email. Mixing them will bring trouble for Casdoor, like Alice's username is Bob's Email. So the login via username/Email/Phone will also be fragile. And there's no gain for this.

PR #2632, opened 2024-01-20
feat: use standalone Twitter OAuth provider instead of goth

LiusCraft:
Currently, casdoor is unable to log in through Twitter. I have written my own Twitter provider to solve this problem


hsluoyz:
@LiusCraft plz fix CI:

image

PR #2660, opened 2024-01-31
feat: ABAC support for /api/enforce endpoint

lordspinach:
Hello! 
Recently i tried to enforce permission with ABAC using 'api/enforce' endpoint. I have model with matchers like this: 

`m = r.sub.Id == r.obj.CreatedBy`

And json body like this:

```json
[
    { "Id": "some-uuid-id" }, 
    { "CreatedBy": "some-uuid-id" },
    "delete"
]
```

If i got it right, currently this endpoint cannot handle objects in 'sub' or 'obj' so i implement this simple function to fix this

### DIFF
--- a/util/json.go
+++ b/util/json.go
@@ -14,7 +14,10 @@
 
 package util
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"reflect"
+)
 
 func StructToJson(v interface{}) string {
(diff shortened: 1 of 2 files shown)


giangdinhtt:
Hi @lordspinach, would you mind supporting ABAC for /api/batch-enforce as well?

PR #2664, opened 2024-02-01
feat: add ability to set Recipient parameter name for Custom HTTP SMS provider

Insei:
It is not recommended to hardcode the parameter name for a phone number; I have added the ability to store the parameter name in the provider metadata field.


Insei:
Force push: renamed Receiver to Recipient

hsluoyz:
No need to add i18n for such a small feature

Insei:
Hi @hsluoyz!, Thanks for reply, but this PR not about i18n translations. I add ability to set phone number parameter name (for GET and POST requests in Custom SMS HTTP provider). Without this changes phone number sends only in "phoneNumber" query parameter for GET requests and in "phoneNumber" form key for POST requests. 

It's can be very useful for integration with other SMS services that is not supported currently by casdoor.

Issue #2683, opened 2024-02-07
Custom HTTP SMS provider "Phone Number" param name can't be changed


Insei:
Hi all!
Now there is a problem with the Custom HTTP SMS provider; it does not support changing the parameter name in which the recipient's phone number is transmitted.
**Casdoor sends recipient's phone number in query for GET:**
GET `https://some-sms-service/sms?<Parameter>=<sms template + code>&phoneNumber=<recipient's phone number>`
https://github.com/casbin/casdoor/blob/5c4e22288e3e821996554b5588acd10cd3432f5c/object/sms_custom_http.go#L69-L71

**And in form, for POST.**
POST `https://some-sms-service/sms`
Form data:
`phoneNumber` = `<recipient's phone number>`
`<Parameter>` = `<sms template + code>`
https://github.com/casbin/casdoor/blob/5c4e22288e3e821996554b5588acd10cd3432f5c/object/sms_custom_http.go#L54-L56

Only the parameter in which the message text is transmitted is supported i.e. `<Parameter>`, but for `<recipient's phone number>` parameter name is hardcoded as `phoneNumber`

I already send PR with needed changes, but it closes with unclear reason: https://github.com/casbin/casdoor/pull/2664

Can we discuss a solution that suits for this project here?

hsluoyz:
@Insei the reason is already said: "No need to add i18n for such a small feature", it means that this is a small feature, and you should not try to add new i18n word for it

Insei:
@hsluoyz Ok, thanks for the answer, but usually maintainers do not close the PR, but say what needs to be changed in it.
I could just remove the i18n translation by rebasing the branch, and ask to look at the same PR again. =)

hsluoyz:
I really don't think this feature is needed for this project. You can make some forwarding code to make it work. This is not a low-code or no-code project to let people customize everything

PR #2684, opened 2024-02-07
feat: add ability to set recipient parameter name for Custom

Insei:
feat: add ability to set recipient parameter name for Custom

without i18n translations

#2683 
#2664


hsluoyz:
Closed due to: https://github.com/casbin/casdoor/issues/2683#issuecomment-1932360320

Issue #2596, opened 2024-01-10
frontend: Regarding the color of loading


hoho2017:
When I'm using Casdoor, for example, when the login page redirects to the registration page, there is a loading effect in between. Why isn't the theme color I set taking effect? The color of this loading effect is purple, not the theme color I set. Can any handsome guy tell me how to fix this? I've even overridden the style of the spin, but it still doesn't work.

hoho2017:
@hsluoyz ok wait a moment

hoho2017:
Uploading 录屏2024-01-11 上午9.32.18.mov…

@hsluoyz

hoho2017:
https://github.com/casbin/casdoor/assets/23206991/ecd68f52-bb22-4346-b8d8-5aa15af4655c

@hsluoyz

hoho2017:
My theme color is green, but the loading icon and spin are purple.

PR #2718, opened 2024-02-18
fix: use "soft router" for signup page for switching pages smoothly

xiao-kong-long:
Fix: #2596

### DIFF
--- a/init_data.json.template
+++ b/init_data.json.template
@@ -8,12 +8,62 @@
       "favicon": "",
       "passwordType": "plain",
       "passwordSalt": "",
-      "passwordOptions": ["AtLeast6"],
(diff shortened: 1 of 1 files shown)


web/src/Setting.js
```diff
@@ -1234,6 +1236,43 @@ export function renderSignupLink(application, text) {
   return renderLink(url, text, storeSigninUrl);
 }
 
+function removeDomain(url) {
+  const domain = window.location.origin;
+  url = url.substring(domain.length);
```
hsluoyz, line 1241: Don't self op string

web/src/Setting.js
```diff
@@ -1234,6 +1236,43 @@ export function renderSignupLink(application, text) {
...
+
+export function renderLoginLink(linkInStorage, application, history, text) {
+  let url;
+
+  const storeSigninUrl = () => {
+    sessionStorage.setItem("signinUrl", window.location.href);
+  };
+
+  if (linkInStorage !== null && linkInStorage !== "") {
+    if (linkInStorage.startsWith("http") || linkInStorage.startsWith("https")) {
```
hsluoyz, line 1253: Is this dup with other code?

hsluoyz (changes requested):

PR #2736, opened 2024-02-23
feat: improve route, separate App.js and use lazy route in EntryPage

dacongda:
feat: #1777 #2596

This pr change lots of file, so it still need a full test and strict review, the following are the main changes
1. use redux to store account data and theme data 
2. separate main content render from App.js and use lazy load to reduce login page size
3. use splitChunks to split the main.js
4. separate getCountryCode function from Setting.js to aviod unnecessary import of i18n-iso-countries
5. use i18next-resources-to-backend to lazy load i18n translation file

with this work, the size of login page successfully reduce to ±950KB

image

### DIFF
--- a/web/craco.config.js
+++ b/web/craco.config.js
@@ -1,6 +1,4 @@
 const CracoLessPlugin = require("craco-less");
-const {BundleAnalyzerPlugin} = require("webpack-bundle-analyzer");
-const path = require("path");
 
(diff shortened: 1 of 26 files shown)


hardywu:
why closed?

hsluoyz:
@hardywu will split into smaller packages

PR #2750, opened 2024-02-26
fix: refactor out ManagementPage.js from App.js

dacongda:
feat: separate management page from app.js and use lazy load to import it , and successfully reduce the size of entry page to 1.4MB when gzip is enable

image

### DIFF
--- a/web/src/ManagementPage.js
+++ b/web/src/ManagementPage.js
@@ -1,3 +1,17 @@
+// Copyright 2024 The Casdoor Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


web/src/ManagementPage.js
```diff
@@ -0,0 +1,427 @@
+import * as Setting from "./Setting";
```
hsluoyz, line 1: Apache header (2024 year)

hsluoyz (changes requested):

hsluoyz:
@dacongda how is it reduced for gzip disabled?

hsluoyz (approved):

Issue #2753, opened 2024-02-27
bug: Email provider not recognized


Lelenaic:
Hi,
On casdoor v1.548.0 I have an issue with email configuration.
An email provider is correctly configured in the providers section.
But when I click on “Send code”, I have the following message displayed: 
```
please add an Email provider to the "Providers" list for the applicationfalse academy
```
Here's the provider configuration:
![CleanShot 2024-02-27 at 14 56 36@2x](https://github.com/casdoor/casdoor/assets/11456619/bf7b6047-325b-4919-aed6-f139603aff29)

And here's the provider configuration in the Application:
![CleanShot 2024-02-27 at 14 58 19@2x](https://github.com/casdoor/casdoor/assets/11456619/46ee095e-78d6-474c-86ba-49bbf8c199cd)

I tried to delete the provider and create it again, but nothing changed.

Lelenaic:
The issue starts from version v1.543.0, I guess it's because of https://github.com/casdoor/casdoor/pull/2733 so.

Lelenaic:
I identified the issue. In the table "application" in the "providers" column, here's the email provider:
`{"owner":"","name":"mail_provider","canSignUp":true,"canSignIn":true,"canUnlink":true,"prompted":false,"signupGroup":"","rule":"None","provider":null}`

"rule" is at "None" but it should be "all". I changed it manually in the DB and it works with the latest Casdoor version.

hsluoyz:
Fixed by: https://github.com/casdoor/casdoor/commit/37232faa07870589ecc03aa21dd5f4ed9637d12c

allenz-srp:
查看评论

PR #2754, opened 2024-02-27
fix: Set default value for email and SMS rule to all instead of none

Lelenaic:
You already fixed https://github.com/casdoor/casdoor/issues/2753 but I added a cleaner way to store the default rule for SMS or email provider.


hsluoyz:
@Lelenaic fix CI

image

Lelenaic:
> @Lelenaic fix CI
> 
> image

Done

Issue #2784, opened 2024-03-07
error on line 14 at column 168: Attribute xmlns redefined


DevOpserzhao:
I'm a local developer.
use http://localhost:7001/api/saml/metadata?application=admin/application_devops&post=true

error  

error on line 14 at column 168: Attribute xmlns redefined

`<EntityDescriptor xmlns:ds="http://www.w3.org/2000/09/xmldsig#" xmlns="urn:oasis:names:tc:SAML:2.0:metadata" xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" entityID="http://localhost:8000">
  <IDPSSODescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol">
    <KeyDescriptor use="signing">
      <KeyInfo xmlns="http://www.w3.org/2000/09/xmldsig#">
        <X509Data xmlns="http://www.w3.org/2000/09/xmldsig#">
[...]

PR #2799, opened 2024-03-13
fix: fix the issue of adding xmlns="" when generating XML

DevOpserzhao:
Fix: https://github.com/casdoor/casdoor/issues/2784

### DIFF
--- a/object/saml_idp.go
+++ b/object/saml_idp.go
@@ -179,7 +179,7 @@ type IdpSSODescriptor struct {
 }
 
 type NameIDFormat struct {
-	//XMLName xml.Name
+	// XMLName xml.Name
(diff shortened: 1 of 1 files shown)


object/saml_idp.go
```diff
@@ -179,8 +179,8 @@ type IdpSSODescriptor struct {
 }
 
 type NameIDFormat struct {
-	XMLName xml.Name
-	Value   string `xml:",innerxml"`
+	//XMLName xml.Name
```
hsluoyz, line 182: What's this PR? You commented two fields and add two prints? Why this PR will work?

hsluoyz (changes requested):

hsluoyz (approved):

PR #2826, opened 2024-03-20
feat: Add Captcha Verification on Signup Page Without Email/Phone Checks

u3breeze:
On the signup page, configured with a Captcha provider:
Captcha verification is done before sending a code for phone or email verification, preventing malicious signups.
For account and password-only registrations, captcha was not checked, posing a security risk. Therefore, I have added captcha verification.

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -17,6 +17,7 @@ package controllers
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


hsluoyz:
@u3breeze feature already available, see docs: https://casdoor.org/docs/provider/captcha/default/#applying-in-your-application

choose "Always"

image

PR #2833, opened 2024-03-22
feat: add DoubleMd5UserSaltCredManager

luliangce:
When I migrated the user system from an old service to Casdoor and synchronized the users, I found that there was no suitable password algorithm. This algorithm may be relatively obscure, but I hope to add it to the Casdoor.

### DIFF
--- a/cred/manager.go
+++ b/cred/manager.go
@@ -34,6 +34,8 @@ func GetCredManager(passwordType string) CredManager {
 		return NewPbkdf2SaltCredManager()
 	} else if passwordType == "argon2id" {
 		return NewArgon2idCredManager()
+	} else if passwordType == "double-md5-salt" {
+		return NewDoubleMd5UserSaltCredManager()
 	}
 	return nil
 }
--- a/web/src/OrganizationEditPage.js
+++ b/web/src/OrganizationEditPage.js
@@ -257,7 +257,7 @@ class OrganizationEditPage extends React.Component {
           </Col>
           <Col span={22} >
             <Select virtual={false} style={{width: "100%"}} value={this.state.organization.passwordType} onChange={(value => {this.updateOrganizationField("passwordType", value);})}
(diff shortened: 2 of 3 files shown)


hsluoyz:
@luliangce can you share which famous systems or services use this? Like: https://casdoor.org/docs/organization/overview/

image

cred/manager.go
```diff
@@ -34,6 +34,8 @@ func GetCredManager(passwordType string) CredManager {
 		return NewPbkdf2SaltCredManager()
 	} else if passwordType == "argon2id" {
 		return NewArgon2idCredManager()
+	} else if passwordType == "double-md5-salt" {
```
hsluoyz, line 37: Put after "md5-salt"

web/src/OrganizationEditPage.js
```diff
@@ -257,7 +257,7 @@ class OrganizationEditPage extends React.Component {
           </Col>
           <Col span={22} >
             <Select virtual={false} style={{width: "100%"}} value={this.state.organization.passwordType} onChange={(value => {this.updateOrganizationField("passwordType", value);})}
-              options={["plain", "salt", "sha512-salt", "md5-salt", "bcrypt", "pbkdf2-salt", "argon2id"].map(item => Setting.getOption(item, item))}
+              options={["plain", "salt", "sha512-salt", "md5-salt", "bcrypt", "pbkdf2-salt", "argon2id", "double-md5-salt"].map(item => Setting.getOption(item, item))}
```
hsluoyz, line 260: Put after "md5-salt"

hsluoyz (changes requested):

hsluoyz:
@luliangce any update?

luliangce:
> @luliangce can you share which famous systems or services use this? Like: https://casdoor.org/docs/organization/overview/
> 
> image

After thinking it over again, I don't think any popular services are using this, so I've decided this change isn't really  good  for merging into casdoor.

PR #2834, opened 2024-03-25
feat: Add 'owner' and 'name' Parameters to /api/get-captcha Response for /api/verify-captcha Usage

u3breeze:
In the response of the /api/get-captcha endpoint, add the parameters "owner" and "name" from captcha provider.  because these two parameters will be used when calling the /api/verify-captcha endpoint:

The vform.ApplicationId is composed of owner and name(owner/name).
```
func (c *ApiController) VerifyCaptcha() {
	var vform form.VerificationForm
	err := c.ParseForm(&vform)
	if err != nil {
		c.ResponseError(err.Error())
		return
	}

	if msg := vform.CheckParameter(form.VerifyCaptcha, c.GetAcceptLanguage()); msg != "" {
		c.ResponseError(msg)
		return
	}

	captchaProvider, err := object.GetCaptchaProviderByOwnerName(vform.ApplicationId, c.GetAcceptLanguage())
	if err != nil {
		c.ResponseError(err.Error())
		return
	}
.....
```

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -44,6 +44,8 @@ type Response struct {
 }
 
 type Captcha struct {
+	Owner         string `json:"owner"`
+	Name          string `json:"name"`
 	Type          string `json:"type"`
 	AppKey        string `json:"appKey"`
 	Scene         string `json:"scene"`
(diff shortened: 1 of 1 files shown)


hsluoyz:
@u3breeze the owner and name in "/api/verify-captcha" API is from provider, not the "/api/get-captcha" API. So your code doesn't work.

image

u3breeze:
@hsluoyz Sure, I understand that /api/get-provider can be used as an alternative to /api/get-captcha, but for external users of the API, /api/get-captcha and /api/verify-captcha seem more intuitive to use together. I've also added 'owner' and 'name' parameters to the /api/get-provider endpoint, which correspond to the owner and name of the captcha provider, and they work as expected. This is my suggestion for optimization, and you can consider whether to make this change.

hsluoyz:
@u3breeze OK

u3breeze:
@hsluoyz Thank you 🎉

Issue #2691, opened 2024-02-11
bug: Google One Tap should be hidden after logged in


hsluoyz:
See: https://door.casdoor.com/ , it should not show up because the user is already logged in

![image](https://github.com/casdoor/casdoor/assets/3787410/e826ab4e-5786-40db-9aaa-d5fec17b3c75)

xyt-dev:
The Google sign-in prompt is likely invoked by the ```useGoogleOneTapLogin()``` function within GoogleLoginButton.js. But the prompt cannot be triggered locally, which consequently make it hard to conduct further tests.

When testing on door.casdoor.com, it's observed that a cookie named ```g_state``` is set automatically when closing the google prompt. This action effectively prevents the prompt from reappearing for a period of time. It might be worth considering the assignment of this ```g_state ``` cookie within the ```renderHomeIfLoggedIn()``` function prior to redirecting to HomePage. Alternatively, using ```window.google.accounts.id.cancel()``` to dismiss the prompt could be explored.

Since I'm unable to trigger the sign-in prompt locally, I cannot verify the effectiveness of the methods.

PR #2852, opened 2024-04-05
fix bug: Google One Tap should be hidden after logged in

xyt-dev:
#2691


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/2853

PR #2904, opened 2024-04-28
Support template rendering on saml assertion

dixyes:
Supports template rendering on custom saml assertion.

For huaweicloud, SAML SSO needs specific assertion, see https://support.huaweicloud.com/usermanual-iam/iam_08_0256.html

so added this

### DIFF
--- a/object/saml_idp.go
+++ b/object/saml_idp.go
@@ -25,9 +25,11 @@ import (
 	"encoding/xml"
 	"errors"
 	"fmt"
+	"html/template"
 	"io"
 	"time"
 
+	"github.com/beego/beego/logs"
 	"github.com/beevik/etree"
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/google/uuid"
@@ -104,10 +106,35 @@ func NewSamlResponse(application *Application, user *User, host string, certific
 	displayName.CreateElement("saml:AttributeValue").CreateAttr("xsi:type", "xs:string").Element().SetText(user.DisplayName)
 
 	for _, item := range application.SamlAttributes {
-		role := attributes.CreateElement("saml:Attribute")
-		role.CreateAttr("Name", item.Name)
-		role.CreateAttr("NameFormat", item.NameFormat)
-		role.CreateElement("saml:AttributeValue").CreateAttr("xsi:type", "xs:string").Element().SetText(item.Value)
+		text := item.Value
+
+		// try use go-template to render it
+		tmpl, err := template.New(item.Name).Parse(item.Value)
+		if err != nil {
+			logs.Warning("failed to parse template on saml attribute %s: %v", item.Name, err)
+		} else {
+			var buf bytes.Buffer
+			err = tmpl.Execute(&buf, map[string]interface{}{
+				"application": application,
+				"user":        user,
+				"host":        host,
+				"certificate": certificate,
+				"destination": destination,
+				"iss":         iss,
+				"requestId":   requestId,
+				"redirectUri": redirectUri,
+			})
+			if err != nil {
+				logs.Warning("failed to execute template on saml attribute %s: %v", item.Name, err)
+			} else {
+				text = buf.String()
+			}
+		}
+
+		customAttr := attributes.CreateElement("saml:Attribute")
+		customAttr.CreateAttr("Name", item.Name)
+		customAttr.CreateAttr("NameFormat", item.NameFormat)
+		customAttr.CreateElement("saml:AttributeValue").CreateAttr("xsi:type", "xs:string").Element().SetText(text)
 	}
 
 	roles := attributes.CreateElement("saml:Attribute")


object/saml_idp.go
```diff
@@ -104,10 +106,35 @@ func NewSamlResponse(application *Application, user *User, host string, certific
...
-		role := attributes.CreateElement("saml:Attribute")
-		role.CreateAttr("Name", item.Name)
-		role.CreateAttr("NameFormat", item.NameFormat)
-		role.CreateElement("saml:AttributeValue").CreateAttr("xsi:type", "xs:string").Element().SetText(item.Value)
+		text := item.Value
+
+		// try use go-template to render it
+		tmpl, err := template.New(item.Name).Parse(item.Value)
+		if err != nil {
+			logs.Warning("failed to parse template on saml attribute %s: %v", item.Name, err)
```
hsluoyz, line 114: Why not return error?

object/saml_idp.go
```diff
@@ -104,10 +106,35 @@ func NewSamlResponse(application *Application, user *User, host string, certific
...
 
 	for _, item := range application.SamlAttributes {
-		role := attributes.CreateElement("saml:Attribute")
-		role.CreateAttr("Name", item.Name)
-		role.CreateAttr("NameFormat", item.NameFormat)
-		role.CreateElement("saml:AttributeValue").CreateAttr("xsi:type", "xs:string").Element().SetText(item.Value)
+		text := item.Value
+
+		// try use go-template to render it
+		tmpl, err := template.New(item.Name).Parse(item.Value)
```
hsluoyz, line 112: Does this affect other SAML vendors? I didn't see any if then code

hsluoyz (changes requested):

object/saml_idp.go
dixyes: for backward compatible: if one use `{{` things in their assertion before, returning error will be a breaking change

object/saml_idp.go
dixyes: What is a "SAML vendor"? this will only affect casdoor as SAML IdP behavior, this do not related to any SAML SP behavior ( at least I cannot find out any relation )

hsluoyz: Does it affect other clouds? AWS, Azure, GCP? Better scope your code within a if branch, so it only affects your cloud

dixyes: 没有理解SAML vendor指什么，

- 如果你在问 这是否影响casdoor作为SAML IdP对其他的下游SAML SP（i.e. 华为云 阿里云 腾讯云）： 可能会影响，如果有人用 `{{` 并且上面选择返回错误，那么旧配置会失效，签发response的时候报错；如果上面不返回错误，但之前有人按照go模板文法配置了字面值（比如之前配置成了 `{{ .user.Phone }}` ），那么assertion会改变（原先是 `{{ .user.Phone }}` 现在是 `12345678910` 这样子）
- 如果你在问 这是否影响casdoor连接上游SAML IdP的行为，这个代码和那部分没啥关系

object/saml_idp.go
dixyes: 我觉得应该不会有人`{{ xxx`这么配 有点太抽象了
所以返回错误 做个BC也是合理的 你看看怎么改合适

object/saml_idp.go
dixyes: > Does it affect other clouds? AWS, Azure, GCP? Better scope your code within a if branch, so it only affects your cloud. I mean it should not fix your cloud's bug while introducing new bugs for other clouds

This is really a generic solution, one can use this template syntax to map any assertion for any SP, even with some if-else logic in template `{{ if host eq "some.sp.com" }}A{{else}}B{{ endif }}`

这是个很通用的解决方案：如果你不需要模板，那么直接字面值就行，如果你需要，这个模板适用于任何SP，你甚至可以在模板里写判断 `{{ if host eq "some.sp.com" }}A{{else}}B{{ endif }}`

dixyes:
Let me explain why use template:

Let's say, there is some strange SAML SP, needs users' birth date in the assertion, so a feature request "Support birth date in SAML assertion" is filed, we add a if-then in that assertion logic
```go
if strings.HasSuffix(host, "somesp.com") {
    // adds assertion
}
```

the second day, another SAML SP needs users' phone:
```go
if strings.HasSuffix(host, "somesp.com") {
    // adds assertion
}
if strings.HasSuffix(host, "somesp2.com") {
    // adds assertion
}
```
after adding that if-thens, the code grows fast and becomes unreadable.

but if we use templates, those SPs just tell user configure their casdoor like
屏幕截图 2024-04-29 183920

you needn't add an assertion mapping function or other SP-specific logics, all SPs needs custom assertion will be supported

BTW, In OIDC, you can also use the template way to make custom claim in JWT-custom.

hsluoyz:
Given your response, I don't think this PR is well tested on other clouds. So for now this PR needs to be closed to avoid introducing new bugs. We try our best to make sure a PR that fixed bug-1 doesn't introduce bug-2, bug-3, etc. Thanks for understanding.

dixyes:
Whatever, we will use that way to support it whether this PR got merged or not.

For others who want to support Huaweicloud, or other custom SAML assertions for any sp, you may use this patch freely.

Issue #2920, opened 2024-05-03
Cannot edit `app-built-in` default application through `init_data.json'


ladung:
Hi guys!
I edited `app-build-in` in init_data.json file as follows
```
...........
"applications": [
      {
        "owner": "admin",
        "name": "app-built-in",
        "displayName": "M y-App-Name",
        "logo": "https://static.wixstatic.com/media/uto/My-logo.png",
        "homepageUrl": "https://mypage",
        "description": "",
        "organization": "built-in",
        "cert": "cert-built-in",
        "enablePassword": true,
....
```
however, when I redeploy  casdoor and login into page, `app-build-in` application still like this:
![image](https://github.com/casdoor/casdoor/assets/43999277/dbf004a1-c56e-4640-9155-a0b57b475e02)
Could you explain it for me or How can I edit the init_data.json file to update default built-in app?

PR #2922, opened 2024-05-04
feat: fix bug that fails to import built-in org via init_data.json

dacongda:
fix: #2920


ladung:
I have a trouble when delete `built-in` orgranization
image

PR #2942, opened 2024-05-10
feat: support Uni SMS

cutedrag0n:


go.mod
```diff
@@ -5,16 +5,19 @@ go 1.16
 require (
 	github.com/Masterminds/squirrel v1.5.3
 	github.com/alexedwards/argon2id v0.0.0-20211130144151-3585854a6387
-	github.com/aws/aws-sdk-go v1.45.5
+	github.com/aliyun/alibaba-cloud-sdk-go v1.62.733 // indirect
```
hsluoyz, line 8: Why update github.com/aliyun/alibaba-cloud-sdk-go?

web/src/ProviderEditPage.js
```diff
@@ -207,7 +207,7 @@ class ProviderEditPage extends React.Component {
         return Setting.getLabel(i18next.t("provider:App key"), i18next.t("provider:App key - Tooltip"));
```
hsluoyz, line 207: You forgot to add it in backend Go code

hsluoyz (changes requested):

go.mod
cutedrag0n: > Why update github.com/aliyun/alibaba-cloud-sdk-go?

To upgrade the sender library, first execute the command `go get -u github.com/casdoor/go-sms-sender`, and then after completing the local development and testing, execute the `go mod tidy` command.

web/src/ProviderEditPage.js
cutedrag0n: > You forgot to add it in backend Go code

What is this parameter for? This platform does not have an option for an app key.

go.mod
hsluoyz: @cutedrag0n I have updated the version, see commit: https://github.com/casdoor/casdoor/commit/3456fc6695f70ab48b4e9c3ee1becdb41b9e9dd0

And I also ran `go mod tidy`, there's no that much diff as yours.

Anyway, you don't need to do this again, as I have did this.

web/src/ProviderEditPage.js
hsluoyz: @cutedrag0n I said you didn't modify the Go code for your new provider, which is wrong. Did it work if you don't modify it? I doubt so.

Issue #2416, opened 2023-10-17
Docker image for arm64?


jum:
I am trying the casdoor docker image on the docker hub and it does not appear to support arm64:

```
! casdoor The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

Is there anything that can be done by that?

hsluoyz:
@jum arm64 build was removed from CI because it's too slow: https://github.com/casdoor/casdoor/pull/2390

The last version that supports arm64 DockerHub build is v1.428.0, you can use this version: https://hub.docker.com/layers/casbin/casdoor/v1.428.0/images/sha256-dca69125720cbf1607dcd2d8e311fbc1b658577d9e85b3475fd5590f9f7b460f?context=repo

https://hub.docker.com/repository/docker/casbin/casdoor/tags?page=1&ordering=last_updated

![image](https://github.com/casdoor/casdoor/assets/3787410/c8b4738b-998d-4869-bece-2be02dc63053)

jum:
Managed to build my own multi-arch container. My observation is that the frontend build is the longest part, so if this would be build only once instead of for each arch it would probably by much faster.

emeryao:
the slow building may related to the cnpm proxy of https://registry.npmmirror.com at https://github.com/casdoor/casdoor/commit/9aec69ef4744f5f49ef8a6605a44a17eada8c65d#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L4

removing the proxy may speed up the building process

Can we please give it another try ?

really need the arm64 docker image

Thanks in advance 😄

emeryao:
@hsluoyz

hsluoyz:
@emeryao not planned to add it back, you can build the image by yourself

hsluoyz:
Thanks to this PR: https://github.com/casdoor/casdoor/pull/2969 , the arm64 image is back starting from v1.621.0: https://hub.docker.com/r/casbin/casdoor/tags

PR #2969, opened 2024-05-25
feat: add back arm64 support in Docker image

AlexanderBabel:
Hey, I saw you had support for ARM64 in the past but dropped it because of [performance concerns](https://github.com/casdoor/casdoor/pull/2390).

I optimized the Dockerfile to significantly speed up the ARM64 build process. Now it should only take around the same time as a build just for AMD64. I did that by specifying the `BUILDPLATFORM`. With that, the GitHub runner does not have to emulate ARM64 during the build of the go binaries and the frontend. Beforehand, every step of the Dockerfile was executed twice. The frontend was built on AMD64 and ARM64, as well as the both go binaries. Now, everything is only built once on AMD64.

[Example pipeline run](https://github.com/AlexanderBabel/casdoor/actions/runs/9235557094/job/25410488736)

|                         | Build Time                                                                 |
| ----------------------- | -------------------------------------------------------------------------- |
| Current build (1.622.0) | [7m6s](https://github.com/casdoor/casdoor/actions/runs/9221672637)         |
| Previous ARM64 build    | [5h12m](https://github.com/casdoor/casdoor/actions/runs/6324483447)        |
| New ARM64 build         | [9m22s](https://github.com/AlexanderBabel/casdoor/actions/runs/9235557094) |

I really hope the team reconsiders the previous standpoint in https://github.com/casdoor/casdoor/issues/2416#issuecomment-2041381467 and considers this Pull Request as a solution for the ARM64 topic. If you have any feedback/adjustment wishes, I am happy to help.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM node:18.19.0 AS FRONT
+FROM --platform=$BUILDPLATFORM node:18.19.0 AS FRONT
 WORKDIR /web
 COPY ./web .
 RUN yarn install --frozen-lockfile --network-timeout 1000000 && yarn run build
 
 
-FROM golang:1.20.12 AS BACK
+FROM --platform=$BUILDPLATFORM golang:1.20.12 AS BACK
 WORKDIR /go/src/casdoor
 COPY . .
 RUN ./build.sh
@@ -13,6 +13,9 @@ RUN go test -v -run TestGetVersionInfo ./util/system_test.go ./util/system.go > 
 FROM alpine:latest AS STANDARD
 LABEL MAINTAINER="https://casdoor.org/"
 ARG USER=casdoor
+ARG TARGETOS
+ARG TARGETARCH
+ENV BUILDX_ARCH="${TARGETOS:-linux}_${TARGETARCH:-amd64}"
 
 RUN sed -i 's/https/http/' /etc/apk/repositories
 RUN apk add --update sudo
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -28,7 +31,7 @@ RUN adduser -D $USER -u 1000 \
 
 USER 1000
 WORKDIR /
-COPY --from=BACK --chown=$USER:$USER /go/src/casdoor/server ./server
+COPY --from=BACK --chown=$USER:$USER /go/src/casdoor/server_${BUILDX_ARCH} ./server
```
hsluoyz, line 34: Why change this folder?

Dockerfile
```diff
@@ -47,12 +50,15 @@ RUN apt update \
...
+ARG TARGETOS
+ARG TARGETARCH
+ENV BUILDX_ARCH="${TARGETOS:-linux}_${TARGETARCH:-amd64}"
 
 RUN apt update
 RUN apt install -y ca-certificates && update-ca-certificates
 
 WORKDIR /
-COPY --from=BACK /go/src/casdoor/server ./server
+COPY --from=BACK /go/src/casdoor/server_${BUILDX_ARCH} ./server
```
hsluoyz, line 61: Why change this folder?

hsluoyz (changes requested):

Dockerfile
AlexanderBabel: This change was already present in the previous implementation. See https://github.com/casdoor/casdoor/commit/9aec69ef4744f5f49ef8a6605a44a17eada8c65d#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L34

The Go build process builds two versions for AMD64 and ARM64. That’s the reason for the line change. The change itself does not change any folder. It’s the name of the binary from the build step that is different. See also https://github.com/casdoor/casdoor/commit/9aec69ef4744f5f49ef8a6605a44a17eada8c65d#diff-4d2a8eefdf2a9783512a35da4dc7676a66404b6f3826a8af9aad038722da6823L11

Dockerfile
AlexanderBabel: The same applies here as well.

Dockerfile
hsluoyz: @AlexanderBabel will this cause any difference in the final amd64 Docker image?

AlexanderBabel: No, the filename of the binary in the final images is unaffected by the change.

AlexanderBabel: You can see that, because the second part of the copy command is still `./server` that is the name of the binary in the final images.

hsluoyz (approved):

Issue #2982, opened 2024-06-02
User's password is still not encrypted after setting the password storage algorithm


seeleclover:
After changing the password storage algorithm of the organization to bcrypt, I immediately created several users and set passwords under this organization. But I saw in the user table of the database that the user's password is still in plain text. May I ask if this is a security problem?

![image](https://github.com/casdoor/casdoor/assets/37256067/c3354faa-517c-47c3-b450-578b5069f561)

seeleclover:
I found in the code that in Line 512 of the file `controllers/user.go` , in the `SetPassword` function, the variable `newPassword` was passed directly to `targetUser.Password` without hash encryption.

![image](https://github.com/casdoor/casdoor/assets/37256067/58caaaea-fdda-4d34-8ee3-eeb5d7968d21)

In addition, I also discovered another issue. In Line 513 of the file, the value of `targetUser.NeedUpdatePassword` is set to `false`, but immediately in Line 515, the `object.UpdateUser` method still takes "need_update_password" as an input parameter. Is there a corresponding relationship here?

dacongda:
> I found in the code that in Line 512 of the file `controllers/user.go` , in the `SetPassword` function, the variable `newPassword` was passed directly to `targetUser.Password` without hash encryption.
> 
[...]

PR #2989, opened 2024-06-04
fix: Fix the bug of changing passwords without encryption

KoiraCMT:
Fix: https://github.com/casdoor/casdoor/issues/2982

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -509,7 +509,19 @@ func (c *ApiController) SetPassword() {
 		return
 	}
 
-	targetUser.Password = newPassword
+	organization, err := object.GetOrganizationByUser(targetUser)
+	if err != nil {
+		c.ResponseError(err.Error())
+		return
+	}
+
+	if organization == nil {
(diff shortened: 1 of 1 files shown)


controllers/user.go
```diff
@@ -509,7 +509,19 @@ func (c *ApiController) SetPassword() {
...
+		return
+	}
+
+	if organization == nil {
+		c.ResponseError(fmt.Sprintf(c.T("the organization: %s is not found"), userOwner))
+		return
+	}
+	targetUser.UpdateUserPassword(organization)
+
+	//targetUser.Password = newPassword
```
dacongda, line 524: have you change the password to the newer one ?

KoiraCMT: Sorry, You are right. 
It should be like this.
```
	targetUser.Password = newPassword
	targetUser.UpdateUserPassword(organization)
```
😂

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/2990

Issue #2981, opened 2024-06-02
[bug] version cannot be displayed in all-in-one docker image


hsluoyz:
Use Docker Option-1 to install Casdoor: https://casdoor.org/docs/basic/try-with-docker#option-1-use-the-toy-database

Go to: http://localhost:8000/sysinfo , the version is empty:

![image](https://github.com/casdoor/casdoor/assets/3787410/847fa569-04f6-43b7-aa2c-684c63868d99)

It should work like the non-Docker demo site: https://door.casdoor.com/sysinfo

![image](https://github.com/casdoor/casdoor/assets/3787410/5db5879a-8451-48d3-9d7e-81815abf89bd)

seeleclover:
I have identified the cause of the problem.

When I used Docker to install Casdoor, I initiated debugging on the `/api/get-version-info` interface and received a return message of `"msg": "repository does not exist"`. This indicates that this interface gets the version by reading the Git History. However, the content packaged into Docker does not include the Git History.

I found in `controllers/systeminfo.go` that although there are two ways to get the version when this interface is called, in any case, when the first method fails, it will immediately return the exception to the front-end.

seeleclover:
I think this interface should determine which method to use to obtain the version based on the `runmode` in `conf/app.conf`. If it is `dev`, then obtain the version through Git History; If it is `prod`, then obtain the version through `version_info.txt`.

seeleclover:
May I ask, how would it be more appropriate to solve this problem?
Or rather, follow the original logic of the program. First, try to obtain the version through Git History. If the retrieval fails, then retrieve the version through the `version_info.txt` ?
@hsluoyz

hsluoyz:
@seeleclover we will fix it by ourselves

seeleclover:
OK, thanks.

PR #2993, opened 2024-06-05
fix: version cannot be displayed in docker image

seeleclover:
fix: #2981 

I think this interface should determine which method to use to obtain the version based on the `runmode` in `conf/app.conf`. If it is `dev`, then obtain the version through Git History; If it is `prod`, then obtain the version through `version_info.txt`.

If users run Casdoor through compilation and in `Production Mode`, they should be required to generate `version_info.txt` after `go build`. I also create a PR in the [Git repository of the Casdoor Document](https://github.com/casdoor/casdoor-website/pull/622).

### DIFF
--- a/controllers/system_info.go
+++ b/controllers/system_info.go
@@ -15,6 +15,7 @@
 package controllers
 
 import (
+	"github.com/casdoor/casdoor/conf"
 	"github.com/casdoor/casdoor/util"
 )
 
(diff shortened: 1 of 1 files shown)


seeleclover:
I think this interface should determine which method to use to obtain the version based on the `runmode` in `conf/app.conf`. If it is `dev`, then obtain the version through Git History; If it is `prod`, then obtain the version through `version_info.txt`.

hsluoyz:
"runmode" is mostly useless, we don't use it in such scenario.

hsluoyz:
@seeleclover as an advice, discuss with the community first before making a PR, to save your time

PR #3002, opened 2024-06-10
feat: support "Use Email as username" in org

### LINKED ISSUE #2756: Default username as email
Hi, how can I auto fill username with email instead of default uuid  if username field is not added to signup table. For other oauth provider like google username is set as email by default.

gregoire-jianda:
Adds the possibility to use the email address as the username if the username field is not visible at signup.
The option can be checked at the organization level.

closes #2756

### DIFF
--- a/web/src/OrganizationEditPage.js
+++ b/web/src/OrganizationEditPage.js
@@ -437,6 +437,16 @@ class OrganizationEditPage extends React.Component {
           </Col>
         </Row>
         <Row style={{marginTop: "20px"}} >
+          <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 19 : 2}>
+            {Setting.getLabel(i18next.t("organization:Use Email as username"), i18next.t("organization:Use Email as username - Tooltip"))} :
+          </Col>
+          <Col span={1} >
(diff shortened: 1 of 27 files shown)


web/src/OrganizationEditPage.js
```diff
@@ -436,6 +436,16 @@ class OrganizationEditPage extends React.Component {
             }} />
           </Col>
         </Row>
+        <Row style={{marginTop: "20px"}} >
+          <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 19 : 2}>
+            {Setting.getLabel(i18next.t("organization:Use email as username"), i18next.t("organization:Use email as username - Tooltip"))} :
```
hsluoyz, line 441: Use email as username -> Use Email as username for everywhere

hsluoyz (changes requested):

hsluoyz:
@gregoire-jianda fix linter:

image

web/src/OrganizationEditPage.js
gregoire-jianda: Done

hsluoyz (approved):

PR #3008, opened 2024-06-19
fix: `passwordRegex` for `maskPassword` being greedy

Sesota:


hsluoyz:
@Sesota what's the difference? Give examples

hsluoyz:
@dacongda greedy is behavior change, not performance. And actually we don't need performance here

hsluoyz:
Actually I tend not to change it here for backward compatibility. You are just causing trouble for us without increasing any benefit.

Sesota:
@hsluoyz sorry for opening this merge request without explanation.

I created this PR because of a bug I noticed when I create/update a user and casdoor sends `new-user`/`update-user` webhook. In the requests sent, the json in `object` key will be malformed. Because the password masking is done by a regex that instead of only matching the password, it matches the rest of the json string. An example will be provided below.

Actual payload:
```json
{..., "password": "foobar", "passwordSalt": "", ...}
```
Masked payload:
```json
{..., "password": "***"
```

PR #3022, opened 2024-06-25
feat: fix bug for access key and secret login

lmk02:
ref https://github.com/casdoor/casdoor/pull/1971

Fix: https://github.com/casdoor/casdoor/issues/3023

It's currently not possible to authenticate by Access key and Access secret if there is no session set with an empty username

### DIFF
--- a/routers/authz_filter.go
+++ b/routers/authz_filter.go
@@ -35,20 +35,13 @@ type Object struct {
 }
 
 func getUsername(ctx *context.Context) (username string) {
-	defer func() {
-		if r := recover(); r != nil {
-			username, _ = getUsernameByClientIdSecret(ctx)
-		}
-	}()
-
-	username = ctx.Input.Session("username").(string)
-
-	if username == "" {
+	username, ok := ctx.Input.Session("username").(string)
+	if !ok || username == "" {
 		username, _ = getUsernameByClientIdSecret(ctx)
 	}
 
 	if username == "" {
-		username = getUsernameByKeys(ctx)
+		username, _ = getUsernameByKeys(ctx)
 	}
 	return
 }
--- a/routers/base.go
+++ b/routers/base.go
@@ -91,17 +91,18 @@ func getUsernameByClientIdSecret(ctx *context.Context) (string, error) {
 	return fmt.Sprintf("app/%s", application.Name), nil
 }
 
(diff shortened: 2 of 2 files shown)


routers/authz_filter.go
```diff
@@ -35,20 +35,13 @@ type Object struct {
...
-	defer func() {
-		if r := recover(); r != nil {
-			username, _ = getUsernameByClientIdSecret(ctx)
-		}
-	}()
-
-	username = ctx.Input.Session("username").(string)
-
-	if username == "" {
+	username, ok := ctx.Input.Session("username").(string)
```
hsluoyz, line 38: Why change here?

routers/authz_filter.go
```diff
@@ -35,20 +35,13 @@ type Object struct {
...
+	if !ok || username == "" {
 		username, _ = getUsernameByClientIdSecret(ctx)
 	}
 
 	if username == "" {
-		username = getUsernameByKeys(ctx)
+		username, _ = getUsernameByKeys(ctx)
```
hsluoyz, line 44: Why not check error?

routers/base.go
```diff
@@ -91,17 +91,22 @@ func getUsernameByClientIdSecret(ctx *context.Context) (string, error) {
...
+		return "", err
+	}
+
+	if user == nil {
+		return "", fmt.Errorf("user not found for access key: %s", accessKey)
 	}
 
-	if user != nil && accessSecret == user.AccessSecret {
-		return user.GetId()
+	if accessSecret != user.AccessSecret {
```
hsluoyz, line 105: Why?

hsluoyz (changes requested):

hsluoyz:
@lmk02 you'd better create an issue first, I don't know what you are doing here

routers/authz_filter.go
lmk02: Because this is causing a runtime panic when there is no username in the session leading to skipping `username = getUsernameByKeys(ctx)`. This part is only reachable when you log out so the username is an empty string.

routers/base.go
lmk02: Thought i would change it to align more with the `getUsernameByClientIdSecret` function above allowing explicit error handling.

hsluoyz (approved):

Issue #3115, opened 2024-08-11
Get user info with accessKey and accessSecret


hamidrabedi:
Can't retrieve current user information with keys:
```
curl -X GET "http://localhost:8000/api/user?owner=customer&accessKey=1d5fc77d-0bcc-4d0e-aa40-f40ad601286b&accessSecret=2c8c0e24-c401-4490-8f8e-6f20a9c47e24"
{"status":"error","msg":"Please login first","sub":"","name":"","data":"Please login first","data2":null}
```

hamidrabedi:
I could add a separate api that checks if accessKey and accessSecret exist in Query, then filter on db to get the corresponding user.

or change old userinfo API:
```
func (c *ApiController) GetUserinfo() {

	user, ok := c.RequireSignedInUser()
	if !ok { // here: && (ctx.request.Query("accessKey") != "" && ctx.request.Query("accessSecret ") != "")
		return
	}
```

PR #3117, opened 2024-08-12
feat: support accessKey and accessSecret login in AutoSigninFilter

dacongda:
fix: #3115


hamidrabedi:
Thanks for the fix, but won't this create a new session on each request? there will be a lot of sessions created which is not necessary at all right? @dacongda

hsluoyz:
@hamidrabedi create a new issue

Issue #3122, opened 2024-08-14
Support Google reCAPTCHA v3 with enterprise api


ViyekShakya:
Enterprise V3: Support all (web, android, iOS)
link: https://developers.google.com/recaptcha/docs/v3

![image](https://github.com/user-attachments/assets/79b74ef9-93f0-41f6-b3dc-5cf13c43a7e2)

![Screenshot from 2024-08-12 15-22-20](https://github.com/user-attachments/assets/76947f8d-2ed6-43f7-b456-d2bb989f74c5)

PR #3130, opened 2024-08-19
feat: support reCAPTCHA v3 captcha provider

ZhaoYP-2001:
fix: https://github.com/casdoor/casdoor/issues/3122

Add reCAPTCHA v3 to the provider type and change the previous type reCAPTCHA to reCAPTCHA v2.

The backend verification logic of google reCAPTCHA v3 and v2 is the same, so the backend code is not modified. Only the captcha operation code is added to the frontend for the v3 version.

### DIFF
--- a/captcha/provider.go
+++ b/captcha/provider.go
@@ -24,7 +24,9 @@ func GetCaptchaProvider(captchaType string) CaptchaProvider {
 	switch captchaType {
 	case "Default":
 		return NewDefaultCaptchaProvider()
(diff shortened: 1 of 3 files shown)


ViyekShakya:
Thank you for the quick resolve. However it would have been better if we could add the back-end verification using API key, instead of using Legacy Secret key, which I found doesn't support Captcha verification for Mobile. We can take this as an enhancement or new feature too.

Thank you once again. @ZhaoYP-2001

PR #3160, opened 2024-09-02
feat: support reCAPTCHA v3 captcha provider

ZhaoYP-2001:


web/src/common/CaptchaWidget.js
```diff
@@ -42,6 +43,32 @@ export const CaptchaWidget = (props) => {
       }, 300);
       break;
     }
+    case "reCAPTCHA v3": {
+      const reTimer = setInterval(() => {
+        if (!window.grecaptcha) {
+          loadScript(`https://recaptcha.net/recaptcha/api.js?render=${siteKey}`);
+        }
+        if (window.grecaptcha && window.grecaptcha.render) {
```
dacongda, line 51: what's different between line 47-69 and line 32-43, can we just use 47-69 or line 32-43?

web/src/common/modal/CaptchaModal.js
```diff
@@ -115,7 +115,7 @@ export const CaptchaModal = (props) => {
     } else {
       return (
         <Col>
-          <Row>
+          <Row style={{justifyContent: "center"}}>
```
dacongda, line 118: Perhaps we can refer to ` https://ant-design.antgroup.com/components/grid#row` for style settings

ZhaoYP-2001: I will modify the implementation method in the code.

web/src/common/CaptchaWidget.js
ZhaoYP-2001: So I don't recommend merging the implementations of reCAPTCHA v2 and reCAPTCHA v3.

dacongda: so reCAPTCHA v2 and v1 use the same code, Maybe we don't need to separate v1 and v2 we just use reCAPTCHA instead

ZhaoYP-2001: In fact, reCAPTCHA v1 has been shut down since March 2018 (https://developers.google.com/recaptcha/docs/versions?hl=zh-cn). Before implementing reCAPTCHA v3, Casdoor used reCAPTCHA to represent reCAPTCHA v2.

In the front-end, I changed the reCAPTCHA option to reCAPTCHA v2 to distinguish it. The code here retains reCAPTCHA for compatibility with the previous database code. But in the front-end Provider page, I deleted the reCAPTCHA option and used reCAPTCHA v2 instead (web/src/Setting.js:1086).

That is, from the user's point of view, the verification code has two versions, v2 and v3, which is in line with the current status of Google reCAPTCHA, but on the back-end, I retained reCAPTCHA for compatibility with the previous database.

Issue #3159, opened 2024-08-31
Bug: Incomplete Logout Process in SSO Single Sign-On (SSO) System


Blackcbears:
We have identified a bug in the Single Sign-On (SSO) system related to the logout process. When a user logs out by sending a request to /api/logout, the system does not always complete the logout correctly, leading to potential security risks.

Steps to Reproduce:

1.Open SSO for casbin/app-saas located in demo casdoor
2.Send a login request to https://demo.casdoor.com/cas/casbin/app-saas/login?service=http://localhost:9000/callback login system.
3.Send a login request to https://demo.casdoor.com/cas/casbin/app-saas/logout?service=http://localhost:9000/callback logout system, `/api/logout` return ok
4.Open Home Page， https://demo.casdoor.com, Description Failed to log out

I tried the test and found logout, but the session in redis was not cleared.

Maybe instead of doing c.ClearUserSession() in `/api/logout`, you do execution c.DestroySession()

ZhaoYP-2001:
> We have identified a bug in the Single Sign-On (SSO) system related to the logout process. When a user logs out by sending a request to /api/logout, the system does not always complete the logout correctly, leading to potential security risks.
> 
> Steps to Reproduce:
> 
> 1.Open SSO for casbin/app-saas located in demo casdoor 2.Send a login request to https://demo.casdoor.com/cas/casbin/app-saas/login?service=http://localhost:9000/callback login system. 3.Send a login request to https://demo.casdoor.com/cas/casbin/app-saas/logout?service=http://localhost:9000/callback logout system, `/api/logout` return ok 4.Open Home Page， https://demo.casdoor.com, Description Failed to log out
> 
> I tried the test and found logout, but the session in redis was not cleared.
> 
> Maybe instead of doing c.ClearUserSession() in `/api/logout`, you do execution c.DestroySession()

I found that this problem is due to the problem of session updates being overwritten when beego has concurrent requests (https://github.com/beego/beego/issues/2666). Our c.ClearUserSession() updates the username in the session to a null value, but it is overwritten by other concurrent requests when the front end is loaded. In other words, although logout returns ok, it is not actually successfully logged out.

I plan to add a callback function to logout, which will query whether the logout has been successful. If not, I will call logout again. Since concurrency only occurs during the initial access and subsequent logouts are not affected by concurrency, I think this can solve your problem.

Do you have any better suggestions?

ZhaoYP-2001:
[...]

PR #3194, opened 2024-09-09
feat: fix CAS logout failure caused by Beego session update problem

ZhaoYP-2001:
fix: https://github.com/casdoor/casdoor/issues/3159

### DIFF
--- a/web/src/auth/CasLogout.js
+++ b/web/src/auth/CasLogout.js
@@ -48,6 +48,7 @@ class CasLogout extends React.Component {
               }
             });
           } else {
+            Setting.showMessage("success", "Logged out successfully");
             this.props.onUpdateAccount(null);
             if (redirectUri !== null && redirectUri !== undefined && redirectUri !== "") {
               Setting.goToLink(redirectUri);
@@ -56,7 +57,6 @@ class CasLogout extends React.Component {
             } else {
               Setting.goToLinkSoft(this, `/cas/${this.state.owner}/${this.state.applicationName}/login`);
             }
-            Setting.showMessage("success", "Logged out successfully");
           }
         });
       }, logoutInterval);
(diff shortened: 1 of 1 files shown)


dacongda (commented):
have you test this function?

web/src/auth/CasLogout.js
```diff
@@ -35,24 +35,53 @@ class CasLogout extends React.Component {
...
+            if (linkSoftFlag) {
+              Setting.goToLinkSoft(this, redirectUri);
+            } else {
+              Setting.goToLink(redirectUri);
+            }
+            Setting.showMessage("success", "Logged out successfully");
+            this.props.onUpdateAccount(null);
+          }
+        });
+      }, s);
```
dacongda, line 67: you pass s to func `logoutTimeOut`, why not replace s with `logoutInterval` ?

dacongda, line 67:
```js
const logoutTimeOut = (redirectUri) => {
      setTimeout(() => {
        AuthBackend.getAccount().then((accountRes) => {
          if (accountRes.status === "ok") {
            AuthBackend.logout().then((logoutRes) => {
              if (logoutRes.status === "ok") {
                const redirectData = logoutRes.data2;
                logoutTimeOut(redirectData);
              } else {
                Setting.showMessage("error", `Failed to log out: ${logoutRes.msg}`);
              }
            });
          } else {
            Setting.showMessage("success", "Logged out successfully");
            this.props.onUpdateAccount(null);
            if (redirectUri !== null && redirectUri !== undefined && redirectUri !== "") {
              Setting.goToLink(redirectUri);
            } else if (params.has("service")) {
              Setting.goToLink(params.get("service"));
            } else {
              Setting.goToLinkSoft(this, `/cas/${this.state.owner}/${this.state.applicationName}/login`);
            }
          }
        });
      },logoutInterval);
```
I think this might be easier to understand.

ZhaoYP-2001: ok, I used a variable name that is not easy to understand and I will correct it in a moment.

ZhaoYP-2001: I updated the timeout implementation. Could you please review the code again?

dacongda (approved):

hsluoyz (approved):

PR #3198, opened 2024-09-12
feat: add an idp for cstnet, that's used inside of branches of cas.cn

liu-kan:
[CSTCloud Passport](https://passport.escience.cn/) is a widely used OAuth authentication tool within the Chinese Academy of Sciences and its affiliated systems. It uses a modified version of the OAuth protocol, with only two https requests required for a single authentication process.


liu-kan:
A new commit is added to pass golangci-lint tests now.

PR #3201, opened 2024-09-13
feat: add an idp for cstnet, that's used inside of branches of cas.cn

liu-kan:
[CSTCloud Passport](https://passport.escience.cn/) is a widely used OAuth authentication tool within the Chinese Academy of Sciences and its affiliated systems. It uses a modified version of the OAuth protocol, with only two https requests required for a single authentication process.

And a fix is added to pass golangci-lint tests after. https://github.com/casdoor/casdoor/pull/3198


hsluoyz:
Not planned to add this as a standalone option. You can use the "Custom OAuth" option to connect.

liu-kan:
> Not planned to add this as a standalone option. You can use the "Custom OAuth" option to connect.
Hello, [CSTCloud Passport](https://passport.escience.cn/) does not offer an endpoint for UserInfoURL, all the info of users comes from extra parts of the token, which means that users can't use the "Custom OAuth" option to connect. In this situation, should I resubmit the pull request?

CSTCloud Passport has been in use for over a decade, serving more than 100,000 people and around 150 organizations.

hsluoyz:
@liu-kan the only way to support it is making the required feature (all the info of users comes from extra parts of the token) available as a general option in "Custom OAuth".

Issue #2415, opened 2023-10-17
Add custom fields to sign up page


ireneeromero:
I need to be able to add custom fields to the registry of my application and not only the default ones. 

More specifically I would need a field that is a dropdown where you can choose all the elements of the dropdown that you want. It would be something similar to the Country/Region field but with the functionality mentioned above.
![image](https://github.com/casdoor/casdoor/assets/61683312/eea1ece9-2da5-4f8b-b6bc-20595bb7f9f9)

Does anyone know if it is possible or how it would be done?

hsluoyz:
@ireneeromero see: https://casdoor.org/docs/application/signup-items-table

ireneeromero:
I had already taken a look at that page, but it does not solve what I want.

I need a field with a custom name, and the possible values to be displayed in a previously defined dropdown. 

For example, a field called "Hobbies" and that the user when registering can choose between playing soccer, watching series, playing tennis, painting or cooking. And that he can choose as many hobbies as he wants, without repeating from the list.

ireneeromero:
I had already taken a look at that page, but it does not solve what I want.

I need a field with a custom name, and the possible values to be displayed in a previously defined dropdown.

For example, a field called "Hobbies" and that the user when registering can choose between playing soccer, watching series, playing tennis, painting or cooking. And that he can choose as many hobbies as he wants, without repeating from the list.

frankie-zeng:
suppose need re-open the issue

hsluoyz:
Support single-choice and multiple-choices signup items

![image](https://github.com/user-attachments/assets/ac392419-04a6-4be1-b610-b65e21ca8e1d)

PR #3211, opened 2024-09-17
feat: add custom signup field

okatu-loli:
Fix: #2415

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -18,6 +18,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/casdoor/casdoor/form"
@@ -88,6 +90,13 @@ func (c *ApiController) Signup() {
 	if err != nil {
 		c.ResponseError(err.Error())
 		return
+	}
+
+	for key := range authForm.CustomFields {
+		if strings.Contains(strconv.Itoa(key), " ") || regexp.MustCompile(`[!@#$%^&*(),.?":{}|<>]`).MatchString(strconv.Itoa(key)) {
+			c.ResponseError(c.T("signup:Custom field names cannot contain spaces or special characters."))
+			return
+		}
 	}
 
 	application, err := object.GetApplication(fmt.Sprintf("admin/%s", authForm.Application))
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -90,16 +90,6 @@ func (c *ApiController) Signup() {
 		return
 	}
 
-	keys := make([]string, 0, len(authForm.CustomFields))
-	for i := range authForm.CustomFields {
-		keys = append(keys, string(authForm.CustomFields[i]))
-	}
-
-	if err := util.ValidateCustomFields(keys); err != nil {
-		c.ResponseError(err.Error())
-		return
-	}
-
 	application, err := object.GetApplication(fmt.Sprintf("admin/%s", authForm.Application))
 	if err != nil {
 		c.ResponseError(err.Error())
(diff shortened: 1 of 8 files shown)


hsluoyz:
@dacongda plz review

dacongda:
can you provide a demo video?

okatu-loli:
> can you provide a demo video?


https://github.com/user-attachments/assets/5a4e6d8c-5131-4d7e-a877-2dba0548f35e

web/src/table/SignupTable.js
```diff
@@ -89,12 +99,20 @@ class SignupTable extends React.Component {
...
 
   renderTable(table) {
+    table = table ?? [];
     const columns = [
       {
         title: i18next.t("general:Name"),
         dataIndex: "name",
         key: "name",
         render: (text, record, index) => {
+          if (record.isCustom) {
```
dacongda, line 109: we can set name field to be disable in line 149

web/src/table/SignupTable.js
```diff
@@ -239,22 +257,25 @@ class SignupTable extends React.Component {
...
-                  }}
-                />
-              </div>
-            } title={i18next.t("application:CSS style")} trigger="click">
-              <Input value={text ? text : SignupTableDefaultCssMap[record.name]} onChange={e => {
-                this.updateField(table, index, "customCss", e.target.value ? e.target.value : SignupTableDefaultCssMap[record.name]);
-              }} />
-            </Popover>
-          );
+          if (!record.name.startsWith("Text ")) {
```
dacongda, line 260: use 
```js
if(record.name.startsWith("Text ") {
    return null;
}
```
will be better

web/src/table/SignupTable.js
```diff
@@ -291,6 +312,33 @@ class SignupTable extends React.Component {
...
         },
       },
+      {
+        title: i18next.t("signup:Custom Item Field"),
+        dataIndex: "customItemField",
+        key: "customItemField",
+        render: (text, record, index) => {
+          const initialValue = text || "";
+          const isMultiple = record.rule === "Multiple Choices";
+          const isSingle = record.rule === "Single Choice";
```
dacongda, line 322: is seems that we can add input or check... in later version, you can check this by use 
```js
["Multiple Choices", "Single Choice"].includes(record.rule)
```

web/src/auth/SignupPage.js
```diff
@@ -198,9 +199,20 @@ class SignupPage extends React.Component {
   onFinish(values) {
     const application = this.getApplicationObj();
 
+    const customFields = {};
+    const customFieldKeys = this.form.current.getFieldsValue(true);
+    for (const [key, value] of Object.entries(customFieldKeys)) {
+      if (key.startsWith("custom_field_")) {
```
dacongda, line 205: we should not allow user to enter string include space or special character , and we should check customField in backend to prevent user add external field

okatu-loli: OK, I have also seen the previous review and I will make changes later.

dacongda: `we should not allow user to enter string include space or special character` this means we we should limit user to input special character to the label field of signup table

okatu-loli:
@dacongda I’ve finished editing the comment. Please review it.

controllers/account.go
```diff
@@ -90,6 +92,13 @@ func (c *ApiController) Signup() {
 		return
 	}
 
+	for key := range authForm.CustomFields {
+		if strings.Contains(strconv.Itoa(key), " ") || regexp.MustCompile(`[!@#$%^&*(),.?":{}|<>]`).MatchString(strconv.Itoa(key)) {
```
dacongda, line 96: compile regex when initial like 
https://github.com/casdoor/casdoor/blob/a330fbc11fd24e6150e6d911f7d230d4ce5d47bc/util/validation.go#L33-L38

dacongda: and have you test this function? authForm.CustomFields looks like a string and you just checking index `01234567...`

dacongda: i don't think this way can validate user's input

okatu-loli: > i don't think this way can validate user's input
I modified the code. The previous code did have some problems. I also adjusted it. Please review it.

okatu-loli: > > i don't think this way can validate user's input

I modified the code. The previous code did have some problems. I also adjusted it. Please review it.

hsluoyz:
Closed for now as it has deviated too far from the original goal

Issue #3232, opened 2024-09-23
MFA : Setup entirely through API


gregoire-jianda:
Hi,

I am trying to implement MFA setup via API, but MfaSetupVerify and MfaSetupEnable both require a session to load key data (totp_secret, recovery_codes). Maybe it could be store in database to allow for API use ?

hsluoyz:
We need to support the stateless usage for MfaSetupVerify and MfaSetupEnable. (the original session usage should be also kept)

gregoire-jianda:
Would it be ok to add : 
- recovery_codes
- country_code
- dest
- totp_secret
as optional form params ?

Since they are part of MfaSetupInitiate response I feel like it's ok to do so. If you agree, I can open a PR

hsluoyz:
@gregoire-jianda sounds good. Can you make a PR?

PR #3239, opened 2024-09-24
feat: allow stateless MFA setup

gregoire-jianda:
Fix: #3232


controllers/mfa.go
```diff
@@ -104,32 +107,38 @@ func (c *ApiController) MfaSetupVerify() {
...
-			c.ResponseError("totp secret is missing")
-			return
-		}
-		config.Secret = secret.(string)
-	} else if mfaType == object.SmsType {
-		dest := c.GetSession(MfaDestSession)
-		if dest == nil {
-			c.ResponseError("destination is missing")
-			return
+		secret := c.Ctx.Request.Form.Get("secret")
```
hsluoyz, line 110: define

```go
c.getSessionOrFormValue(field string) (string, error) {
res:= c.Ctx.Request.Form.Get(field)
if res != nil {
  return res, nil
}

res, ok := c.GetSession(fmt.Sprintf("mfa_%s", field)).(string)
if ok {
  return res, nil
}

return "", fmt.Errorf("MFA field: %s is missing", field)
}
```

Similar changes for all other ocurrances

Similar changes for SetSession, DelSession, etc.

hsluoyz (changes requested):

controllers/mfa.go
gregoire-jianda: I refactored the code. Regarding SetSession and DelSession, I need to determine if the request is indeed stateless, see isSessionOidc.

hsluoyz:
Closed for low quality PR. We may fix it later by ourselves

gregoire-jianda:
@hsluoyz I apologize if my PR didn't meet your standards, would you mind giving me more feedback ? Like a better way to determine if the process is/will be stateless ? Thanks

Issue #3230, opened 2024-09-23
feature: logout if there's no activities in 30 minutes


hsluoyz:
Add a config to app.conf, like logoutMinutes = 30

If there's no activity for the logged-in user in 30 minutes, then Casdoor will log the user out (aka delete session)

PR #3244, opened 2024-09-25
feat: logout if app.conf's inactiveTimeoutMinutes is reached

ZhaoYP-2001:
fix: https://github.com/casdoor/casdoor/issues/3230


hsluoyz:
@dacongda plz review

routers/logout_filter.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+func LogoutFilter(ctx *context.Context) {
+	owner, name := getSubject(ctx)
+	if owner != "anonymous" && name != "anonymous" {
+		sessionId := ctx.Input.CruSession.SessionID()
+		currentTime := time.Now()
+		if cookieTime, exist := cookie2LastTime[sessionId]; exist && cookieTime.Add(logoutMinutes).Before(currentTime) {
+			delete(cookie2LastTime, sessionId)
```
dacongda, line 32: map is not thread safety, maybe we should add lock or use sync.Map to solve it.

dacongda: or we can add last time to session to avoid use map

ZhaoYP-2001: @dacongda I use sync.map to ensure thread safety.
Because all requests need to determine whether they have not been operated for a long time, locking the entire dictionary during the judgment may cause performance bottlenecks. However, beego session data has the problem of update overwriting in concurrent scenarios. Therefore, I did not take these two methods.

conf/app.conf
```diff
@@ -29,4 +29,5 @@ radiusSecret = "secret"
 quota = {"organization": -1, "user": -1, "application": -1, "provider": -1}
 logConfig = {"filename": "logs/casdoor.log", "maxdays":99999, "perm":"0770"}
 initDataFile = "./init_data.json"
-frontendBaseDir = "../casdoor"
\ No newline at end of file
+frontendBaseDir = "../casdoor"
+logoutMinutes = 30
```
hsluoyz, line 33: -1, 0 (<= 0) or empty means no restriction

Put it after `enableGzip`

routers/logout_filter.go
```diff
@@ -0,0 +1,42 @@
+package routers
```
hsluoyz, line 1: Apache header

hsluoyz (changes requested):

routers/logout_filter.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/casdoor/casdoor/conf"
+)
+
+var (
+	logoutMinutes   = time.Minute * 30
+	cookie2LastTime sync.Map
+)
+
+func init() {
+	logoutMinutes_int, err := conf.GetConfigInt64("logoutMinutes")
```
dacongda, line 19: use logoutMinutesInt

hsluoyz: I suggest following the existing code, use string, which supports empty:

```go
func (org *Organization) GetInitScore() (int, error) {
	if org != nil {
		return org.InitScore, nil
	} else {
		return strconv.Atoi(conf.GetConfigString("initScore"))
	}
}
```

conf/app.conf
```diff
@@ -23,6 +23,7 @@ isDemoMode = false
 batchSize = 100
 enableErrorMask = false
 enableGzip = true
+logoutMinutes = 30
```
hsluoyz, line 26: use empty by default

routers/logout_filter.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			logs.Info(fmt.Sprintf("get logoutMinutes failed, err:%v. use default time duration: 30 minutes", err))
+		}
+		logoutMinutes = time.Minute * time.Duration(logoutMinutesInt)
+	}
+}
+
+func LogoutFilter(ctx *context.Context) {
+	owner, name := getSubject(ctx)
+	if owner != "anonymous" && name != "anonymous" {
```
hsluoyz, line 48: return early

if xxxx {
return
}

routers/logout_filter.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var (
+	logoutMinutes   = time.Minute * 30
+	cookie2LastTime sync.Map
+)
+
+func init() {
+	logoutMinutesStr := conf.GetConfigString("logoutMinutes")
+	if logoutMinutesStr == "" {
```
hsluoyz, line 35: <=0 or empty means no restriction

routers/logout_filter.go
```diff
@@ -0,0 +1,61 @@
...
+		logs.Info("get logoutMinutes failed. use default time duration: 30 minutes")
+	} else {
+		logoutMinutesInt, err := strconv.Atoi(logoutMinutesStr)
+		if err != nil {
+			logs.Info(fmt.Sprintf("get logoutMinutes failed, err:%v. use default time duration: 30 minutes", err))
```
hsluoyz, line 40: Use other existing code as I gave last time, don't write by yourself

routers/logout_filter.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+func LogoutFilter(ctx *context.Context) {
+	owner, name := getSubject(ctx)
+	if owner != "anonymous" && name != "anonymous" {
+		sessionId := ctx.Input.CruSession.SessionID()
+		currentTime := time.Now()
+		if cookieTime, exist := cookie2LastTime.Load(sessionId); exist && cookieTime.(time.Time).Add(logoutMinutes).Before(currentTime) {
+			cookie2LastTime.Delete(sessionId)
```
hsluoyz, line 52: Define a new function for logging out action

routers/logout_filter.go
hsluoyz, line 40: Don't use logs

hsluoyz (changes requested):

hsluoyz (approved):

PR #3251, opened 2024-09-27
feat: add verification code step when submit set-password api

sonnht1409:
Add verification step when received set-password call, validate whether target user with verification code is valid with record in verification_record


dacongda (approved):

dacongda:
have you test, could you please provide a demo video?

hsluoyz:
@sonnht1409 you have broken the existing work flow by intercepting with a verification code enforcely

PR #3262, opened 2024-10-02
Use accountItems if part of init_data otherwise use default

cgaspard:
I need the ability to specify accountItems as part of the init_data.json.

This code was blocking me.   This solution allows it to be specified, and if it not, then it uses the default values.


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/3263

Issue #3254, opened 2024-09-28
[feature] support IP limitation for user entry pages (login/signup/forget-password)


hsluoyz:
The IP limitation should be comma separated CIDR list, like `192.168.1.0/24,25.112.0.0/16;232.45.32.86`

The IP limitation field will be available in multiple places: user (the user himself), application (all logins for this application) and organization (all logins for the users in this org).

PR #3267, opened 2024-10-06
feat: support IP limitation for user entry pages

ZhaoYP-2001:
fix: https://github.com/casdoor/casdoor/issues/3254


web/src/Setting.js
```diff
@@ -1557,3 +1557,15 @@ export function getCurrencyText(product) {
...
+  const ipv6CidrPattern = /^((\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?\s*)(\/(([1-9])|([1-9][0-9])|(1[0-1][0-9]|12[0-8]))))*$/;
```
hsluoyz, line 1564: Is it really so long regex?

web/src/UserEditPage.js
```diff
@@ -1070,15 +1070,15 @@ class UserEditPage extends React.Component {
           </Col>
         </Row>
       );
-    } else if (accountItem.name === "Limited ips") {
+    } else if (accountItem.name === "Ip whitelist") {
```
hsluoyz, line 1073: IP whitelist

For UI, it's always "IP"

object/application.go
```diff
@@ -108,6 +108,7 @@ type Application struct {
 	SigninUrl            string     `xorm:"varchar(200)" json:"signinUrl"`
 	ForgetUrl            string     `xorm:"varchar(200)" json:"forgetUrl"`
 	AffiliationUrl       string     `xorm:"varchar(100)" json:"affiliationUrl"`
+	IpWhitelist          string     `xorm:"varchar(1000)" json:"ipWhitelist"`
```
hsluoyz, line 111: varchar 200

routers/router.go
```diff
@@ -66,6 +66,7 @@ func initAPI() {
 	beego.Router("/api/get-webhook-event", &controllers.ApiController{}, "GET:GetWebhookEventType")
 	beego.Router("/api/get-captcha-status", &controllers.ApiController{}, "GET:GetCaptchaStatus")
 	beego.Router("/api/callback", &controllers.ApiController{}, "POST:Callback")
+	beego.Router("/api/check-entry-ip", &controllers.ApiController{}, "GET:CheckEntryIp")
```
hsluoyz, line 69: No need to add new API

hsluoyz (changes requested):

hsluoyz:
@dacongda plz review

web/src/UserEditPage.js
```diff
@@ -1176,6 +1189,10 @@ class UserEditPage extends React.Component {
 
   submitUserEdit(exitAfterSave) {
     const user = Setting.deepCopy(this.state.user);
+    if (user.ipWhitelist !== "" && !Setting.isIpWhitelistValid(user.ipWhitelist)) {
```
hsluoyz, line 1192: Check in Go

hsluoyz (changes requested):

controllers/auth.go
```diff
@@ -55,6 +55,12 @@ func tokenToResponse(token *object.Token) *Response {
 func (c *ApiController) HandleLoggedIn(application *object.Application, user *object.User, form *form.AuthForm) (resp *Response) {
 	userId := user.GetId()
 
+	entryIpCheckError := object.CheckEntryIpByUser(user, c.Ctx.Request.RemoteAddr, c.GetAcceptLanguage())
```
dacongda, line 58: how about signup interface? we have fetch application and organization in this function many times, we should reduce unnecessary database access,

object/check.go
```diff
@@ -33,6 +34,11 @@ const (
 	DefaultFailedSigninFrozenTime = 15
 )
 
+var (
+	IPv4CIDRRegex = regexp.MustCompile(`^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\/([0-9]|[1-2][0-9]|3[0-2]))$/`)
```
dacongda, line 38: you can use `net.ParseCIDR` to check instead of use regex

ZhaoYP-2001: yes, i have changed it just now

controllers/verification.go
```diff
@@ -531,6 +531,35 @@ func (c *ApiController) VerifyCode() {
 		return
 	}
 
+	var application *object.Application
```
dacongda, line 534: we will use this api for other usage, check ip here is not a good idea, and this api can't help user to login

web/src/auth/ForgetPage.js
```diff
@@ -65,7 +65,10 @@ class ForgetPage extends React.Component {
     ApplicationBackend.getApplication("admin", this.state.applicationName)
       .then((res) => {
         if (res.status === "error") {
-          Setting.showMessage("error", res.msg);
+          this.onUpdateApplication(null);
```
dacongda, line 68: why set null in here? and why we need msg to render large message, we haven't this demand in the issue

controllers/application.go
```diff
@@ -110,6 +110,14 @@ func (c *ApiController) GetApplication() {
 		}
 	}
 
+	if application != nil {
```
dacongda, line 113: we should just check api that used to login / signup, now, we should check ip in `Signup() and Login()`

web/src/auth/ForgetPage.js
ZhaoYP-2001: This is for UI design, which allows the front-end to directly display the error message in the entire window after receiving an IP ban error. This implementation method refers to the implementation method of loginPage.js and SignupPage.js.

dacongda: according to 

https://github.com/casdoor/casdoor/blob/347b25676f8a28d06c0a576d1ba1f8ebef80aa6f/web/src/auth/ForgetPage.js#L151-L154

the core api of forgePassword is `SetPassword` in 

https://github.com/casdoor/casdoor/blob/347b25676f8a28d06c0a576d1ba1f8ebef80aa6f/controllers/user.go#L458

controllers/application.go
ZhaoYP-2001: This is to check whether the user's IP meets the rules directly through the application and organization's IP whitelist when the user visits the entry page.
It is equivalent to a pre-check. Compared with this check, the check in login and signup has an additional user's IP whitelist.

web/src/auth/ForgetPage.js
ZhaoYP-2001: I didn't add the ip check in setPassword, I will add that later.

dacongda: > This is for UI design, which allows the front-end to directly display the error message in the entire window after receiving an IP ban error. This implementation method refers to the implementation method of loginPage.js.

so why not use Setting.showMessage("error", res.msg);

controllers/application.go
dacongda: but when we use GetApplication as a restful api, maybe we don't need whitelist function, we just want to check ip in entryPage

ZhaoYP-2001: Could you please explain the restful api in detail?
My current logic is: if the user's IP is not in the whitelist, when it accesses the entryPage to log in, it will not be able to continue to access other pages because of the IP forbidden error returned by the get-application request. In other words, when the user's IP is not in the whitelist, this API will only be used by the entryPage.
So I think we don't need to consider that other scenarios that will use this API will be affected by the IP forbidden error, because when your IP is banned and cannot access the entryPage, you should not be able to continue to access our services.

controllers/verification.go
ZhaoYP-2001: you are right, i have changed the check here to set-password API. This is my mistake. I didn't correctly judge the core API of ForgetPage.

controllers/application.go
ZhaoYP-2001: In my first implementation commit:05c098ff1b690a23e6e131f2450430ab5391f362, I added a check-entry-ip API to check whether the IP meets the whitelist of application and organization.
The check-entry-ip API is only used in LoginPage, SignupPage and ForgetPage, and will be called after the get-application API succeeds. When check-entry-ip returns an error message, the page will jump to the error page.
I think this implementation method may meet your needs, but on the one hand, it makes too many changes to the front end (compared to the current modification of the display method of get-application errors); on the other hand, it brings additional request APIs.

The practice of adding entryIp check to get-application can realize IP check for application and organization whitelists on the one hand, and on the other hand, although this will affect casdoor-sdk's direct call to get-application, I think this impact is tolerable when users cannot log in.

Because if the user can't even log in, there's no point in him calling get-application. And if he wants to check why he can't log in, our error message also gives the specific level (organization/application) and specific name of the whitelist.

dacongda: check ip in CRUD is useless and may cause some issue, you can use a external parameter to decide whether check IP in `GetApplication` interface.

hsluoyz: @ZhaoYP-2001 you should add `bool IsRestricted` field to Application and Organization. `IsRestricted` will be assigned in the Go's REST API handler. Frontend will respect this field and show error if needed

web/src/auth/SignupPage.js
```diff
@@ -812,9 +816,12 @@ class SignupPage extends React.Component {
 
   render() {
     const application = this.getApplicationObj();
-    if (application === undefined || application === null) {
+    if (application === undefined) {
       return null;
     }
+    if (application === null) {
```
hsluoyz, line 822: Why not show error in EntryPage.js ? It's the entrance for pages like signup, login, forget-password, etc. So only one hook is needed

hsluoyz (changes requested):

object/check.go
```diff
@@ -572,3 +578,55 @@ func CheckToEnableCaptcha(application *Application, organization, username strin
 
 	return false, nil
 }
+
+func CheckEntryIp(user *User, application *Application, organization *Organization, remoteAddress string, lang string) error {
```
hsluoyz, line 582: Put your new functions into `check_ip.go`

object/check.go
```diff
@@ -572,3 +578,55 @@ func CheckToEnableCaptcha(application *Application, organization, username strin
...
+	entryIp, _, err := net.SplitHostPort(remoteAddress)
+	if err != nil {
+		return err
+	}
+
+	if user != nil && !isEntryIpAllowd(user.IpWhitelist, entryIp) {
+		return fmt.Errorf(i18n.Translate(lang, "check:Your IP address %s has been banned according to the configuration of user %s"), entryIp, user.Name)
```
hsluoyz, line 589: `Your IP address: %s`

object/check.go
```diff
@@ -572,3 +578,55 @@ func CheckToEnableCaptcha(application *Application, organization, username strin
...
+	}
+
+	if application != nil && !isEntryIpAllowd(application.IpWhitelist, entryIp) {
+		return fmt.Errorf(i18n.Translate(lang, "check:Your IP address %s has been banned according to the configuration of application %s"), entryIp, application.Name)
```
hsluoyz, line 593: `has been banned according to the configuration of: %s"`

So we don't need three i18n texts

web/src/EntryPage.js
```diff
@@ -94,6 +95,14 @@ class EntryPage extends React.Component {
         });
     };
 
+    if (this.state.application?.isRestricted) {
+      return Util.renderMessageLarge(this, `${i18next.t("check:Your IP address has been banned according to the configuration of application")} ${this.state.application.name}`);
```
hsluoyz, line 99: `has been banned according to the configuration of: %s"`

So we don't need three i18n texts

web/src/EntryPage.js
```diff
@@ -94,6 +95,14 @@ class EntryPage extends React.Component {
         });
     };
 
+    if (this.state.application?.isRestricted) {
```
hsluoyz, line 98: Change to Application.BannedIp (string), so we can store the client IP address here to show it in error message

hsluoyz (changes requested):

object/check_ip.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	if organization != nil && !isEntryIpAllowd(organization.IpWhitelist, entryIp) {
+		return fmt.Sprintf(i18n.Translate(lang, "check:Your IP address: %s has been banned according to the configuration of: %s"), entryIp, organization.Name)
+	}
+
+	return ""
+}
+
+func isEntryIpAllowd(ipWhitelistStr string, entryIp string) bool {
```
hsluoyz, line 46: This function returns `fmt.Sprintf(i18n.Translate(lang, "check:Your IP address: %s has been banned according to the configuration of: %s")`

object/check_ip.go
```diff
@@ -0,0 +1,75 @@
...
+	if user != nil && !isEntryIpAllowd(user.IpWhitelist, entryIp) {
+		return fmt.Sprintf(i18n.Translate(lang, "check:Your IP address: %s has been banned according to the configuration of: %s"), entryIp, user.Name)
+	}
+
+	if application != nil && !isEntryIpAllowd(application.IpWhitelist, entryIp) {
+		return fmt.Sprintf(i18n.Translate(lang, "check:Your IP address: %s has been banned according to the configuration of: %s"), entryIp, application.Name)
+	}
+
+	if organization != nil && !isEntryIpAllowd(organization.IpWhitelist, entryIp) {
+		return fmt.Sprintf(i18n.Translate(lang, "check:Your IP address: %s has been banned according to the configuration of: %s"), entryIp, organization.Name)
```
hsluoyz, line 40: return isEntryIpAllowd(xxx)

hsluoyz (changes requested):

controllers/account.go
```diff
@@ -116,6 +116,12 @@ func (c *ApiController) Signup() {
 		return
 	}
 
+	err = object.CheckEntryIp(c.Ctx.Request.RemoteAddr, nil, application, organization, c.GetAcceptLanguage())
```
hsluoyz, line 119: All change to:

```go
clientIp := util.GetClientIpFromRequest(c.Ctx.Request)
err = object.CheckEntryIp(clientIp, nil, application, organization, c.GetAcceptLanguage())
```

object/check_ip.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/casdoor/casdoor/i18n"
+)
+
+func CheckEntryIp(remoteAddress string, user *User, application *Application, organization *Organization, lang string) error {
```
hsluoyz, line 25: clientIp

web/src/table/AccountTable.js
```diff
@@ -104,6 +104,7 @@ class AccountTable extends React.Component {
       {name: "Is forbidden", label: i18next.t("user:Is forbidden")},
       {name: "Is deleted", label: i18next.t("user:Is deleted")},
       {name: "Need update password", label: i18next.t("user:Need update password")},
+      {name: "Ip whitelist", label: i18next.t("general:IP whitelist")},
```
hsluoyz, line 107: For all UI stuff, Ip -> IP

web/src/UserEditPage.js
```diff
@@ -1070,6 +1070,19 @@ class UserEditPage extends React.Component {
           </Col>
         </Row>
       );
+    } else if (accountItem.name === "Ip whitelist") {
```
hsluoyz, line 1073: For all UI stuff, Ip -> IP

object/user_util.go
```diff
@@ -557,6 +557,14 @@ func CheckPermissionForUpdateUser(oldUser, newUser *User, isAdmin bool, lang str
 			itemsChanged = append(itemsChanged, item)
 		}
 	}
+	if oldUser.IpWhitelist != newUser.IpWhitelist {
+		item := GetAccountItemByName("Ip whitelist", organization)
```
hsluoyz, line 561: For all UI stuff, Ip -> IP

hsluoyz (changes requested):

hsluoyz (approved):

Issue #3290, opened 2024-10-16
Refresh Token API does not work when application is shared and user is from different organization than application's.


maknahar:
Thanks for quickly fixing the issue https://github.com/casdoor/casdoor/issues/3288 I raised earlier.

The fix for that issue, seems to have fixed the panic and error is coming out of API in json format than a HTML. However, it did not fix the root cause of the issue. 

Refresh Token API does not work when application is shared and user is from different organization than application. It is returning the error that user does not exist when application is shared and user is from different organization than application's own organization. Application own organization is always `built-in` for shared application.

maknahar:
@hsluoyz I have raised this PR with a potential fix. Please take a look: https://github.com/casdoor/casdoor/pull/3291

hsluoyz:
Closed due to: https://github.com/casdoor/casdoor/pull/3291#issuecomment-2417037436

PR #3291, opened 2024-10-16
feat: fix issue with refresh token when application is shared 

maknahar:
If application is a shared applicaiton, organization of the user can be different from the organization of the application. In this case, we need to get the organization of the user from the token rather than finding the user in the organization of the application.

Fixes: https://github.com/casdoor/casdoor/issues/3290

Tested the changes in local and changes seems to works. Please take a look.

Screenshot 2024-10-16 at 4 52 53 PM

### DIFF
--- a/object/token_oauth.go
+++ b/object/token_oauth.go
@@ -328,10 +328,17 @@ func RefreshToken(grantType string, refreshToken string, scope string, clientId 
 	}
 
 	// generate a new token
-	user, err := getUser(application.Organization, token.User)
+	organization := application.Organization
+	// If application is a shared applicaiton, organization of the user can be different from the organization of the application. In this case, we need to get the organization of the user from the token.
+	if application.IsShared {
+		organization = token.Organization
+	}
+
+	user, err := getUser(organization, token.User)
 	if err != nil {
 		return nil, err
 	}
+
 	if user == nil {
 		return "", fmt.Errorf("The user: %s doesn't exist", util.GetId(application.Organization, token.User))
 	}


object/token_oauth.go
```diff
@@ -328,10 +328,17 @@ func RefreshToken(grantType string, refreshToken string, scope string, clientId
 	}
 
 	// generate a new token
-	user, err := getUser(application.Organization, token.User)
+	organization := application.Organization
```
hsluoyz, line 331: I think we may need a more thorough fix for all such cases - for cross-org users

So define a function like `getUserFromApplication(application *Application)` and use it in all places

hsluoyz (changes requested):

object/token_oauth.go
maknahar: @hsluoyz That make sense. However, I am not that familiar with the code base to do that big refactoring. Also, it will make PR big and difficult to review for you like previous one. 

This is a very critical flow that is broken right now causing user logouts on token refresh as this API is failing. I can take a look at making the separate PR that fixes this issue from other places if that works for you.

hsluoyz: @dacongda

maknahar: Also, I am not sure what you mean by `getUserFromApplication(application *Application)` here. Problem is that user's organization should be assumed to be same as application's organization. In refresh token API, we have user's information from refresh token. I am not sure how we are going to solve this problem by creating a new function `getUserFromApplication(application *Application)`.

maknahar:
@dacongda this is a sample client ID: e326ca828e9992fbcfaa

dacongda:
> @dacongda this is a sample client ID: e326ca828e9992fbcfaa

according to [shared app doc](https://casdoor.org/docs/application/shared-application) you should use {clientID}-org-{org} to specify the org you are using

dacongda:
imho, this issue is caused by incorrectly use and no need to make a pr to fix

hsluoyz:
@dacongda good point. Closed here

maknahar:
@dacongda The document you are referring to asks to update client ID in `login url` where user is sent in browser for authentication in oAuth flow. I am already following that doc and that is how users are getting the auth code and access and refresh token first time. 

Issue here is coming from the refresh token API. Do you mean if I append client with `-org-<orgname>` in the API call, this issue will not occur?

dacongda:
> @dacongda The document you are referring to asks to update client ID in `login url` where user is sent in browser for authentication in oAuth flow. I am already following that doc and that is how users are getting the auth code and access and refresh token first time after token exchange.
> 
> Issue here is coming from the refresh token API. Do you mean if I append client with `-org-<orgname>` in the API call, this issue will not occur?

yes

Issue #3293, opened 2024-10-17
feature: add-user/update-user API should check if username/id/email/phone has duplicated with existing user


hsluoyz:
Should add checks if not

PR #3295, opened 2024-10-18
feat: add-user/update-user API should check if username/id/email/phone has duplicated with existing user

AtCliffUnderline:
Fix: https://github.com/casdoor/casdoor/issues/3293

Not sure should it be in controller or better to move to `check.go`?


AtCliffUnderline:
> @AtCliffUnderline fix CI:
> 
> image

done thx

hsluoyz:
@dacongda plz review

controllers/user.go
```diff
@@ -375,6 +376,20 @@ func (c *ApiController) AddUser() {
 		return
 	}
 
+	if user.Phone != "" {
```
dacongda, line 379: imho, same code should be apply in UpdateUser controller maybe you should pack these code as  a function in check.go name `CheckUserEmailAndPhone`

AtCliffUnderline: object/check.go:532

These checks are already implemented in UpdateUser, however first level `if` is not same, so i am confused how to re-use this code in any other way.

dacongda: emmmm, maybe we can create a empty user object and pass it to function `CheckUpdateUser`

dacongda: Example
```go
func (c *ApiController) AddUser() {
	var user object.User
	err := json.Unmarshal(c.Ctx.Input.RequestBody, &user)
	if err != nil {
		c.ResponseError(err.Error())
		return
	}

	count, err := object.GetUserCount("", "", "", "")
	if err != nil {
		c.ResponseError(err.Error())
		return
	}

	if err := checkQuotaForUser(int(count)); err != nil {
		c.ResponseError(err.Error())
		return
	}

	emptyUser := object.User{}
	if msg := object.CheckUpdateUser(&emptyUser, &user, c.GetAcceptLanguage()); msg != "" {
		c.ResponseError(msg)
		return
	}

	c.Data["json"] = wrapActionResponse(object.AddUser(&user))
	c.ServeJSON()
}
```

AtCliffUnderline: Well, if you think this is OK solution

dacongda (approved):

PR #3311, opened 2024-10-24
fix: fix bug in direct resource URL not updated after updating the storage provider's endpoint

### LINKED ISSUE #3310: bug: resources URL not updated after the qiniu storage provider's endpoint update
before
  Using `http://a.example.com` when creating the qiniu storage provider, the resources uploaded in the Identity-Resources page will have the URL pointing to `http://a.example.com`.

  However, when updating the qiniu storage provider's endpoint later, for example, to `http://bbbb.example.com`, the resources will still point to `http://a.example.com`.

  If I delete the resource and reupload, the new resource's URL still points to `http://a.example.com` instead of `http://bbbb.example.com`.

ithilelda:
fix #3310 

When the Domain property and the Endpoint property doesn't match in a storage provider, the code triggers the existing update path to update Domain to match the new Endpoint.

### DIFF
--- a/object/application.go
+++ b/object/application.go
@@ -723,8 +723,15 @@ func (application *Application) GetId() string {
 }
 
 func (application *Application) IsRedirectUriValid(redirectUri string) bool {
-	redirectUris := append([]string{"http://localhost:", "https://localhost:", "http://127.0.0.1:", "http://casdoor-app", ".chromiumapp.org"}, application.RedirectUris...)
-	for _, targetUri := range redirectUris {
+	isValid, err := util.IsValidOrigin(redirectUri)
+	if err != nil {
+		panic(err)
+	}
+	if isValid {
+		return true
+	}
+
+	for _, targetUri := range application.RedirectUris {
 		targetUriRegex := regexp.MustCompile(targetUri)
 		if targetUriRegex.MatchString(redirectUri) || strings.Contains(redirectUri, targetUri) {
 			return true
--- a/routers/cors_filter.go
+++ b/routers/cors_filter.go
@@ -16,11 +16,11 @@ package routers
 
 import (
 	"net/http"
-	"strings"
 
 	"github.com/beego/beego/context"
 	"github.com/casdoor/casdoor/conf"
 	"github.com/casdoor/casdoor/object"
+	"github.com/casdoor/casdoor/util"
 )
 
 const (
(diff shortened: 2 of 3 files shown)


object/storage.go
```diff
@@ -136,7 +136,7 @@ func getStorageProvider(provider *Provider, lang string) (oss.StorageInterface,
 		return nil, fmt.Errorf(i18n.Translate(lang, "storage:The provider type: %s is not supported"), provider.Type)
 	}
 
-	if provider.Domain == "" {
+	if provider.Domain == "" || provider.Domain != storageProvider.GetEndpoint() {
```
hsluoyz, line 139: Was this any difference with just removing the if branch?

hsluoyz (changes requested):

hsluoyz:
@dacongda plz review

object/storage.go
ithilelda: well with the if statement, you don't have to update the database every time the getStorageProvider is called, only when there's a difference. That might save something depending on the calling frequency. From what I've read in this source file, it shouldn't be much.

hsluoyz: OK, keep the if branch then. Another question is that the impact your change caused to other providers, and why we write like that at the first time.

ithilelda: Looking into the code, I only find two places outside object/storage.go using the provider.Domain property:
https://github.com/casdoor/casdoor/blob/1e0b709c73322300340be9eca8f91972025db20f/object/provider.go#L424-L426
https://github.com/casdoor/casdoor/blob/1e0b709c73322300340be9eca8f91972025db20f/deployment/deploy.go#L51
  The first is not related to storage provider, and the second is. However, since the function getStorageProvider, in which my modification resides, only applies to storage providers, we can safely say that it doesn't affect the first instance, and the impact will only be within the storage providers.

  Further inspection revealed that most storage providers have both the Endpoint and the Domain property input in their edit page:
2024-10-24 16_09_58-Window
except Qiniu and Synology:
2024-10-24 16_09_12-Window
  I think the code with the if check for `provider.Domain == ""` is to ensure that these two storage providers don't bug out when uploading files because the URL generation only uses the Domain property.
 
  In such case, my edit would cause any other storage providers to lose the ability to set a different Endpoint value than the Domain. I'll update my code to find another solution. Maybe I could modify the frontend to add the Domain input box for these two storage providers?

PR #3318, opened 2024-10-25
feat: show Domain field for Qiniu storage provider

### LINKED ISSUE #3310: bug: resources URL not updated after the qiniu storage provider's endpoint update

ithilelda:
allow Qiniu Provider to edit the Domain property in the edit page. I don't know why it's excluded in the first place, but being able to edit it will fix #3310.

Fix: https://github.com/casdoor/casdoor/issues/3310

### DIFF
--- a/web/src/ProviderEditPage.js
+++ b/web/src/ProviderEditPage.js
@@ -908,7 +908,7 @@ class ProviderEditPage extends React.Component {
                 </Col>
               </Row>
             )}
-            {["Custom HTTP SMS", "Qiniu Cloud Kodo", "Synology", "Casdoor"].includes(this.state.provider.type) ? null : (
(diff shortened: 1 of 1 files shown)


hsluoyz:
@dacongda plz review

dacongda (approved):

Issue #3316, opened 2024-10-25
[querstion] User load slowly in RoleEdit page


shiningstarts:
<img width="2244" alt="image" src="https://github.com/user-attachments/assets/b5ffd206-a65c-46d7-ae4f-81fa6466ef9e">
We have 2,000 users , user load slowly in RoleEdit page, the page is also very laggy.   

How can I optimize this page?

shiningstarts:
```js
<Row style={{marginTop: "20px"}} >
  <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 22 : 2}>
    {Setting.getLabel(i18next.t("role:Sub users"), i18next.t("role:Sub users - Tooltip"))} :
  </Col>
  <Col span={22} >
    <Select virtual={false} mode="multiple" style={{width: "100%"}} value={this.state.role.users}
      onChange={(value => {this.updateRoleField("users", value);})}
      options={this.state.users.map((user) => Setting.getOption(`${user.owner}/${user.name}`, `${user.owner}/${user.name}`))}
    />
  </Col>
</Row>
```
When I set virtual=true, it works well

hsluoyz:
@shiningstarts can you make a PR?

PR #3322, opened 2024-10-30
feat: use virtual select UI in role edit page

ljl66-66:
Fix: https://github.com/casdoor/casdoor/issues/3316

using virtual list technology,more data is dynamically loaded. let variable virtual=true

### DIFF
--- a/web/src/RoleEditPage.js
+++ b/web/src/RoleEditPage.js
@@ -147,7 +147,7 @@ class RoleEditPage extends React.Component {
             {Setting.getLabel(i18next.t("general:Organization"), i18next.t("general:Organization - Tooltip"))} :
           </Col>
           <Col span={22} >
-            <Select virtual={true} style={{width: "100%"}} disabled={!Setting.isAdminUser(this.props.account)} value={this.state.role.owner} onChange={(value => {this.updateRoleField("owner", value);})}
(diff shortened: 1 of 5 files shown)


ljl66-66:
In fact, there is another solution that I have tested: similar to lazy loading technology. When entering the character editing page, get a small part of the data by pagination, and then click the drop-down selection box, and then load the remaining data, which can fundamentally solve the problem of slow loading into the character editing page. But after my test, this solution is not effectively compatible with the data before entering the editing page, and after entering the editing page, if you want to get more data in the drop-down list, you will continue to call the backend API, which will also have time loss, affecting the user experience, and also increasing the burden on the backend server. Therefore, I did not adopt this solution.

dacongda (approved):

dacongda:
in face, i think those new changes is match the theme of this pr, plz make a new pr to implement the new feather

ljl66-66:
> in face, i think those new changes is not match the theme of this pr, plz make a new pr to implement the new feather

How is it now?

web/src/RoleEditPage.js
```diff
@@ -187,7 +187,7 @@ class RoleEditPage extends React.Component {
             {Setting.getLabel(i18next.t("role:Sub users"), i18next.t("role:Sub users - Tooltip"))} :
           </Col>
           <Col span={22} >
-            <Select virtual={false} mode="multiple" style={{width: "100%"}} value={this.state.role.users}
+            <Select virtual={true} mode="multiple" style={{width: "100%"}} value={this.state.role.users}
```
dacongda, line 190: remove this line cause it should be a new pr

dacongda (approved):

Issue #3275, opened 2024-10-09
Inconsistent username verification


baznikin:
In my organization turned on `Use Email as username` switch. Users join to organization with invite. Their usernames equal to emails.
If I create user manually I can't set its username to email, got validation error `Failed to save: The username may only contain alphanumeric characters, underlines or hyphens, cannot have consecutive hyphens or underlines, and cannot begin or end with a hyphen or underline.`

1. I propose to allow manually set username to email format.
2. Id like to have ability to set username to user part of email (all before `@`). For example if user have email `name.surname@domain.tld` enforce his username to `name.surname`. We use such format everywhere in organization but I have hard times to get this format out of casdoor. Should I add another issue for this one?

![Screenshot from 2024-10-09 11-06-46](https://github.com/user-attachments/assets/ee0391ac-cc03-46f5-aa44-5c67ff5e9bd8)

hsluoyz:
@baznikin create new issue

hsluoyz:
@dacongda

PR #3329, opened 2024-11-02
feat: Username support email format

ljl66-66:
[issue#3275](https://github.com/casdoor/casdoor/issues/3275)

Resolve user name verification inconsistencies. 
@baznikin  Is this the effect you want?

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -176,8 +176,12 @@ func (c *ApiController) Signup() {
 
 	username := authForm.Username
 	if !application.IsSignupItemVisible("Username") {
-		if organization.UseEmailAsUsername && application.IsSignupItemVisible("Email") {
-			username = authForm.Email
+		if organization.UseEmailAsUsername {
+			if application.IsSignupItemVisible("Email") {
+				username = authForm.Email
+			} else {
+				username = util.GetUsernameFromEmail(authForm.Email)
+			}
 		} else {
 			username = id
 		}
--- a/object/check.go
+++ b/object/check.go
@@ -520,11 +520,38 @@ func CheckUsername(username string, lang string) string {
 	return ""
 }
 
+func CheckUsernameWithEmail(username string, lang string) string {
+	if username == "" {
+		return i18n.Translate(lang, "check:Empty username.")
+	} else if len(username) > 39 {
+		return i18n.Translate(lang, "check:Username is too long (maximum is 39 characters).")
+	}
+
+	// https://stackoverflow.com/questions/58726546/github-username-convention-using-regex
+
+	if !util.ReUserNameWithEmail.MatchString(username) {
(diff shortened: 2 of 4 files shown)


controllers/account.go
```diff
@@ -176,8 +176,12 @@ func (c *ApiController) Signup() {
 
 	username := authForm.Username
 	if !application.IsSignupItemVisible("Username") {
-		if organization.UseEmailAsUsername && application.IsSignupItemVisible("Email") {
-			username = authForm.Email
+		if organization.UseEmailAsUsername {
+			if application.IsSignupItemVisible("Email") {
+				username = authForm.Email
+			} else {
+				username = util.GetUsernameFromEmail(authForm.Email)
```
dacongda, line 183: I think if Email is invisible in signup item, authForm.Email will be empty, so why pass any empty string to this function and get some random string? It will not generate a virtual email address .

web/src/UserListPage.js
```diff
@@ -64,7 +64,7 @@ class UserListPage extends BaseListPage {
     const owner = (Setting.isDefaultOrganizationSelected(this.props.account) || this.props.groupName) ? this.state.organizationName : Setting.getRequestOrganization(this.props.account);
     return {
       owner: owner,
-      name: `user_${randomName}`,
+      name: this.state.organization["useEmailAsUsername"] ? `${randomName}` : `user_${randomName}`,
```
dacongda, line 67: same problem, you have set use EmailAsUsername, and use a none email format string

object/check.go
```diff
@@ -520,11 +520,38 @@ func CheckUsername(username string, lang string) string {
...
+	}
+	return ""
+}
+
 func CheckUpdateUser(oldUser, user *User, lang string) string {
 	if oldUser.Name != user.Name {
-		if msg := CheckUsername(user.Name, lang); msg != "" {
-			return msg
+		// Check if the organization uses email as the username
+		organization, err := getOrganization("admin", oldUser.Owner)
```
hsluoyz, line 541: use GetOrganizationByUser

object/check.go
```diff
@@ -520,11 +520,38 @@ func CheckUsername(username string, lang string) string {
...
+		if err != nil {
+			panic(err)
```
hsluoyz, line 543: return error instead of panic

object/check.go
```diff
@@ -520,11 +520,38 @@ func CheckUsername(username string, lang string) string {
...
 		}
```
hsluoyz, line 544: check organization == nil

hsluoyz (changes requested):

hsluoyz (approved):

JosimarGomes:
@hsluoyz I'm using the Casdoor application with public API and today I started receiving the error: "does not exist" when registering a new user (/api/add-user). Can you check if this has anything to do with this merge? If you need more information, just let me know.

PR #3332, opened 2024-11-04
fix: enforce `email` and `phone` data on `signup`

Danipulok:
**Current behavior:**
When email is setup like following:
```json
{
   "name":"Email",
   "visible":true,
   "required":true,
   "prompted":false,
   "type":"",
   "customCss":"",
   "label":"",
   "placeholder":"",
   "options":null,
   "regex":"",
   "rule":"Normal"
}
```

And the following request is sent:
```bash
curl --location 'baseUrl/api/signup' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'Accept: */*' \
--data-urlencode 'username=my_org/username' \
--data-urlencode 'password=CoolPassword420'
```

The response is:
```
{
  "status": "error",
  "data": null,
  "data2": null
}
```

**Expected behavior:**
Error is raised since email data is not provided.

**Description:**
This PR enforces `phone` and `email` data check, even if data is not passed.

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -139,7 +139,7 @@ func (c *ApiController) Signup() {
 		invitationName = invitation.Name
 	}
 
-	if application.IsSignupItemVisible("Email") && application.GetSignupItemRule("Email") != "No verification" && authForm.Email != "" {
+	if application.IsSignupItemVisible("Email") && application.GetSignupItemRule("Email") != "No verification" {
 		var checkResult *object.VerifyResult
 		checkResult, err = object.CheckVerificationCode(authForm.Email, authForm.EmailCode, c.GetAcceptLanguage())
 		if err != nil {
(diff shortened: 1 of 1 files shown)


controllers/account.go
```diff
@@ -139,7 +139,7 @@ func (c *ApiController) Signup() {
 		invitationName = invitation.Name
 	}
 
-	if application.IsSignupItemVisible("Email") && application.GetSignupItemRule("Email") != "No verification" && authForm.Email != "" {
```
dacongda, line 142: but in some case, Email or Phone number is not required, we should consider those situation

Danipulok: Doesn't `application.GetSignupItemRule("Email") != "No verification"` check for this?
It can be either `No verification` or `Normal` in UI

dacongda: ok

dacongda (approved):

hsluoyz:
@Danipulok we allow users to register without Email/phone. Some people want it.

@Danipulok from the several PRs you have created, it seems that you only cared about your requirement (maybe a security requirement to obfuscate error strings, I don't know), but you clearly should care about all other use cases. It's not your personal "Security Enhanced" SE version. It's a OSS repo to cover wild general-purpose needs.

Danipulok:
@hsluoyz , thank you for your review!
I want the generic solution that works for everyone.
As I understood from the code, I did not force requirements for `email` and `phone` for all users.
Just when the admin of organization set it as required, as it works now on UI.
Is it not correct?

PR #3335, opened 2024-11-04
fix: respect `enableErrorMask` in `AutoSigninFilter`

Danipulok:
**Current behavior:**
When `AutoSigninFilter` is getting used (so, always), and `username` and `password` are passed in query params or some other way, it tries to autologin it. But when it fails, it reveals information if user with data like this exists.

_Request:_
```bash
curl --location --request POST 'https://auth.stage.aliyoop.com/api/I_DONT_EXIST?username=org%2Fusername&password=password'
```

_Response:_
```json
{
  "status": "error",
  "data": null,
  "data2": null
}
```

The error is raised [here](https://github.com/casdoor/casdoor/blob/457c6098a49ea6c9bb3fa328807103175effcead/object/check.go#L343-L345).
It leads to having a possibility of enumeration users and knowing if user with username/phone/emails exists.

**PR behavior:**
Now when using `enableErrorMask` param, the error is mapped to `check:password or code is incorrect`, if username is invalid OR password is invalid and `failedSigninLimit` is enabled.
Same as in from [controllers\util.go](https://github.com/casdoor/casdoor/blob/457c6098a49ea6c9bb3fa328807103175effcead/controllers/util.go#L57-L62).

### DIFF
--- a/object/check.go
+++ b/object/check.go
@@ -21,6 +21,7 @@ import (
 	"time"
 	"unicode"
 
+	"github.com/casdoor/casdoor/conf"
 	"github.com/casdoor/casdoor/cred"
 	"github.com/casdoor/casdoor/form"
 	"github.com/casdoor/casdoor/i18n"
@@ -341,6 +342,10 @@ func CheckUserPassword(organization string, username string, password string, la
 	}
 
 	if user == nil || user.IsDeleted {
+		enableErrorMask := conf.GetConfigBool("enableErrorMask")
+		if enableErrorMask {
+			return nil, fmt.Errorf(i18n.Translate(lang, "check:password or code is incorrect"))
+		}
 		return nil, fmt.Errorf(i18n.Translate(lang, "general:The user: %s doesn't exist"), util.GetId(organization, username))
 	}
 
--- a/object/check_util.go
+++ b/object/check_util.go
@@ -19,6 +19,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/casdoor/casdoor/conf"
 	"github.com/casdoor/casdoor/i18n"
 )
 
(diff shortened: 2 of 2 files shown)


dacongda:
lgtm, but is there any place that need to use enableErrorMask ?  Fix them in one pr will be better.

Danipulok:
@dacongda do you have maybe a list of places / errors that must be masked?
We receive only following errors, so I don't quite know what other places / errors must be masked

dacongda (approved):

hsluoyz:
@Danipulok Can't you do your job here? We don't want to add a lot of those checks in different places

image

Danipulok:
@hsluoyz , `AutoSigninFilter` does not use `ApiController`, so I guess the similar solution of using `ApiController ResponseError` would be to mask the error in `responseError` in [base.go](https://github.com/casdoor/casdoor/blob/457c6098a49ea6c9bb3fa328807103175effcead/routers/base.go#L37-L53).

Would it be better?

But then it will have a lot of `if` statements, since the error will be already translated, as in [ResponseError](https://github.com/casdoor/casdoor/blob/457c6098a49ea6c9bb3fa328807103175effcead/controllers/util.go#L57-L62). Is it considered okay?

hsluoyz:
Not planned, as it provides no clear gains

Danipulok:
@halozhy , currently, anyone can send multiple requests and understand if a user exists in a database and brute force the password.
It is considered okay and does not have enough gain?

hsluoyz:
@Danipulok maybe this is worth a low-level or medium-level issue. But your code is not a good choice. It requires someone more senior and qualified to do it.

Danipulok:
@halozhy , could you please explain what exactly in my solution is not good enough?
I proposed a generic option too, with minimal changes to the current code and project structure

PR #3336, opened 2024-11-05
feat: improve Vietnamese translations for password complexity

giangdinhtt:
Proper Vietnamese translations for password complexity


hsluoyz:
@giangdinhtt can you translate all Vietnam words with ChatGPT? It's too trivial to just add several i18n strings

PR #3341, opened 2024-11-08
feat: Adding contents of memberOf fields with connecting via LDAP

ysirotinskii:
Hi, could you please look at the PR, the fixes include the addition and aggregation of MemberOf elements, which we receive from an external LDAP server

Looking forward to PR approval

### DIFF
--- a/object/ldap_conn.go
+++ b/object/ldap_conn.go
@@ -52,10 +52,10 @@ type LdapUser struct {
 	MobileTelephoneNumber string
 	RegisteredAddress     string
 	PostalAddress         string
-
-	GroupId  string `json:"groupId"`
-	Address  string `json:"address"`
-	MemberOf string `json:"memberOf"`
+	GroupId  	      string   `json:"groupId"`
+	Address               string   `json:"address"`
+	MemberOf              string   `json:"memberOf"`
+	Member                []string `json:"member"`
 }
 
 func (ldap *Ldap) GetLdapConn() (c *LdapConn, err error) {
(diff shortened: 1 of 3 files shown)


hsluoyz:
@ysirotinskii first pass all CIs, it will not go to review stage before you pass all CIs

object/ldap_conn.go
```diff
@@ -202,7 +202,8 @@ func (l *LdapConn) GetLdapUsers(ldapServer *Ldap) ([]LdapUser, error) {
 			case "postalAddress":
 				user.PostalAddress = attribute.Values[0]
 			case "memberOf":
-				user.MemberOf = attribute.Values[0]
+				fmt.Printf("Synchronized user %s: %v\n", user, attribute.Values)
```
hsluoyz, line 205: Don't print to console

object/user.go
```diff
@@ -47,11 +47,12 @@ func InitUserManager() {
...
-	Name        string `xorm:"varchar(100) notnull pk" json:"name"`
-	CreatedTime string `xorm:"varchar(100) index" json:"createdTime"`
-	UpdatedTime string `xorm:"varchar(100)" json:"updatedTime"`
-	DeletedTime string `xorm:"varchar(100)" json:"deletedTime"`
+	Owner       string   `xorm:"varchar(100) notnull pk" json:"owner"`
+	Name        string   `xorm:"varchar(100) notnull pk" json:"name"`
+	CreatedTime string   `xorm:"varchar(100) index" json:"createdTime"`
+	UpdatedTime string   `xorm:"varchar(100)" json:"updatedTime"`
+	DeletedTime string   `xorm:"varchar(100)" json:"deletedTime"`
+	Member      []string `json:"member"`
```
hsluoyz, line 55: You should not add new field. What's member? Is it the groups the user belongs to? We already have groups field

hsluoyz (changes requested):

dacongda:
after research, memberOf seems to be a optional module of ldap and it can be used to find the group via user, I don't think it can be a independence field of user.

hsluoyz:
@ysirotinskii like @dacongda said, the problem you are trying to resolve is meaningful. But your code quality is below the bar. It needs a more skilled person to do the job and make the PR. So closed here for now.

PR #3343, opened 2024-11-12
feat: validate token before get from DB, use GetTokenByValueOrId instead of GetTokenByValue

notdu:
Optimize token validation and retrieval
- Validate the token before retrieving it from the database to reduce unnecessary database calls.
- Replace `GetTokenByValue` with `GetTokenByValueOrId` for token retrieval, as querying by token value (text) is inefficient for indexing and querying.

### DIFF
--- a/object/token.go
+++ b/object/token.go
@@ -116,6 +116,11 @@ func GetTokenByAccessToken(accessToken string) (*Token, error) {
 	return &token, nil
 }
 
+func GetTokenByAccessTokenOrId(accessToken, tokenId string) (*Token, error) {
+	token := Token{AccessTokenHash: getTokenHash(accessToken)}
+	return GetTokenByTokenOrId(token, tokenId)
+}
+
 func GetTokenByRefreshToken(refreshToken string) (*Token, error) {
 	token := Token{RefreshTokenHash: getTokenHash(refreshToken)}
 	existed, err := ormer.Engine.Get(&token)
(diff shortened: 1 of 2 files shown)


hsluoyz:
@dacongda plz review

object/token.go
```diff
@@ -137,18 +159,39 @@ func GetTokenByRefreshToken(refreshToken string) (*Token, error) {
 	return &token, nil
 }
 
-func GetTokenByTokenValue(tokenValue, tokenTypeHint string) (*Token, error) {
+func GetTokenByRefreshTokenOrName(refreshToken, tokenName string) (*Token, error) {
```
dacongda: this function is totally  same with `GetTokenByAccessTokenOrName`, it should be remove and name as `GetTokenByTokenOrName`

notdu:
Hi @dacongda , I’ve made the changes and force-pushed them. Could you take another look when you have a moment?

dacongda (changes requested):

object/token.go
dacongda: you can just pack the same code to a new func instead of remove them

notdu: hi @dacongda , i've just updated

dacongda: > hi @dacongda , i've just updated

I know, and those code is same with previous one,  I just suggest that make the same part between `GetTokenByRefreshTokenOrName` and GetTokenByAccessTokenOrName to be a new function

notdu: @dacongda, these two methods are distinct: one retrieves tokens by RefreshTokenHash, while the other retrieves them by AccessTokenHash

dacongda:
```go

		owner, name := util.GetOwnerAndNameFromIdNoCheck(tokenId)
		token = Token{Owner: owner, Name: name}
		existed, err = ormer.Engine.Get(&token)
		if err != nil {
			return nil, err
		}
	
```

these line can be a new function named `getTokenByTokenId`

notdu: hi @dacongda, i've updated

dacongda (approved):
lgtm

hsluoyz:
@notdu I don't think your changes are reasonable. DB query is already indexed, already very fast, no need to improve. Even if the attacker wants to access the login page of Casdoor, it requires several DB queries like get-application. DB queries are inevitable.

And the token ID is also hard to understand, making code ugly. Don't know what you are trying to resolve.

First make issue first.

Issue #3376, opened 2024-11-25
LDAP server returns incorrect attribute for user name (cn instead of title)


keisni:
Description
When applications integrate with Casdoor's LDAP server, they cannot properly retrieve user names because the LDAP server is incorrectly mapping the user name to the cn attribute instead of the standard title attribute.

Current Behavior
In ldap/server.go, line 145:
if string(attr) == "cn" {

The user's title is being mapped to the cn attribute in LDAP responses.

Expected Behavior
The user's name should be mapped to the cn attribute:

Proposed Solution
Change the const string from cn to title in ldap/server.go line 45 .

PR #3378, opened 2024-11-25
fix: change LDAP attribute from cn to title for correct username mapping

keisni:
**Problem**
When integrating applications with Casdoor's LDAP server, the user's name is not correctly retrieved because it's currently mapped to the `title` attribute instead of the standard `cn` attribute.

**Solution**
Change the attribute mapping from `cn` to `title` in the LDAP search response.

**Changes**
#3376 

**Testing**
1. Started Casdoor with LDAP server enabled
2. Connected an LDAP client application
3. Verified that user names are now correctly retrieved using the `cn` attribute


hsluoyz:
@dacongda plz review

dacongda (approved):

Issue #3371, opened 2024-11-23
AI assistant in MFA page


hamidrabedi:
Allow to remove the AI assistant in MFA configuration page.

![image](https://github.com/user-attachments/assets/160dffbe-50a8-4ee0-876b-c00704801022)

hsluoyz:
@hamidrabedi why remove?

hamidrabedi:
Just allow to remove it easily, either in configurations or App or Org conf. users can see this page when they want to configure their MFA, therefore everyone can see and click on it. First of all the tab is not personalized and its showing hard coded urls and images, and also some users can not even load it because of restrictions:
![image](https://github.com/user-attachments/assets/d83e66c4-7a97-4ee1-8d03-d6221fb0fd9a)

hsluoyz:
Currently, the AI chatbot URL is hard-coded in: https://github.com/search?q=repo%3Acasdoor%2Fcasdoor%20https%3A%2F%2Fai.casbin.com&type=code

We should make it a configurable item in: https://github.com/casdoor/casdoor/blob/master/web/src/Conf.js

How to do it?

Add `AiAssistantUrl`, the default value is `https://ai.casbin.com`. If it's empty value, hide the button in web UI.

PR #3385, opened 2024-11-29
feat: add AiAssistantUrl to frontend config

imp2002:
fix: #3371 

https://github.com/casdoor/casdoor/issues/3371#issuecomment-2499379833


dacongda:
I think put ai assistant url setting in org setting could be better

hsluoyz:
@dacongda this is the first step, can be improved later

Issue #3389, opened 2024-11-30
[feature] support LDAPS in TCP port 636


hsluoyz:
Now Casdoor already supported LDAP in TCP port 389, we should also support LDAPS in TCP port 636. 

Similar to: https://github.com/casdoor/casdoor/issues/2555

How to do it?

Casdoor should handle the SSL cert like CasWAF: https://github.com/casbin/caswaf/blob/8b6f4f779a971b13b3f2b22ccff9ef162e8aea45/service/proxy.go#L280 , then unwrap the SSL shell and redirect to Casdoor's LDAP port (aka 389). The SSL cert can be stored in Casdoor's Certs page: https://door.casdoor.com/certs

PR #3390, opened 2024-12-01
feat: support LDAPS protocol

dacongda:
Fix: https://github.com/casdoor/casdoor/issues/3389


ldap/server.go
```diff
@@ -27,21 +28,65 @@ import (
...
+	}()
+}
+
+func getTLSconfig() (*tls.Config, error) {
+	ldapsCertId := conf.GetConfigString("ldapsCertId")
+	if ldapsCertId == "" {
+		return nil, fmt.Errorf("ldapsCertId is empty")
+	}
+	rawCert, err := object.GetCert(ldapsCertId)
 	if err != nil {
```
hsluoyz, line 77: check rawCert == nil

ldap/server.go
```diff
@@ -27,21 +28,65 @@ import (
...
+		}
+		err = serverSsl.ListenAndServe("0.0.0.0:"+ldapsServerPort, secureConn)
+		if err != nil {
+			log.Printf("StartLdapsServer() failed, err = %s", err.Error())
+		}
+	}()
+}
+
+func getTLSconfig() (*tls.Config, error) {
+	ldapsCertId := conf.GetConfigString("ldapsCertId")
```
hsluoyz, line 72: Will it work if ldapsCertId is empty? If not, LDAPS should not be enabled if ldapsCertId is empty

hsluoyz (changes requested):

ldap/server.go
dacongda: 636 will not be listen when ldapCertId is empty

hsluoyz (approved):

Issue #3392, opened 2024-12-02
[Feature] Reset Password to Recognize LDAP Users and Modify LDAP User Password Instead of the Local Database


Xing-Fax:
When using LDAP as the user source, I found that while logging in, the LDAP server can be associated for password authentication. However, resetting the password does not trigger the LDAP server to reset the password—it only resets the local user database. This is not what the user expects. It leads to situations where users are unable to log in after resetting their password on the login page because the password correctness is determined by LDAP, not the local database.

If this feature is acceptable, I will submit a Pull Request. Thank you.

PR #3395, opened 2024-12-03
feat: support LDAP's SetPassword

Xing-Fax:
Fix: https://github.com/casdoor/casdoor/issues/3392

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -481,7 +481,7 @@ func (c *ApiController) SetPassword() {
 		return
 	}
 	if user == nil {
-		c.ResponseError("User doesn't exist")
+		c.ResponseError(fmt.Sprintf(c.T("general:The user: %s doesn't exist"), userId))
 		return
 	}
 
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -475,7 +475,11 @@ func (c *ApiController) SetPassword() {
 
 	userId := util.GetId(userOwner, userName)
 
-	user, _ := object.GetUser(userId)
+	user, err := object.GetUser(userId)
+	if err != nil {
+		c.ResponseError(err.Error())
+		return
+	}
 
 	requestUserId := c.GetSessionUsername()
 	if requestUserId == "" && code == "" {
@@ -522,16 +526,12 @@ func (c *ApiController) SetPassword() {
 	} else if code == "" {
 		if user.Ldap == "" {
 			err = object.CheckPassword(targetUser, oldPassword, c.GetAcceptLanguage())
-			if err != nil {
-				c.ResponseError(err.Error())
-				return
-			}
 		} else {
 			err = object.CheckLdapUserPassword(targetUser, oldPassword, c.GetAcceptLanguage())
-			if err != nil {
-				c.ResponseError(err.Error())
-				return
-			}
+		}
+		if err != nil {
+			c.ResponseError(err.Error())
+			return
 		}
 	}
 
@@ -575,16 +575,13 @@ func (c *ApiController) SetPassword() {
 
 	if user.Ldap == "" {
 		_, err = object.UpdateUser(userId, targetUser, []string{"password", "need_update_password", "password_type", "last_change_password_time"}, false)
-		if err != nil {
-			c.ResponseError(err.Error())
-			return
-		}
 	} else {
-		err := object.ResetLdapPassword(targetUser, newPassword)
-		if err != nil {
-			c.ResponseError(err.Error())
-			return
-		}
+		err = object.ResetLdapPassword(targetUser, newPassword, c.GetAcceptLanguage())
+	}
+
+	if err != nil {
+		c.ResponseError(err.Error())
+		return
 	}
 
 	c.ResponseOk()
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -475,6 +475,8 @@ func (c *ApiController) SetPassword() {
 
 	userId := util.GetId(userOwner, userName)
 
+	user, _ := object.GetUser(userId)
+
 	requestUserId := c.GetSessionUsername()
 	if requestUserId == "" && code == "" {
 		c.ResponseError(c.T("general:Please login first"), "Please login first")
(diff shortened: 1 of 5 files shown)


hsluoyz:
@dacongda plz review

object/check.go
```diff
@@ -273,6 +273,10 @@ func CheckPasswordComplexity(user *User, password string) string {
 	return CheckPasswordComplexityByOrg(organization, password)
 }
 
+func CheckLdapUserPassword(user *User, password string, lang string) error {
```
dacongda, line 276: you can change `checkLdapUserPassword` to `CheckLdapUserPassword` instead of create a new function

controllers/user.go
```diff
@@ -563,10 +573,18 @@ func (c *ApiController) SetPassword() {
 	targetUser.NeedUpdatePassword = false
 	targetUser.LastChangePasswordTime = util.GetCurrentTime()
 
-	_, err = object.UpdateUser(userId, targetUser, []string{"password", "need_update_password", "password_type", "last_change_password_time"}, false)
-	if err != nil {
-		c.ResponseError(err.Error())
-		return
+	if user.Ldap == "" {
```
dacongda, line 576: how about user want login with password store in casdoor? maybe we can create a extra button named `Resert LDAP password`

dacongda (changes requested):

controllers/user.go
Xing-Fax: The password authentication source for LDAP users can only come from the LDAP server. Even if a password is configured for the user in the database, the login verification will still be associated with LDAP. Therefore, the local password is redundant for the user, and since users do not know whether they are LDAP users, this may cause confusion.

dacongda: ok

object/check.go
Xing-Fax: Has been adjusted.

object/ldap_conn.go
```diff
@@ -371,6 +372,64 @@ func GetExistUuids(owner string, uuids []string) ([]string, error) {
...
+			conn.Close()
+			return err
+		}
+
+		userDN := searchResult.Entries[0].DN
+
+		var pwdEncoded string
+		passwordModify := goldap.NewModifyRequest(userDN, nil)
+		if conn.IsAD {
+			utf16 := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)
```
dacongda, line 410: have you test these line with MicrosoftAD?

dacongda (approved):

controllers/user.go
```diff
@@ -475,6 +475,8 @@ func (c *ApiController) SetPassword() {
 
 	userId := util.GetId(userOwner, userName)
 
+	user, _ := object.GetUser(userId)
```
hsluoyz, line 478: why not check error?

controllers/user.go
```diff
@@ -518,10 +520,18 @@ func (c *ApiController) SetPassword() {
 			}
 		}
 	} else if code == "" {
-		err = object.CheckPassword(targetUser, oldPassword, c.GetAcceptLanguage())
-		if err != nil {
-			c.ResponseError(err.Error())
-			return
+		if user.Ldap == "" {
+			err = object.CheckPassword(targetUser, oldPassword, c.GetAcceptLanguage())
+			if err != nil {
```
hsluoyz, line 525:
```
if err != nil {
	c.ResponseError(err.Error())
	return
}
```

should be outside

controllers/user.go
```diff
@@ -563,10 +573,18 @@ func (c *ApiController) SetPassword() {
 	targetUser.NeedUpdatePassword = false
 	targetUser.LastChangePasswordTime = util.GetCurrentTime()
 
-	_, err = object.UpdateUser(userId, targetUser, []string{"password", "need_update_password", "password_type", "last_change_password_time"}, false)
-	if err != nil {
-		c.ResponseError(err.Error())
-		return
+	if user.Ldap == "" {
+		_, err = object.UpdateUser(userId, targetUser, []string{"password", "need_update_password", "password_type", "last_change_password_time"}, false)
+		if err != nil {
```
hsluoyz, line 578:
```
if err != nil {
	c.ResponseError(err.Error())
	return
}
```

should be outside

object/ldap_conn.go
```diff
@@ -371,6 +372,64 @@ func GetExistUuids(owner string, uuids []string) ([]string, error) {
...
+			return err
+		}
+
+		if len(searchResult.Entries) == 0 {
+			conn.Close()
+			continue
+		}
+		if len(searchResult.Entries) > 1 {
+			conn.Close()
+			return err
```
hsluoyz, line 402: err is nil

go.mod
```diff
@@ -63,6 +63,7 @@ require (
 	golang.org/x/crypto v0.21.0
 	golang.org/x/net v0.21.0
 	golang.org/x/oauth2 v0.17.0
+	golang.org/x/text v0.14.0
```
hsluoyz, line 66: Why import this?

object/ldap_conn.go
```diff
@@ -371,6 +372,64 @@ func GetExistUuids(owner string, uuids []string) ([]string, error) {
...
+		}
+
+		userDN := searchResult.Entries[0].DN
```
hsluoyz, line 405: userDn

object/ldap_conn.go
```diff
@@ -371,6 +372,64 @@ func GetExistUuids(owner string, uuids []string) ([]string, error) {
...
+
+		var pwdEncoded string
+		passwordModify := goldap.NewModifyRequest(userDN, nil)
```
hsluoyz, line 408: passwordModify naming is strange, should be a noun

hsluoyz (changes requested):

go.mod
Xing-Fax: This is to encode the password as UTF-16, because in Microsoft AD domains, passwords are required to be encoded in UTF-16. If you design your own function to implement the encoding, it may not be as stable and could lead to unexpected errors.

controllers/user.go
```diff
@@ -475,6 +475,12 @@ func (c *ApiController) SetPassword() {
 
 	userId := util.GetId(userOwner, userName)
 
+	user, err := object.GetUser(userId)
+	if err != nil {
```
hsluoyz, line 479: user can be nil

go.mod
```diff
@@ -63,6 +63,7 @@ require (
...
 	google.golang.org/api v0.150.0
```
hsluoyz, line 67: Why is go.sum not updated?

hsluoyz (changes requested):

go.mod
Xing-Fax: golang.org/x/text v0.14.0 was already in the go.sum file.

controllers/user.go
```diff
@@ -475,6 +475,16 @@ func (c *ApiController) SetPassword() {
...
+		c.ResponseError(err.Error())
+		return
+	}
+	if user == nil {
+		c.ResponseError("User doesn't exist")
```
hsluoyz, line 484: use existing i18n message

hsluoyz (changes requested):

Issue #3396, opened 2024-12-03
`token_type_hint` should be optional in introspect url


hadestructhor:
Hi.
I've been trying to setup OIDC with a custom mail server but I'm having an issue that is related to the introspection url requiring `token_type_hint` in parameters.

As stated in the OIDC RFC found [here]()

I've reproduced this issue using the demo website of casdoor: https://door.casdoor.com/

Step 1 : Login request with curl
```sh
curl --request POST \
  --url https://door.casdoor.com/api/login/oauth/access_token \
  --header 'content-type: application/json' \
  --data '{
    "grant_type": "password",
    "client_id": "541738959670d221d59d",
    "client_secret": "66863369a64a5863827cf949bab70ed560ba24bf",
    "username": "admin",
    "password": "123"
}
'
```

Step 2: Get the access token from the previous request, and send an introspection request: 
```sh
[...]

PR #3397, opened 2024-12-03
fix: make token_type_hint optional

dacongda:
fix: #3396 

**PR HAVEN'T BEEN TESTED!**

### DIFF
--- a/object/token.go
+++ b/object/token.go
@@ -124,6 +124,7 @@ func GetTokenByRefreshToken(refreshToken string) (*Token, error) {
 func GetTokenByTokenValue(tokenValue, tokenTypeHint string) (*Token, error) {
 	switch tokenTypeHint {
 	case "access_token":
+	case "access-token":
 		token, err := GetTokenByAccessToken(tokenValue)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


object/token.go
```diff
@@ -124,6 +124,7 @@ func GetTokenByRefreshToken(refreshToken string) (*Token, error) {
 func GetTokenByTokenValue(tokenValue, tokenTypeHint string) (*Token, error) {
 	switch tokenTypeHint {
 	case "access_token":
+	case "access-token":
 		token, err := GetTokenByAccessToken(tokenValue)
```
hadestructhor, line 128: I am still installing the packages and trying to run the server mode of the casdoor app, but in the meantime, I checked the RFC for Oauth 2 and the values before for tokenTypeHint were correct. "access_token" and "refresh_token" are the values listed here:
https://datatracker.ietf.org/doc/html/rfc7009#section-2.1

hsluoyz:
@hadestructhor test this PR against your env, give feedback

object/token.go
dacongda: I know, but in casdoor the token_type is set to `-` instead of `_`, I will try to fix it in later PR, I must ensure the compatibility.

hadestructhor:
Just tested the PR, it works now when I run it locally, the token introspection works without the `token_type_hint` attribute. Tested both with the access and refresh token. The only remark I have is the same one as above, with the value of the token_type_hint being access-token instead of the defined one in the RFC access_token. And same with the refresh token.

Issue #3398, opened 2024-12-04
introspect `token_type` should be 'access_token' not 'access-token'


hadestructhor:
I thought it would be an issue from the merge of my [previous issue here](https://github.com/casdoor/casdoor/issues/3396) and in the pull request for it found [here](https://github.com/casdoor/casdoor/pull/3397) it is the case.

Token_type_hint and the returned response should contain either 'access_token' or 'refresh_token'. The newly changed values 'access-token' and 'refresh-token' are invalid and not what is expected by some systems.

At least now the introspect request goes through from my mail server that I setup to use OAUTH2 with casdoor. It's just giving me this error:

```
oauth2 failed: Introspection failed: Expected Bearer token, got 'access-token'
```

Would it be possible to make it compliant with Oauth2's expected return types as I mentioned in the previous issue ?

hsluoyz:
@dacongda

PR #3399, opened 2024-12-05
feat: fix issue that introspectionResponse uses `Bearer` instead of raw tokenType

dacongda:
fix: #3398 

my fault, according to [RFC7662 SECTION 2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2) token type in introspectionResponse should be `Bearer` or `MAC` which defined in [RFC6749 SECTION 7.1](https://datatracker.ietf.org/doc/html/rfc6749#section-7.1)

@hadestructhor plz help me to test

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -402,6 +402,7 @@ func (c *ApiController) IntrospectToken() {
 			return
 		}
 	}
+	introspectionResponse.TokenType = token.TokenType
 
 	c.Data["json"] = introspectionResponse
 	c.ServeJSON()


hadestructhor:
Hi @dacongda, you are right, the return 'token_type' should be 'Bearer' or 'MAC', I just tried it out and it returns 'Bearer' for the access token I'm passing. 
The only thing I'm noticing is the 'token_type_hint' in the request not being either 'access_token' or 'refresh_token' as mentioned in [this RFC](https://datatracker.ietf.org/doc/html/rfc7009) in the [section-2.1](https://datatracker.ietf.org/doc/html/rfc7009#section-2.1):

```
token_type_hint  OPTIONAL.  A hint about the type of the token
           submitted for revocation.  Clients MAY pass this parameter in
           order to help the authorization server to optimize the token
           lookup.  If the server is unable to locate the token using
           the given hint, it MUST extend its search across all of its
           supported token types.  An authorization server MAY ignore
           this parameter, particularly if it is able to detect the
           token type automatically.  This specification defines two
           such values:

           * access_token: An access token as defined in [RFC6749],
             Section 1.4

           * refresh_token: A refresh token as defined in [RFC6749],
             Section 1.5

```

Otherwise, this pr seems to fix my issue with the bearer type

dacongda:
> Hi @dacongda, you are right, the return 'token_type' should be 'Bearer' or 'MAC', I just tried it out and it returns 'Bearer' for the access token I'm passing. The only thing I'm noticing is the 'token_type_hint' in the request not being either 'access_token' or 'refresh_token' as mentioned in [this RFC](https://datatracker.ietf.org/doc/html/rfc7009) in the [section-2.1](https://datatracker.ietf.org/doc/html/rfc7009#section-2.1):
> 
> ```
> token_type_hint  OPTIONAL.  A hint about the type of the token
>            submitted for revocation.  Clients MAY pass this parameter in
>            order to help the authorization server to optimize the token
>            lookup.  If the server is unable to locate the token using
>            the given hint, it MUST extend its search across all of its
>            supported token types.  An authorization server MAY ignore
>            this parameter, particularly if it is able to detect the
>            token type automatically.  This specification defines two
>            such values:
> 
>            * access_token: An access token as defined in [RFC6749],
>              Section 1.4
> 
>            * refresh_token: A refresh token as defined in [RFC6749],
>              Section 1.5
> ```
> 
> Otherwise, this pr seems to fix my issue with the bearer type

thanks for reply

Issue #3411, opened 2024-12-09
xml parsed failed when I use cas auth


nanoguo:
My usages：
1st. create an application for some app(in my case, it's jumpserver), then the auth is work well.
2nd. bind my lark account in casdoor.
3rd. when I open jumpserver login page, it will crash.

so, I debug the error, found that cause, xml parses cas response failed:

```text
File "/opt/py3/lib/python3.11/site-packages/cas.py", line 259, in parse_response_xml
    tree = ElementTree.fromstring(response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/xml/etree/ElementTree.py", line 1350, in XML
    parser.feed(text)
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 1, column 1662
```

beacuse there're some `&` in  the avatar url in the cas response, `&` is not leagal character for the xml.

I encode the avatar url and then update the database record, it works.

then I read this project code, found it in [this line](https://github.com/casdoor/casdoor/blob/master/idp/lark.go#L219), maybe we should encode it before?

I try to do this, it works:
```go
	var encodedAvatarUrl strings.Builder
	if err := xml.EscapeText(&encodedAvatarUrl, []byte(larkUserInfo.Data.AvatarUrl)); err != nil {
		return nil, err
	}
	userInfo := UserInfo{
		Id:          larkUserInfo.Data.OpenId,
		DisplayName: larkUserInfo.Data.Name,
		Username:    larkUserInfo.Data.UserId,
		Email:       larkUserInfo.Data.Email,
		AvatarUrl:   encodedAvatarUrl.String(),
		Phone:       phoneNumber,
		CountryCode: countryCode,
	}
```

so, is  this case should be fix? maybe other idp's avatar url has same problem for cas auth.

hsluoyz:
@nanoguo https://github.com/casdoor/casdoor/blob/master/idp/lark.go#L219 is Casdoor's internal UserInfo struct. It should not do the escape job. The escaping should be done in the XML generating part.

hsluoyz:
@dacongda

PR #3434, opened 2024-12-16
feat: escape the avatal URL in CAS response

nanoguo:
Fix: https://github.com/casdoor/casdoor/issues/3411

Is it ok if I only escape the avatar url?


hsluoyz:
@dacongda plz review

object/token_cas.go
```diff
@@ -225,6 +235,11 @@ func GenerateCasToken(userId string, service string) (string, error) {
 		}
 
 		if value != "" {
+			if k == "avatar" {
```
dacongda, line 238: all value will be string, maybe escape all value could be a good idea

nanoguo: It's ok to escape all value. if other values have special characters such as '&', they can also make crash.

hsluoyz:
@dacongda

object/token_cas.go
```diff
@@ -225,6 +235,9 @@ func GenerateCasToken(userId string, service string) (string, error) {
 		}
 
 		if value != "" {
+			if escapedValue, err := escapeXMLText(value); err == nil {
```
hsluoyz, line 238: What error will it generate? We usually return error if error found

hsluoyz (changes requested):

object/token_cas.go
nanoguo: The error only occur in io.Write function, this escapeXMLText function is try to escape a original string to xml safe string, if it failed, it did nothing, and the response will use the original string like before. should I catch the error and break loop?

hsluoyz: @nanoguo if that error is unpected and critical, we should handle it, not ignore it

nanoguo: ok

hsluoyz:
@dacongda

dacongda (approved):

hsluoyz (approved):

PR #3436, opened 2024-12-16
fix: allow API users to specify default groups for oauth users

mdilshan:
## Summary
When a API user (client system) wants to create a user in Casdoor, we can simply use [this](https://door.casdoor.com/swagger/#/User%20API/ApiController.AddUser]) API to create the user with necessary groups.

However if we allow 3rd party OAuth login for API users we don't have a way to set the default groups for these user.

In this pull request I propose to allow any API user to set the default groups when using OAuth login.

## Example Usage

```
  public async LoginWithOAuth(provider: string, oauthCode: string, redirectUri: string) {
    const res = await this.openIDClientService.startAuthFlow();
    const fetchUrl = res.ajaxUrl;

    const response = await fetch(fetchUrl, {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      body: JSON.stringify({
        type: "code",
        application: this.appConfig.casdoor.appName,
        provider,
        code: oauthCode,
        samlRequest: null,
        state: this.appConfig.casdoor.appName,
        redirectUri,
        method: "signup",
        groups: this.dynamicConfig.defaultGroups, // <-- Here I can add default groups for OAuth users
      })
    });
    return response
  }
```

### DIFF
--- a/form/auth.go
+++ b/form/auth.go
@@ -20,22 +20,23 @@ type AuthForm struct {
 	Type         string `json:"type"`
 	SigninMethod string `json:"signinMethod"`
 
-	Organization   string `json:"organization"`
-	Username       string `json:"username"`
-	Password       string `json:"password"`
-	Name           string `json:"name"`
-	FirstName      string `json:"firstName"`
-	LastName       string `json:"lastName"`
-	Gender         string `json:"gender"`
-	Bio            string `json:"bio"`
-	Tag            string `json:"tag"`
-	Education      string `json:"education"`
-	Email          string `json:"email"`
-	Phone          string `json:"phone"`
-	Affiliation    string `json:"affiliation"`
-	IdCard         string `json:"idCard"`
-	Region         string `json:"region"`
-	InvitationCode string `json:"invitationCode"`
+	Organization   string   `json:"organization"`
+	Username       string   `json:"username"`
+	Password       string   `json:"password"`
+	Name           string   `json:"name"`
+	FirstName      string   `json:"firstName"`
+	LastName       string   `json:"lastName"`
+	Gender         string   `json:"gender"`
+	Bio            string   `json:"bio"`
+	Tag            string   `json:"tag"`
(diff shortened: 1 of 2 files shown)


hsluoyz:
@dacongda plz review

form/auth.go
```diff
@@ -20,22 +20,23 @@ type AuthForm struct {
...
+	Bio            string   `json:"bio"`
+	Tag            string   `json:"tag"`
+	Education      string   `json:"education"`
+	Email          string   `json:"email"`
+	Phone          string   `json:"phone"`
+	Affiliation    string   `json:"affiliation"`
+	IdCard         string   `json:"idCard"`
+	Region         string   `json:"region"`
+	InvitationCode string   `json:"invitationCode"`
+	Groups         []string `json:"groups"`
```
dacongda, line 39: I think we should check whether the group is exist. and according the code, it seems that user can specify group by themselves, will this cause some potential priviledge issue?

mdilshan: I could add a check for the existence of the groups but you are right it could cause some privilege issue as I think now. 

Feel like there should be a field in application edit page to select default groups for new users.

mdilshan: For now I ended up doing something like this on my side,

```
   ...
    let { tokens, claims } = await this.openIDClientService.callback(
      `${this.appConfig.openid.redirectUri}?code=${code}&state=${res.sessionId}`
    );

    if ((claims as any).groups.length === 0) { //<-- update the user if there's no group
      const user = await this.casdoorService.getUserByUserId(claims.id);
      await this.casdoorService.updateUser({
        ...user,
        groups: this.dynamicConfig.defaultGroups,
      });

      // Refresh the tokenSet
      tokens = await this.openIDClientService.getAccessTokenFromRefreshToken(tokens.refresh_token);
    }
    ...
```

hsluoyz:
Closed that this PR has security hole.

Issue #3449, opened 2024-12-19
how to make  casdoor as an idp provider


wayneliu0019:
hi all:
 
    I've already configured casdoor as my two app's IAM, and work well.
 ```
app1 ---- casdoor1
app2 ---- casdoor2
``` 
  My question is: `how to add casdoor as another casdoor1 instance as oauth provider to casdoor2?`

 the doc shows it already support this:
<img width="1300" alt="image" src="https://github.com/user-attachments/assets/4ee7040b-8fb8-4bba-a1dc-3c54484362ff" />

 Does anybody be able to figure out how to do necessary configure? 

thanks!

hsluoyz:
We need to add a "Casdoor" docs page in: https://casdoor.org/docs/category/oauth/

wayneliu0019:
on casdoor2's provider page, add below item: 

![image](https://github.com/user-attachments/assets/b413a1a5-ea41-4402-8d13-20dde327807a)

I don't think it is enough, there is no cadoor1's auth url information! casdoor2 can not found casdoor1

wayneliu0019:
@hsluoyz  BTW，is there any help doc about my issue?

hsluoyz:
@wayneliu0019 try domain or provider URL, I forgot which one

wayneliu0019:
that's cool!  I've added casdoor1's domain to casdoor2's provider domain and provider URL: 
<img width="1255" alt="image" src="https://github.com/user-attachments/assets/c8ba2559-0aed-4370-8335-69a09103377f" />

but still failed after input casdoor1's user&passwd : 
<img width="1611" alt="image" src="https://github.com/user-attachments/assets/9031b975-600a-4302-b08c-19b6873e19aa" />

it seems there is no protocal for `/api/login/oauth/access_token` call.   it is weird~ the post request's endpoint should have protocal:  https://github.com/casdoor/casdoor/blob/v1.733.0/idp/casdoor.go#L53 

Do I missed something? @hsluoyz

hsluoyz:
@dacongda

wayneliu0019:
BTW， the casdoor version I am using is:

casdoor1  ---- `1.571.0`   （will be used as provider for casdoor2）
casdoor2 ---- `1.733.0`

hsluoyz:
@wayneliu0019 you should always use the latest version (which is v1.776.0 now).

wayneliu0019:
[...]

PR #3456, opened 2024-12-27
fix: cannot use casdoor as idp because casdoor idp use domain field as its host

dacongda:
fix: #3449


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/commit/4d0f73c84e2ec282a01f41f1962cd7f6eb53d1be

Issue #3455, opened 2024-12-26
[feature] dashboard page is slow and consuming 32 GB RAM for 20 Million+ users


hsluoyz:
See: https://door.casdoor.com/

![image](https://github.com/user-attachments/assets/6fecd1c7-c179-4b11-ba12-cd8f8574f6cd)

![image](https://github.com/user-attachments/assets/3eb088c7-1c53-40fe-80fd-53161efe9083)

glenn0n:
Hello.

When we migrate few thousand users - its fine and dashboard renders correctly.

A large number of users (millions +) in itself does not cause a problem. Its only with web-interface and dashboard, which cannot render properly. When we login into web, pod starts utilize RAM and crashes (restarting) when he reaches to 32GB RAM.

glenn0n:
64GB its not enough either.

I think the main reason is in the database queries and the returned results:
```
[...]

PR #3457, opened 2024-12-28
feat: only fetch created_time field to reduce data size in get-dashboard API

dacongda:
Part of: https://github.com/casdoor/casdoor/issues/3455

a simple solution, better solution is on the way

### DIFF
--- a/object/get-dashboard.go
+++ b/object/get-dashboard.go
@@ -36,6 +36,10 @@ type Dashboard struct {
 	EnforcerCounts     []int `json:"enforcerCounts"`
 }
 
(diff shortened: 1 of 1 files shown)


hsluoyz:
@glenn0n can you test this PR?

glenn0n:
@hsluoyz Hello!

Thank you for this solution!
It's better now, because if we go to the dashboard, the RAM consumption increases for a short time, but the pod continues to work.

However, if we refresh the dashboard page, the pod runs out of memory. I guess this is due to sending multiple queries to the DB.

image

Issue #3461, opened 2024-12-29
[feature] add provider.EmailRegex to filter the Email of OAuth provider like Google


hsluoyz:
`provider.EmailRegex` should be an "Allow" regex. if `provider.EmailRegex is not empty and the OAuth user's Email field is not empty, we match the regex. If matched, then allow the access like before. If not, block the OAuth process and return error in the REST API

### Discussed in https://github.com/casdoor/casdoor/discussions/3460

<div type='discussions-op-text'>

<sup>Originally posted by **parasiteoflife** December 29, 2024</sup>
Can we limit access by email domain? Like I enable several providers but after the user login in using the provider I want casdoor to check the email domain and deny access if it's not in the allowed list, we can do this with Authentik with the flow policies, can we do this in Casdoor? How?</div>

PR #3465, opened 2024-12-30
feat: add regex to restrict Email addresses in OAuth provider

dacongda:
Fix: https://github.com/casdoor/casdoor/issues/3461

demo img

image

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -616,6 +617,13 @@ func (c *ApiController) Login() {
 			if err != nil {
 				c.ResponseError(fmt.Sprintf(c.T("auth:Failed to login in: %s"), err.Error()))
 				return
+			}
+
+			if provider.EmailRegex != "" {
+				reg := regexp.MustCompile(provider.EmailRegex)
+				if !reg.MatchString(userInfo.Email) {
+					c.ResponseError(fmt.Sprintf(c.T("auth:email is not in allowed list")))
+				}
 			}
 		}
 
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -620,9 +620,13 @@ func (c *ApiController) Login() {
 			}
 
 			if provider.EmailRegex != "" {
-				reg := regexp.MustCompile(provider.EmailRegex)
+				reg, err := regexp.Compile(provider.EmailRegex)
+				if err != nil {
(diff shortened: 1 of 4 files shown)


object/email.go
```diff
@@ -43,6 +45,11 @@ func SendEmail(provider *Provider, title string, content string, dest string, se
 		fromAddress = provider.ClientId
 	}
 
+	reg := regexp.MustCompile(provider.Metadata)
```
hsluoyz, line 48: Why metadata?

And why add it here in SendEmail?

object/provider.go
```diff
@@ -70,6 +70,7 @@ type Provider struct {
 	IdP                    string `xorm:"mediumtext" json:"idP"`
 	IssuerUrl              string `xorm:"varchar(100)" json:"issuerUrl"`
 	EnableSignAuthnRequest bool   `json:"enableSignAuthnRequest"`
+	EmailRegex             string `xorm:"mediumtext" json:"emailRegex"`
```
hsluoyz, line 73: varchar(200) is enough

web/src/ProviderEditPage.js
```diff
@@ -633,6 +633,20 @@ class ProviderEditPage extends React.Component {
             </React.Fragment>
           )
         }
+        {
+          this.state.provider.category === "OAuth" ? (
+            <Row style={{marginTop: "20px"}} >
+              <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 22 : 2}>
+                {Setting.getLabel(i18next.t("signup:Regex"), i18next.t("signup:Regex"))} :
```
hsluoyz, line 640: Label should be `Email regex`

controllers/auth.go
```diff
@@ -617,6 +618,13 @@ func (c *ApiController) Login() {
 				c.ResponseError(fmt.Sprintf(c.T("auth:Failed to login in: %s"), err.Error()))
 				return
 			}
+
+			if provider.EmailRegex != "" {
```
hsluoyz, line 622: First, `provider.EmailRegex` should be validated in UpdateProvider() API

Here you should also return error if the regex is invalid instead of panic. We don't use panic any more.

We should already have such code, copy it

controllers/auth.go
```diff
@@ -617,6 +618,13 @@ func (c *ApiController) Login() {
...
+				reg := regexp.MustCompile(provider.EmailRegex)
+				if !reg.MatchString(userInfo.Email) {
+					c.ResponseError(fmt.Sprintf(c.T("auth:email is not in allowed list")))
```
hsluoyz, line 625: Refer to existing i18n string

hsluoyz (changes requested):

hsluoyz (approved):

Issue #3423, opened 2024-12-12
Login logic issue using email verification code when email MFA verification is enabled


Sun-drenched:
When using an email verification code to log in to an account with email MFA verification turned on, you will be asked to verify the email verification code twice in a row (you must wait 60 seconds to get the second verification code). I think there is a logical problem here, and it should only be required to verify once.

hsluoyz:
@Sun-drenched plz provide a demo video to reproduce it

Sun-drenched:
> @Sun-drenched plz provide a demo video to reproduce it

https://github.com/user-attachments/assets/ef203391-e1b4-409b-8dd5-2ca1934fc621

hsluoyz:
@Sun-drenched the code login method should skip the MFA like: https://door.casdoor.com/login

![image](https://github.com/user-attachments/assets/4b42553f-79df-4131-91cc-6605b1ae294d)

Sun-drenched:
> @Sun-drenched the code login method should skip the MFA like: https://door.casdoor.com/login
> 
[...]

PR #3475, opened 2025-01-05
fix: Skip redundant MFA verification for same type during email/SMS login (#3423)

Zerek-Cheng:
Fix: https://github.com/casdoor/casdoor/issues/3423

This PR fixes an issue where users are required to verify twice when logging in with email or SMS verification and the MFA method is of the same type (e.g., email MFA for email login).
﻿  
* The login process now skips MFA verification if the MFA method matches the login method (e.g., email login skips email MFA, SMS login skips SMS MFA).
* If the MFA method is different from the login method (e.g., email login with SMS MFA), the MFA step will still be required.
  
Why this fix is needed:
﻿
Requiring users to verify twice with the same method adds unnecessary friction to the login process without improving security. This change simplifies the user experience while maintaining security for cases where the MFA method is different.
﻿
This should address the issue described in #3423.

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -525,10 +525,17 @@ func (c *ApiController) Login() {
 				return
 			}
 
-			if user.IsMfaEnabled() && (!useVerificationCode || verificationCodeType != user.PreferredMfaType) {
+			if user.IsMfaEnabled() {
+				var skipMfaType []string
+				if useVerificationCode {
+					skipMfaType = []string{verificationCodeType}
+				}
 				c.setMfaUserSession(user.GetId())
-				c.ResponseOk(object.NextMfa, user.GetPreferredMfaProps(true))
-				return
+				preferredMfaProps := user.GetPreferredMfaProps(true, skipMfaType...)
+				if preferredMfaProps != nil {
+					c.ResponseOk(object.NextMfa, preferredMfaProps)
+					return
+				}
 			}
 
 			resp = c.HandleLoggedIn(application, user, &authForm)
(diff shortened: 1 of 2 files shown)


Zerek-Cheng:
test video:

https://github.com/user-attachments/assets/af8cfebc-a00a-4cdd-81e2-47eaf5f3c852

 @hsluoyz check pls.

hsluoyz:
@dacongda plz review

controllers/auth.go
```diff
@@ -522,7 +525,7 @@ func (c *ApiController) Login() {
 				return
 			}
 
-			if user.IsMfaEnabled() {
+			if user.IsMfaEnabled() && (!useVerificationCode || verificationCodeType != user.PreferredMfaType) {
```
dacongda, line 528: it seems that useVerificationCode will always be true in this branch, when will it be false?

Zerek-Cheng: when u use username and password to login un
it will be false(like the video)

dacongda: ok

dacongda (approved):

hsluoyz:
@Zerek-Cheng @dacongda 

the user should be required to use the other MFA method if exists. Not just skip. For example, if user uses SMS verification code to login and there are three MFA methods: SMS, Email and TOTP App. Then the user will be required to use Email or TOTP App as MFA (SMS is removed because the user uses SMS to login).

It should not be the case that the user is allowed to choose SMS as MFA again and then skip the second check.

Zerek-Cheng:
> @Zerek-Cheng @dacongda
>
> the user should be required to use the other MFA method if exists. Not just skip. For example, if user uses SMS verification code to login and there are three MFA methods: SMS, Email and TOTP App. Then the user will be required to use Email or TOTP App as MFA (SMS is removed because the user uses SMS to login).
>
> It should not be the case that the user is allowed to choose SMS as MFA again and then skip the second check.

You have the right idea. I'll think about it.

Zerek-Cheng:
@dacongda  @hsluoyz  review plz

hsluoyz:
@dacongda the code is too complicated. Can you give a simpler PR?

PR #3476, opened 2025-01-06
fix: Dockerfile's stage name should be lowercase (#1)

cutsin:
eg: https://docs.docker.com/reference/build-checks/stage-name-casing/#:~:text=To%20help%20distinguish%20Dockerfile%20instruction,Dockerfile%20to%20be%20all%20lowercase.


hsluoyz:
@dacongda plz review

dacongda:
we should also change github action file

hsluoyz:
@cutsin we don't want to introduce new bugs only for fixing such small issues (like small-issue naming convention), base on the review of the reviewer

cutsin:
> we should also change github action file

Oops, I forgot it...

PR #3477, opened 2025-01-06
fix: some dependencies were missing during build stage

cutsin:


cutsin:
As you wish. But you can reproduce it with an empty node_modules.
I just got a problem, I solved it.

PR #3516, opened 2025-01-20
feat: LDAP user can reset password with old password and new password

dacongda:
feat: support user reset password with old password and new password

https://github.com/user-attachments/assets/81f4e1b4-2cbb-46b4-809b-ed12f0346448

now user can reset password with oldPassword, and reset password by user will use default encrypt method set by ldap server


dacongda:
need to test with MicrosoftAD

Issue #3524, opened 2025-01-22
delete group fail


lqa0408:
When I delete a user and then go to delete a group, the group shows that there are no associated users, but when deleting the group, it prompts that there are still users in the group

clarin-ebtio800090:
Hello, we tried to solve the issue.

This is what we did:

Modify the DeleteGroup function to properly check for associated users, including those that might have been deleted. Update the user deletion process to remove group associations.

You can review changes in this commit: https://github.com/clarin-ebtio800090/casdoor-casdoor-3524/commit/53f95462f4a5a19fcaec66787edebb6425573532.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

dacongda:
will be fix later, you can enter table `casbin_user_rule` to check and delete user that has been deleted and restart the app

PR #3525, opened 2025-01-22
feat: delete groups for user while deleting user

dacongda:
fix: #3524

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -965,6 +965,14 @@ func DeleteUser(user *User) (bool, error) {
 		return false, err
 	}
 
+	ok, err := userEnforcer.DeleteGroupsForUser(user.GetId())
+	if err != nil {
+		return false, err
+	}
+	if !ok {
+		return false, nil
+	}
+
 	organization, err := GetOrganizationByUser(user)
 	if err != nil {
 		return false, err


object/user.go
```diff
@@ -965,6 +965,14 @@ func DeleteUser(user *User) (bool, error) {
 		return false, err
 	}
 
+	ok, err := userEnforcer.DeleteGroupsForUser(user.GetId())
+	if err != nil {
+		return false, err
+	}
+	if !ok {
```
Sophrinix, line 972: this is breaking the ability to delete a user even if an admin in built-in in this and all subsequent builds up to the latest. If not this code. this behavior I am describing begins on this specific release. 1.813 is unaffected, but every version since this 1.814 all the way to 1.820 has this issue.

hsluoyz: @dacongda

dacongda: How this code breaks the delete user function

dacongda: > this is breaking the ability to delete a user even if an admin in built-in in this and all subsequent builds up to the latest. If not this code. this behavior I am describing begins on this specific release. 1.813 is unaffected, but every version since this 1.814 all the way to 1.820 has this issue.

solved in https://github.com/casdoor/casdoor/pull/3544 can you help me test?

PR #3530, opened 2025-01-25
[WIP] Switch to custom Discord implementation

IngwiePhoenix:
This is my initial attempt at switching Casdoor to use a custom Discord Oauth2 implementation. Reason being that I would like to use the user's `ExtraData` field to filter user access by the Discord Servers (`guilds`) they are being joined to.

The code is implemented as a copy and edit of `idp/gitee.go`.

What I am currently stuck at is actually informing the rest of the system about the change. I saw `controller/auth.go` and `object/provider.go`...mind giving me a pointer?

I introduced the `discordgo` library as a dependency, since it has predefined data structures and functions - no need to reinvent the wheel ;). This might also come in handy down the line (i.e.: Casdoor living as a bot on a server to automate access grants and restricts based off of server roles).

The implementation is not complete yet (the current `ExtraData` layout is very basic...for now).

Thanks a lot and kind regards,
Ingwie!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/aws/aws-sdk-go v1.45.5
 	github.com/beego/beego v1.12.12
 	github.com/beevik/etree v1.1.0
+	github.com/bwmarrin/discordgo v0.27.1 // indirect
 	github.com/casbin/casbin/v2 v2.77.2
 	github.com/casdoor/go-sms-sender v0.25.0
 	github.com/casdoor/gomail/v2 v2.1.0
--- a/idp/discord.go
+++ b/idp/discord.go
@@ -1,0 +1,169 @@
+// Copyright 2021 The Casdoor Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package idp
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strconv"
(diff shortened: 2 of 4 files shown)


dacongda:
do you want to change user's info while login with casdoor? you should check `object.SetUserOAuthProperties`

IngwiePhoenix:
@hsluoyz Goth is great, but it's greatness stems from being "generic". What I aim to do is to implement user grouping based on the servers that the user is in - because I have been trying to find an IdP that does that since forever and have tried many, many solutions for that too... I ended up trying my luck with Casdoor since it's management UI is very straight forward and it technically has all the features. That is to say, all but grouping users by their discord servers, since no other IdP actually goes and gathers them, at all.

That's why I am implementing it here - step by step, I want to get this feature to solve the problem that I have had.

To give a little bit more detail: I have a k3s cluster at home and a VPS acting as the outbound ingress that communicates with the cluster via a Headscale VPN (`k3s <- vpn -> vps`). Some of the services I run, like Pinchflat, do not have a login or even user management - and others could profit from being connected via OAuth2 like Grafana, so there is no need to save a bunch of passwords - and further, since my friends and I are on a small discord server, it'd be great if they could just use their Discord login instead and become authorized by being in that server. Vice versa, if their account were to be compromised, I'd just remove them from the server, and thus also deauthorize them in the process.

Basically, I want to reuse what we already have (Discord accounts and a server) as a means to handle authentication and authorization. And a generic provider like Goth unfortunately does not support that I am afraid... Hence, why I am working on this - well, for now I am trying to understand the structure better to "advertise" the "new" provider to the rest of the system. Once done, I will move on to see if I can implement auto-grouping based off of the user's `extraData`. :)

IngwiePhoenix:
> do you want to change user's info while login with casdoor? you should check `object.SetUserOAuthProperties`

More like at sign-up and possibly refresh every once in a while (a-la "profile sync").

hsluoyz:
@IngwiePhoenix Does your new Discord provider change the original provider's behavior? If yes, you should name a new provider, not to replace the old one

IngwiePhoenix:
> @IngwiePhoenix Does your new Discord provider change the original provider's behavior? If yes, you should name a new provider, not to replace the old one

It requests an additional scope (`guilds`) that the original one does not. Other than that, and the additional information stored in the user profile, there is not a big difference.

hsluoyz:
@IngwiePhoenix you should make PR to: https://github.com/markbates/goth, we don't want to diverge with goth official community too far away.

Issue #3342, opened 2024-11-12
Casbin Permissions Not Working When Auto-login is Enabled


Xing-Fax:
Description: When auto-login is enabled in an application, initiating an authorized login from the application login page can bypass the Casbin deny policy, allowing access directly without enforcing the intended restrictions.

Steps to Reproduce:

Enable auto-login in the application, such as in the Nightingale application.
In Casbin, create a new permission and set Nightingale to "deny."
Attempt to initiate an authorized login from the application's login page.
You'll find that the account is logged in automatically without any error message, indicating an issue.
Disable auto-login in the application.
Try logging in again.
This time, an "Unauthorized Access" error appears, which is the expected behavior.

![image](https://github.com/user-attachments/assets/ef15065d-4d97-48c7-ac81-93b7b33e1de0)
![image](https://github.com/user-attachments/assets/8f6b1a47-ed53-4bc1-954f-5cc3204fd2f8)

phoenixsqf:
Same issue, need this PR.

PR #3537, opened 2025-01-28
fix: Casbin Permissions Not Working When Auto-login is Enabled

dacongda:
fix: #3342


dacongda:
@phoenixsqf can you help me test?

phoenixsqf:
> @phoenixsqf can you help me test?

I will test it

Issue #3536, opened 2025-01-27
Accounts created by OAuth provider have no password


fischermario:
I have discovered that accounts created by Casdoor via an OAuth provider have an empty password. When using the "password" grant it is possible to obtain an access token for these users.

Steps to reproduce:

1. Allow the "password" grant for an application:
![Image](https://github.com/user-attachments/assets/f7c797b3-6c8b-4860-a716-2614d723cb91)
2. Add an OAuth provider (e.g. AzureAD):
![Image](https://github.com/user-attachments/assets/b3963ed5-b9c2-4700-b1d2-771761a4edd6)
3. Now log in to an app with an user that does not exist yet in Casdoor using the OAuth provider. A new user will be created in Casdoor.
4. Next obtain an access token with the client_id of the app and the users' email address while leaving the password field empty:
![Image](https://github.com/user-attachments/assets/60ecd4f6-ba22-4670-a15d-f3f749446887)

Please fix this by checking the password field. If the password field is empty the login should be denied under all circumstances.

hsluoyz:
@fischermario can you make a PR to fix it?

fischermario:
@hsluoyz I have created PR #3538 with an appropriate fix. Can you please review it?

hsluoyz:
Fixed by: https://github.com/casdoor/casdoor/commit/bad21fb6bb6bcca3e48b9721339c4b2fd3178305

PR #3538, opened 2025-01-28
fix: empty check for username and password field with password grant

fischermario:
fix: #3536

### DIFF
--- a/object/token_oauth.go
+++ b/object/token_oauth.go
@@ -499,6 +499,13 @@ func GetAuthorizationCodeToken(application *Application, clientSecret string, co
 // GetPasswordToken
 // Resource Owner Password Credentials flow
 func GetPasswordToken(application *Application, username string, password string, scope string, host string) (*Token, *TokenError, error) {
+	if username == "" || password == "" {
+		return &TokenError{
+			Error:            InvalidGrant,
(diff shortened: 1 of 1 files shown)


hsluoyz:
@fischermario the code already checked username and password, why check null again?

image

fischermario:
I have described the problem in #3536. If the user was created by a login provider the password of the user will be empty. The check in line 523 
`err = CheckPassword(user, password, "en")`
will pass because the password in the database is empty. This is why I check the input data.

Do you want me to modify the CheckPassword method instead?

hsluoyz:
@fischermario CheckPassword should not let empty password pass, you can see the code, can you point out where CheckPassword lets empty password pass?

image

fischermario:
@hsluoyz The match between `password` and `user.Password` inside CheckPassword happens at line 256 (given that the org as no master password):
`credManager.IsPasswordCorrect(password, user.Password, user.PasswordSalt, organization.PasswordSalt)`
If the password in the database is empty (therefore `user.Password == ""`) the check will pass. There is no explicit check in the code that prevents logins with an empty password.

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/commit/bad21fb6bb6bcca3e48b9721339c4b2fd3178305

Issue #3533, opened 2025-01-27
[feature] define what Casdoor pages an org admin can see


hsluoyz:
For now, an org admin (except built-in org) can see a lot of pages in Casdoor, like most Navbar items, and their org's items. But sometimes we only want them to see some of these Navbar items (including frontend UI elements and also the backend APIs). Even let org admins just be like org users inside Casdoor (except their isAdmin is true)

We need a way to configure this, like 

- [x] Applications
- [ ] Users
- [ ] Profile
- [ ] xxx 

![Image](https://github.com/user-attachments/assets/98e4b5a5-b093-4545-bf97-2e672885ca93)

PR #3539, opened 2025-01-28
feat: can define what Casdoor pages an org admin can see via Organization.NavItems

dacongda:
Fix: https://github.com/casdoor/casdoor/issues/3533

demo:

https://github.com/user-attachments/assets/f6e59b26-0030-49fa-b05c-8266d0a746ff


hsluoyz:
@dacongda it should be the global admin to control what navbar items an org admin can see. The org admin cannot decide by himself. Different orgs can have different configs (admin for org1 can see item A, B, C, admin for org2 can see item A, D, E, etc.)

dacongda:
> @dacongda it should be the global admin to control what navbar items an org admin can see. The org admin cannot decide by himself. Different orgs can have different configs (admin for org1 can see item A, B, C, admin for org2 can see item A, D, E, etc.)
> 
> Global admin himself can also decide his own navbar items. But this is mainly for web UI visual experience instead of permission. Because he can always turn on items by himself

solved in commit 2554482

web/src/OrganizationEditPage.js
```diff
@@ -522,6 +603,23 @@ class OrganizationEditPage extends React.Component {
             }} />
           </Col>
         </Row>
+        <Row style={{marginTop: "20px"}} >
```
hsluoyz, line 606: Define this new Tree UI widget in a new JS file, decouple with existing code

hsluoyz (changes requested):

hsluoyz (approved):

PR #3550, opened 2025-02-04
Handle individual User Properties for custom claims

JustinJudd:
Using a "." notation in the application custom token fields list to enable surfacing them to the top level, without needing to expose all of the Properties for a user.

Example:
If the user has a property named `preferred_username`, the application would add `Properties.preferred_username` to the token fields list.

This also addresses the logstanding comment for
`// FIXME: A workaround for custom claim by reusing `tag` in user info`


hsluoyz:
@JustinJudd don't understand what you did

PR #3555, opened 2025-02-06
Fix reset password flow for shared application

maknahar:
fix: Ensure to add org in application name if application is shared


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/3556

PR #3556, opened 2025-02-06
fix: Fix reset password flow for shared application

maknahar:
fix: Ensure to add org in application name if application is shared


hsluoyz:
@dacongda plz review

dacongda (approved):

PR #3571, opened 2025-02-11
Enable using User Properties as custom claims

JustinJudd:
Enable adding custom claims that can match fields that are needed by downstream applications. Using a "." notation in the application custom token fields list to enable surfacing them to the top level, without needing to expose all of the Properties for a user.

Example:
If the user has a property named `preferred_username`, the application would add `Properties.preferred_username` to the token fields list.

Application Setting:
image

User Setting:
image

Generated JWT (notice the `preferred_username` field, which is not a field of the user in http://casdoor.org/docs/user/overview/):

```
{
  "alg": "RS256",
  "kid": "cert-built-in",
  "typ": "JWT"
}.{
  "aud": [
    "0ac0f951d0b93ff0cbab"
  ],
  "exp": 1739829942,
  "iat": 1739225142,
  "iss": "http://localhost:8000",
  "jti": "admin/b5544cbd-a3c0-4e48-88b4-04a6493ef782",
  "nbf": 1739225142,
  "nonce": "",
  "owner": "organization_queimk",
  "preferred_username": "my_username",
  "scope": "profile",
  "sub": "7a3d19df-d0f8-4d86-a376-688c82d48c42",
  "tag": "staff",
  "title": "Awesome Guy",
  "tokenType": "access-token"
}
```

Note: This also will enable not needing the workaround used in the Minio documentation (https://casdoor.org/docs/integration/go/minio/#step-2-configure-casdoor-application Step 4).
Instead of using a tag for the claim name, any field name could be chosen and set in the properties.

### DIFF
--- a/object/token_jwt.go
+++ b/object/token_jwt.go
@@ -303,10 +303,25 @@ func getClaimsCustom(claims Claims, tokenField []string) jwt.MapClaims {
 	res["scope"] = claims.Scope
 
 	for _, field := range tokenField {
-		userField := userValue.FieldByName(field)
-		if userField.IsValid() {
-			newfield := util.SnakeToCamel(util.CamelToSnakeCase(field))
-			res[newfield] = userField.Interface()
+		if strings.HasPrefix(field, "Properties") {
+			parts := strings.Split(field, ".")
+			if len(parts) == 2 {
+				base, fieldName := parts[0], parts[1]
+				mField := userValue.FieldByName(base)
+				if mField.IsValid() {
+					finalField := mField.MapIndex(reflect.ValueOf(fieldName))
+					if finalField.IsValid() {
+						newfield := util.SnakeToCamel(util.CamelToSnakeCase(fieldName))
+						res[newfield] = finalField.Interface()
+					}
+				}
+			}
+		} else {
+			userField := userValue.FieldByName(field)
+			if userField.IsValid() {
+				newfield := util.SnakeToCamel(util.CamelToSnakeCase(field))
+				res[newfield] = userField.Interface()
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


object/token_jwt.go
```diff
@@ -303,10 +303,24 @@ func getClaimsCustom(claims Claims, tokenField []string) jwt.MapClaims {
 	res["scope"] = claims.Scope
 
 	for _, field := range tokenField {
-		userField := userValue.FieldByName(field)
-		if userField.IsValid() {
-			newfield := util.SnakeToCamel(util.CamelToSnakeCase(field))
-			res[newfield] = userField.Interface()
+		if strings.HasPrefix(field, "Properties") {
```
hsluoyz, line 306: The code is hard to read. Maybe you should use ChatGPT or Claude to write the code for you

hsluoyz (changes requested):

object/token_jwt.go
JustinJudd: Reorganized the code and added some comments.

hsluoyz:
@JustinJudd you should create an issue to discuss first

AtCliffUnderline:
I want to upvote this feature, should be very useful for me

hsluoyz:
@JustinJudd the problem is that the code quality is below bar. It doesn't even handle out of index error. So many bugs at the first glance. Need a more experienced person to propose a better PR

JustinJudd:
The only indexing is on `parts`, and before any indexing is done there is a length check to ensure that it has the expected number of fields - `if len(parts) != 2`

JustinJudd:
Also created a feature request to capture this - https://github.com/casdoor/casdoor/issues/3617

avivbintangaringga:
I desperately needed this feature

PR #3585, opened 2025-02-14
feat: Synology compatibility improvements

tianyanli:
Synology compatibility minimization improvement, adding configuration switch to ensure original functions are not affected


hsluoyz:
@tianyanli 

1. You should create issue first
2. What compat does this PR improve, is it for supporting old Synology version?
3. The code becomes dirty. We don't want the main LDAP code to become dirty work just for supporting old version and non-standard Synology.

Issue #3594, opened 2025-02-18
feat: Support multi RPOrigins & follow webauthn lib


nb150301:
webauthn already Deprecated RPOrigin: https://github.com/go-webauthn/webauthn/blob/a328b580940fa709b0e85fba422587dfaf149690/webauthn/main.go#L21-L23

We should use RPOrigins for multi Origins supported in `GetWebAuthnObject`:
https://github.com/casdoor/casdoor/blob/d61f9a1856572138afae63091ee31051afb98999/object/user_webauthn.go#L39-L42

hsluoyz:
@nb150301 can you make a PR?

PR #3610, opened 2025-02-22
feat: Webauthn support multi origins

nb150301:
Fix: https://github.com/casdoor/casdoor/issues/3594

This issue was created because I have many applications that call `webauthn` api in `casdoor`. I discovered that I can use the application's `homepage_url` instead of configuring it from the env

### DIFF
--- a/controllers/webauthn.go
+++ b/controllers/webauthn.go
@@ -33,15 +33,15 @@ import (
 // @Success 200 {object} protocol.CredentialCreation The CredentialCreationOptions object
 // @router /webauthn/signup/begin [get]
 func (c *ApiController) WebAuthnSignupBegin() {
-	webauthnObj, err := object.GetWebAuthnObject(c.Ctx.Request.Host)
-	if err != nil {
-		c.ResponseError(err.Error())
-		return
-	}
-
 	user := c.getCurrentUser()
 	if user == nil {
 		c.ResponseError(c.T("general:Please login first"))
+		return
+	}
+
+	webauthnObj, err := object.GetWebAuthnObjectByUser(user)
+	if err != nil {
+		c.ResponseError(err.Error())
 		return
 	}
 
@@ -69,17 +69,18 @@ func (c *ApiController) WebAuthnSignupBegin() {
 // @Success 200 {object} controllers.Response "The Response object"
 // @router /webauthn/signup/finish [post]
 func (c *ApiController) WebAuthnSignupFinish() {
-	webauthnObj, err := object.GetWebAuthnObject(c.Ctx.Request.Host)
-	if err != nil {
-		c.ResponseError(err.Error())
-		return
-	}
-
 	user := c.getCurrentUser()
 	if user == nil {
 		c.ResponseError(c.T("general:Please login first"))
 		return
 	}
+
+	webauthnObj, err := object.GetWebAuthnObjectByUser(user)
+	if err != nil {
+		c.ResponseError(err.Error())
+		return
+	}
+
 	sessionObj := c.GetSession("registration")
 	sessionData, ok := sessionObj.(webauthn.SessionData)
 	if !ok {
@@ -112,12 +113,6 @@ func (c *ApiController) WebAuthnSignupFinish() {
 // @Success 200 {object} protocol.CredentialAssertion The CredentialAssertion object
 // @router /webauthn/signin/begin [get]
 func (c *ApiController) WebAuthnSigninBegin() {
-	webauthnObj, err := object.GetWebAuthnObject(c.Ctx.Request.Host)
-	if err != nil {
-		c.ResponseError(err.Error())
-		return
-	}
-
 	userOwner := c.Input().Get("owner")
 	userName := c.Input().Get("name")
 	user, err := object.GetUserByFields(userOwner, userName)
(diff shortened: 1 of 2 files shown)


hsluoyz:
@dacongda plz review

dacongda (approved):

object/user_webauthn.go
```diff
@@ -20,26 +20,22 @@ import (
...
-	_, originBackend := getOriginFromHost(host)
-
-	localUrl, err := url.Parse(originBackend)
+func GetWebAuthnObjectByApplication(application *Application) (*webauthn.WebAuthn, error) {
+	localUrl, err := url.Parse(application.HomepageUrl)
 	if err != nil {
 		return nil, fmt.Errorf("error when parsing origin:" + err.Error())
 	}
 
+	rpOrigins := []string{application.HomepageUrl}
```
hsluoyz, line 33: Why you only use one element if rpOrigins is multiple? You should not delete the original `originBackend` value because maybe there are a lot of people already are using it. You should consider them. You should not only consider yourself. You should keep the compatibility

hsluoyz (changes requested):

object/user_webauthn.go
nb150301: I already revert the `originBackend`. Thank you for your careful review

nb150301:
@hsluoyz hi, please check this

minhthuong031103:
As in my case, I use 1 casdoor for multiple applications with different teams but can only set 1 origin URLs so in some cases, for example, the webAuthn it has to be the exact same Origin url = Web url so it only works in one application. I think this PR can help to resolve these problems

object/user_webauthn.go
```diff
@@ -20,26 +20,25 @@ import (
...
 	"github.com/go-webauthn/webauthn/protocol"
 	"github.com/go-webauthn/webauthn/webauthn"
 )
 
-func GetWebAuthnObject(host string) (*webauthn.WebAuthn, error) {
-	var err error
-
+func GetWebAuthnObjectByApplication(application *Application, host string) (*webauthn.WebAuthn, error) {
 	_, originBackend := getOriginFromHost(host)
+	if application.HomepageUrl != "" {
```
hsluoyz, line 29: application.HomepageUrl is usually not empty, you still break old systems

object/user_webauthn.go
```diff
@@ -20,26 +20,25 @@ import (
...
 	}
 
 	webAuthn, err := webauthn.New(&webauthn.Config{
-		RPDisplayName: conf.GetConfigString("appname"),      // Display Name for your site
-		RPID:          strings.Split(localUrl.Host, ":")[0], // Generally the domain name for your site, it's ok because splits cannot return empty array
-		RPOrigin:      originBackend,                        // The origin URL for WebAuthn requests
-		// RPIcon:     "https://duo.com/logo.png",           // Optional icon URL for your site
+		RPDisplayName: application.Name,                     // Display Name for your site
+		RPID:          strings.Split(localUrl.Host, ":")[0], // Generally the domain for your site, it's ok because splits cannot return empty array
+		RPOrigins:     []string{originBackend},              // The origin URL for WebAuthn requests
```
hsluoyz, line 41: Why don't use `RPOrigins: []string{host, originBackend}`?

controllers/webauthn.go
```diff
@@ -155,11 +156,6 @@ func (c *ApiController) WebAuthnSigninBegin() {
 func (c *ApiController) WebAuthnSigninFinish() {
 	responseType := c.Input().Get("responseType")
 	clientId := c.Input().Get("clientId")
-	webauthnObj, err := object.GetWebAuthnObject(c.Ctx.Request.Host)
```
hsluoyz, line 158: Why change the position?

controllers/webauthn.go
```diff
@@ -175,14 +171,6 @@ func (c *ApiController) WebAuthnSigninFinish() {
 		return
 	}
 
-	_, err = webauthnObj.FinishLogin(user, sessionData, c.Ctx.Request)
```
hsluoyz, line 178: Why change the position?

controllers/webauthn.go
```diff
@@ -112,12 +113,6 @@ func (c *ApiController) WebAuthnSignupFinish() {
 // @Success 200 {object} protocol.CredentialAssertion The CredentialAssertion object
 // @router /webauthn/signin/begin [get]
 func (c *ApiController) WebAuthnSigninBegin() {
-	webauthnObj, err := object.GetWebAuthnObject(c.Ctx.Request.Host)
```
hsluoyz, line 115: Why change the position?

controllers/webauthn.go
```diff
@@ -69,17 +69,18 @@ func (c *ApiController) WebAuthnSignupBegin() {
 // @Success 200 {object} controllers.Response "The Response object"
 // @router /webauthn/signup/finish [post]
 func (c *ApiController) WebAuthnSignupFinish() {
-	webauthnObj, err := object.GetWebAuthnObject(c.Ctx.Request.Host)
```
hsluoyz, line 72: Why change the position?

controllers/webauthn.go
```diff
@@ -33,18 +33,18 @@ import (
 // @Success 200 {object} protocol.CredentialCreation The CredentialCreationOptions object
 // @router /webauthn/signup/begin [get]
 func (c *ApiController) WebAuthnSignupBegin() {
-	webauthnObj, err := object.GetWebAuthnObject(c.Ctx.Request.Host)
```
hsluoyz, line 36: Why change the position?

hsluoyz (changes requested):

nb150301:
In my case: I have casdoor deploy at domain a.com and 3 applications at b.com, c.com and d.com. Currently, casdoor just support 1 due to only 1 rp.ID in the config. I need to change the position of the `GetWebAuthObject` to use the config homepage of application as rp.ID, that will let the webauthn can support multi applications. I need to get the application that the User belongs to to get the homepage URL

hsluoyz:
You didn't reply to each comment. Closed for before the bar of code quality

PR #3650, opened 2025-03-10
fix: del validate for 403

gongxianjin:
i use not localhost or allowed url
and fix this isssue
hope the system is more open and good for every one


hsluoyz:
The code in this PR doesn't make sense. You create security holes by bypassing the origin checks

PR #3658, opened 2025-03-13
fix: user id generation on create user with oauth

berejant:
Usually casdoor generate uuid for any new user registered at system.

For oauth casdoor user external id from oauth provider which mostly aren't valid UUID.
Also it not good practice to use external id as internal primary id. For example in case that two different providers have same id for two users in their system (let's say like user Bob have id 123456 in Facebook and user John has is 123456 in X).

For example compare users on screenshot: first is registered through casdoor, another three are user received from Google oauth
Знімок екрана 2025-03-13 о 10 36 19 дп

This fix make random uuid generation for any new user.

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -771,10 +771,7 @@ func (c *ApiController) Login() {
 						return
 					}
 
-					userId := userInfo.Id
-					if userId == "" {
-						userId = util.GenerateId()
-					}
+					userId := util.GenerateId()
 
 					user = &object.User{
 						Owner:             application.Organization,


hsluoyz:
Casdoor user ID doesn't have to be a UUID. No one said that. It first repsects the ID from OAuth, if there' none, then go to UUID as fallback

berejant:
@hsluoyz On your opinion can we add option to provider configuration `dontUserProviderId` ?

In our system we keep uuid for user and want to have approach in casdoor.

hsluoyz:
Not interested in adding it as an option. Too trivial thing.

PR #3687, opened 2025-03-29
fix: country icon lazy-load

WindSpiritSR:
DO NOT **ABUSE** the `preload`


hsluoyz:
Discussed and not planned

Issue #3655, opened 2025-03-12
Add provider hint to redirect URI


topgun100:
### Why

As a developer, I want to show a "Sign in with Google" button on my application. This improves the user experience:

- It's idiomatic and is how users expect social logins to work
- It helps the user to understand the available options (without navigating to another page).
- It allows the user to login with one click instead of two

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/7ee027ce-45c2-4643-975e-ac83da9b3ba5" />

Please can you add a "provider_hint" parameter hint to the redirect uri which takes the user straight through?

Keycloak calls this kc_idp_hint

hsluoyz:
It seems that Casdoor doesn't have the long button now, see: https://door.casdoor.com/applications/casbin/app-casnode

![Image](https://github.com/user-attachments/assets/4f2ec630-a239-4c59-af03-c74a2fc6e13b)

![Image](https://github.com/user-attachments/assets/071f6407-6c8a-4562-9e0f-0f3051dc95b8)

We want this UI back:

![Image](https://github.com/user-attachments/assets/7f94b6ca-22fd-4a08-a86a-5e7b8902ad1f)

topgun100:
@hsluoyz thanks for your input. Actually you can get the long buttons by setting "rule" "Big icon", however, that's not what I'm trying to achieve.

I want to have the buttons hosted on *my* application and not the Casdoor login page.

PR #3698, opened 2025-04-01
feat: can use provider_hint arg to do OAuth redirect automatically

dacongda:
Fix: https://github.com/casdoor/casdoor/issues/3655

https://github.com/user-attachments/assets/0a333e09-f2ed-4e57-8536-d98424b65b11

@topgun100 is this what you want ?


topgun100:
Yes, that is the functionality I wanted. 🙏

Is there a way to prevent the login screen from flashing momentarily (without resorting to CSS hacks)? The whole purpose of this is to tightly control the UX...

dacongda:
> Yes, that is the functionality I wanted. 🙏
> 
> Is there a way to prevent the login screen from flashing momentarily (without resorting to CSS hacks)? The whole purpose of this is to tightly control the UX...

I will try to fix this later

PR #3708, opened 2025-04-04
feat: update `github.com/golang-jwt/jwt` dependency to v5

Juneezee:
1. https://github.com/golang-jwt/jwt/security/advisories/GHSA-mh63-6h87-95cp
2. https://github.com/golang-jwt/jwt/security/advisories/GHSA-29wx-vh33-7x7r

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -340,7 +340,7 @@ func (c *ApiController) IntrospectToken() {
 
 	if application.TokenFormat == "JWT-Standard" {
 		jwtToken, err := object.ParseStandardJwtTokenByApplication(tokenValue, application)
-		if err != nil || jwtToken.Valid() != nil {
+		if err != nil {
 			// and token revoked case. but we not implement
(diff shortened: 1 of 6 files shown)


hsluoyz:
@Juneezee fix all CIs:
image

controllers/token.go
```diff
@@ -340,7 +340,7 @@ func (c *ApiController) IntrospectToken() {
 
 	if application.TokenFormat == "JWT-Standard" {
 		jwtToken, err := object.ParseStandardJwtTokenByApplication(tokenValue, application)
-		if err != nil || jwtToken.Valid() != nil {
+		if err != nil {
```
Juneezee, line 343: The `Valid()` method has been removed in v5. 

`jwt.ParseWithClaims` already validates the claims for us and will return `err` if there are any validation errors.

https://github.com/golang-jwt/jwt/blob/v5.2.2/parser.go#L116-L126

PR #3745, opened 2025-04-24
feat: add Pbkdf2DjangoCredManager

KhaledOmara:
Adding support for the django hashing.
Now, the existing project on django passwords can easily migrate to casdoor with the same password without need to reset the passwords.


KhaledOmara:
I've added pbkdf2-django hashing to the manager.go to support migrating from django apps.

PR #3760, opened 2025-05-02
refactor: use slices.Contains to simplify code

findnature:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/object/product.go
+++ b/object/product.go
@@ -16,6 +16,7 @@ package object
 
 import (
 	"fmt"
+	"slices"
 
 	"github.com/casdoor/casdoor/idp"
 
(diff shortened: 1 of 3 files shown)


hsluoyz:
We don't use Go 1.21 APIs for now to keep good backward compatibility

findnature:
> We don't use Go 1.21 APIs for now to keep good backward compatibility

Thank you for your reply. I see that the Go version in go.mod is already 1.21. Sorry for the disturbance.

Issue #3763, opened 2025-05-04
Bug Report: Microsoft Teams Webhook URL Pattern Too Strict (Power Automate Compatibility Issue)


SerhiiZahuba:
Problem sending a message to Teams

![Image](https://github.com/user-attachments/assets/0cd5d81f-5ecf-4826-a163-dd24e7b722e9)

When I click "Send test message" I get an error

`failed to send message to Microsoft Teams via webhook 'https://prod-06.westeurope.logic.azure.com:443/workflows/7d72835f7967/triggers/manual/paths/invoke?api-version=2016-06-01&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig=qKlYGnagAbKVw': webhook URL does not match one of expected patterns; got: "https://prod-06.westeurope.logic.azure.com:443/workflows/7d7285f47967/triggers/manual/paths/invoke?api-version=2016-06-01&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig=qKlYGgAbKVw", patterns: ^https:\/\/(?:.*\.webhook|outlook)\.office(?:365)?\.com: send notification`

hsluoyz:
@SerhiiZahuba you said "Casdoor currently enforces a strict regular expression check on the Microsoft Teams webhook URL, allowing only endpoints that match the pattern:", I didn't see Casdoor has such regex check. Can you provide the Casdoor source code link?

SerhiiZahuba:
Sorry, maybe I didn't describe the problem correctly (I asked gpt to describe it nicely but he couldn't). When I try to send a test message through teams, I get an error

`failed to send message to Microsoft Teams via webhook 'https://prod-206.westeurope.logic.azure.com:443/workflows/381db8d47ac46e46eb/triggers/manual/paths/invoke?api-version=2016-06-01&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig=PdrkqcrV0OkCzMqAOU': webhook URL does not match one of expected patterns; got: "https://prod-206.westeurope.logic.azure.com:443/workflows/381db8d47ac46e46eb/triggers/manual/paths/invoke?api-version=2016-06-01&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig=PdrkqcrV0OkCzMqAOU", patterns: ^https:\/\/(?:.*\.webhook|outlook)\.office(?:365)?\.com: send notification`

SerhiiZahuba:
[...]

PR #3764, opened 2025-05-04
Update go-teams-notify to v2.13.0

SerhiiZahuba:
Fix: https://github.com/casdoor/casdoor/issues/3763

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -101,7 +101,7 @@ require (
 	github.com/aliyun/aliyun-oss-go-sdk v2.2.2+incompatible // indirect
 	github.com/aliyun/credentials-go v1.3.10 // indirect
 	github.com/apistd/uni-go-sdk v0.0.2 // indirect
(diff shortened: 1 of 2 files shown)


hsluoyz:
@SerhiiZahuba you should not manually upgrade the "indirect" dependency, you should find out which one is the correct "direct" dependency. Upgrade it first.

image

SerhiiZahuba:
I understand the difference between direct and indirect dependencies. However, I’m not very experienced with Go internals, and I’m not sure how to correctly refactor the code to make go-teams-notify a direct dependency the way you expect.

If this is something that requires structural changes to how the library is used inside Casdoor, it might be better handled by someone more familiar with the codebase.

I’m happy to help test or verify the update, but I may not be able to complete the PR myself beyond updating the version.

Thanks again for your help and patience!

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/commit/8f307dd90725397f8a310774557b499f01cadc8d

PR #3784, opened 2025-05-13
feat: support adapter in app.conf logConfig

dacongda:
Fix: https://github.com/casdoor/casdoor/discussions/3783


main.go
```diff
@@ -77,7 +77,15 @@ func main() {
...
 	// beego.BConfig.WebConfig.Session.SessionCookieSameSite = http.SameSiteNoneMode
 
-	err := logs.SetLogger(logs.AdapterFile, conf.GetConfigString("logConfig"))
+	logAdapter := conf.GetConfigString("logAdapter")
+	if logAdapter == "" {
+		logAdapter = logs.AdapterFile
+	}
+
+	var err error
+	if logAdapter != "console" {
```
MattTaylorQuodd, line 86: I don't think we want this if clause because users would want (I want) to be able to set the level

```
logs.SetLogger(logs.AdapterConsole, {"level":1})
```

`logs.AdapterFile` is [a string constant](https://github.com/beego/beego/blob/master/core/logs/log.go#L62) so you could write this change like so:
```go
err := logs.SetLogger(conf.GetConfigString("logAdapter"), conf.GetConfigString("logConfig"))
if err != nil {
	panic(err)
}
```

MattTaylorQuodd:
If `app.conf` needs a special rule, you can set it like

```
conf/app.conf=text eol=lf
```

dacongda:
> text=auto

I just confuse with the diff error of github

> If `app.conf` needs a special rule, you can set it like
> 
> ```
> conf/app.conf=text eol=lf
> ```
> 
> see the [git docs](https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings#example)

I'm just confused about the github error diff

dacongda:
> > text=auto
> 
> I just confuse with the diff error of github
> 
> > If `app.conf` needs a special rule, you can set it like
> > ```
> > conf/app.conf=text eol=lf
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > see the [git docs](https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings#example)
> 
> I'm just confused about the github error diff

finally,I fix this, maybe this error is cause by goland, it will automatically set my git's core.autocrlf to true

PR #3794, opened 2025-05-16
feat: fix missing redirect_uri in DingTalk registration

hzzcc:
When binding DingTalk on the registration page, the app redirects to the /apps route afterward, instead of redirecting to the user's original intended page.


hsluoyz:
@dacongda plz review

dacongda:
could you provide a demo video that what this pr fix?

dacongda:
after run the code, I think this is not a good idea. for example, some application will redirect user to a callback page instead of index page, so if use this pr, it will display error page without code and state. Maybe redirect user to login page if we are using oauth to login will be better

hsluoyz:
Closed for now. @dacongda maybe you can make a PR to ensure that the code change won't break anything

Issue #3806, opened 2025-05-19
User name in JWT token is different from UserInfo


maggch97:
With OIDC integration, I observed that the name in token is different from the name from api/userinfo. First name is Name(Unique, string-based ID) and the second one is Display Name.

How can I map the Display Name to the name field in the token?

hsluoyz:
@dacongda

PR #3809, opened 2025-05-20
feat: use the standard user struct for JWT-Standard to get a correct userinfo

dacongda:
fix: #3806 

@maggch97 can you pull this repo and help me test this pr? You need to set Token format to JWT-Standard first.


maggch97:
Thanks, I will try it tomorrow.

maggch97:
I verified locally that the `name` in the JWT `claims` now is same with the `name` from `userinfo`.

PR #3820, opened 2025-05-22
feat: can redirect user to login page after linking provider in prompt page

dacongda:
Replace: https://github.com/casdoor/casdoor/pull/3794


dacongda:
demo video
https://github.com/user-attachments/assets/9004e8e3-16b0-437d-805b-b4e17f62ebe4

Issue #3849, opened 2025-05-30
Enhancement: Add Trivy Vulnerability Scanning to the Pipeline


guimoreirar:
It could be a good idea to add a Trivy security scan step to the CI pipeline. This would help check for vulnerabilities in Docker images or dependencies before releasing. By configuring the scan to fail on detected vulnerabilities (exit-code: 1), it would prevent publishing containers with known security issues

like this:

```
      - name: Scan casdoor image with Trivy
        if: github.repository == 'casdoor/casdoor' && github.event_name == 'push' && steps.should_push.outputs.push=='true'
        uses: aquasecurity/trivy-action@master
        with:
          image-ref: 'casbin/casdoor:${{steps.get-current-tag.outputs.tag }}'
          format: 'table'
          exit-code: '1'
          ignore-unfixed: true
          severity: 'HIGH,CRITICAL'

      - name: Push All In One Version to Docker Hub
        uses: docker/build-push-action@v3
        if: github.repository == 'casdoor/casdoor' && github.event_name == 'push' && steps.should_push.outputs.push=='true'
        with:
          context: .
          target: ALLINONE
          platforms: linux/amd64,linux/arm64
          push: true
          tags: casbin/casdoor-all-in-one:${{steps.get-current-tag.outputs.tag }},casbin/casdoor-all-in-one:latest

      - name: Scan casdoor-all-in-one image with Trivy
        if: github.repository == 'casdoor/casdoor' && github.event_name == 'push' && steps.should_push.outputs.push=='true'
        uses: aquasecurity/trivy-action@master
        with:
          image-ref: 'casbin/casdoor-all-in-one:${{steps.get-current-tag.outputs.tag }}'
          format: 'table'
          exit-code: '1'
          ignore-unfixed: true
          severity: 'HIGH,CRITICAL'
```

hsluoyz:
@guimoreirar No, we will not add the pipeline.

We will only find a contributor to manually run the Trivy command and see if we can have some dependency updates manually. Note,  this is totally a manual and ad-hoc process, NOT a pipeline!!!

Someone provided the manual command:

```
docker run aquasec/trivy image casbin/casdoor:v1.924.0
docker run aquasec/trivy image casbin/casdoor-all-in-one:v1.924.0
```

PR #3852, opened 2025-06-02
feat: add trivy vulnerability scanning to the pipeline

guimoreirar:
https://github.com/casdoor/casdoor/issues/3849


hsluoyz:
@guimoreirar it's not suitable to use Trivy as a hard dependency in our CI flow now. It's going too far. Actually I don't know what Trivy is before this and how many big OSS projects use it as hard dependency. And I don't think everything it reports are 100% useful for us. e.g., if there's a hole in a dependency, it must mean there's a hole in our project? I dare not. It just has a lot of such cases in future and cause our process to be slowed down.

We can manually do a dependency upgrade time to time (like every month), but won't integrate it to CI now.

Issue #3860, opened 2025-06-05
Application name with "/" inside is not possible to modify


progys:
* Create application (any) with a name `Something/1` and save it
* Try to enter edit page by clicking "edit" or on the `Something/1` link - 404 page is presented.

This happens because of incorrect URL encoding. 

Maybe using "/" inside the application name should be forbidden?

hsluoyz:
@progys hi, can you make a PR to fix it?

progys:
What would be the preferred way to fix it? 

I see two possibilities:

- Allow entering any characters and make sure routing works. (more involved)
- Disallow certain characters like `/` and others which break routing. (easier, but not user friendly)

hsluoyz:
@progys Disallow "/"

PR #3877, opened 2025-06-12
feat: add FieldValidationFilter to check object names

okatu-loli:
fix: #3860

### DIFF
--- a/routers/field_validation_filter.go
+++ b/routers/field_validation_filter.go
@@ -17,8 +17,9 @@ package routers
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/beego/beego/context"
 	"strings"
+
+	"github.com/beego/beego/context"
 )
 
 var forbiddenChars = "/"
--- a/routers/field_validation_filter.go
+++ b/routers/field_validation_filter.go
@@ -1,4 +1,4 @@
-// Copyright 2022 The Casdoor Authors. All Rights Reserved.
+// Copyright 2025 The Casdoor Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


hsluoyz:
@dacongda plz review

dacongda:
fix linter issue

dacongda:
add a extra filter to do this is not a good  idea,  it will have unnecessary cost

okatu-loli:
> add a extra filter to do this is not a good idea, it will have unnecessary cost

@dacongda I noticed that the existing filters are organized based on different responsibilities, so I added an additional filter. Do you have any suggestions for improvement?

okatu-loli:
> add a extra filter to do this is not a good idea, it will have unnecessary cost

If you're referring to the potential performance overhead of adding a new filter to address issue #3860, the main reason I chose to implement it as a separate filter instead of putting the logic directly in the controller is to keep the code structure clean and maintainable by centralizing the validation logic.

dacongda:
You can put this code in `util/validation.go` and check name in object package

okatu-loli:
> You can put this code in `util/validation.go` and check name in object package

Yeah, but it feels less flexible. With the field_validation_filter, we can handle not just name but other fields later too — it's easier to scale this way.

dacongda:
maybe we should ask @hsluoyz

hsluoyz:
@okatu-loli @dacongda we only want to write this check code once. The filter is a good place. You need to check the "add-xxx" and "update-xxx" prefix at first to make sure what APIs you want to check.

routers/field_validation_filter.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/beego/beego/context"
+)
+
+var forbiddenChars = "/"
+
+var fieldValidationRules = map[string][]string{
+	"/api/add-application":    {"name"},
```
hsluoyz: check the "add-xxx" and "update-xxx" prefix, not only application APIs

routers/field_validation_filter.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2022 The Casdoor Authors. All Rights Reserved.
```
hsluoyz: 2025

routers/field_validation_filter.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	fields, ok := fieldValidationRules[ctx.Request.URL.Path]
+	if !ok {
+		return
+	}
+
+	var requestData map[string]interface{}
+	if err := json.NewDecoder(ctx.Request.Body).Decode(&requestData); err != nil {
+		responseError(ctx, "Invalid JSON body")
```
hsluoyz: Maybe not JSON should not fail?

routers/field_validation_filter.go
```diff
@@ -0,0 +1,56 @@
...
+		return
+	}
+
+	for _, field := range fields {
+		if value, ok := requestData[field].(string); ok {
```
hsluoyz: Only check "name", not other fields

hsluoyz (changes requested):

routers/field_validation_filter.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/beego/beego/context"
+)
+
+var forbiddenChars = "/"
```
hsluoyz: Not only this "/" is forbidden

hsluoyz (changes requested):

routers/field_validation_filter.go
dacongda: `/?:@=&;` is reserved url char

okatu-loli:
test video

https://github.com/user-attachments/assets/0073ef8d-c9fb-4e56-8299-485e3e73d227

okatu-loli:
I plan to open a new PR to solve the runtime error encountered in the test video

routers/field_validation_filter.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/beego/beego/context"
+)
+
+var forbiddenChars = `/?:@#&%=+;`
```
ReGetALife, line 26: Perhaps ’@‘ should not  be forbidden, since user's name will contain '@' when you turn on "Use Email as username". And you can't update user anymore in that case.

hsluoyz: @ReGetALife fixed in: https://github.com/casdoor/casdoor/commit/f4ad2b4034a57f815421173a5ef6014e51a706fa

Issue #3880, opened 2025-06-12
"/api/get-user" does not respect ViewRules


karatekaneen:
### What happens

I created an organization where I put "Is profile public" to false and set all fields to view role "self" or "admin".

![Image](https://github.com/user-attachments/assets/c127da2d-7a18-4fc2-ac92-081cc1bc13f8)

I then created two users, none of which are admins, `testorg/user1` and `testorg/user2`.

**Now I expect when I get another regular user that I should only see the fields `owner` and `id`.**

### Actual Behaviour

When authenticated as user2 I request `http://localhost:8000/api/get-user?id=testorg/user1` and I get all data.

<details>
<summary>**Full Response here**</summary>

```json
{
  "status": "ok",
  "msg": "",
  "sub": "",
  "name": "",
  "data": {
    "owner": "testorg",
    "name": "User1",
    "createdTime": "2025-06-12T15:00:35+02:00",
    "updatedTime": "2025-06-12T14:36:29Z",
    "deletedTime": "",
    "id": "88809f58-f584-441f-ad7c-3fd8f0e73431",
    "externalId": "",
    "type": "normal-user",
    "password": "***",
    "passwordSalt": "",
    "passwordType": "plain",
    "displayName": "User1",
    "firstName": "",
    "lastName": "",
    "avatar": "https://cdn.casbin.org/img/casbin.svg",
    "avatarType": "",
    "permanentAvatar": "",
    "email": "user1@example.com",
    "emailVerified": false,
    "phone": "22970802327",
    "countryCode": "US",
    "region": "SE",
    "location": "Ankeborg",
    "address": [],
    "affiliation": "Example Inc.",
    "title": "Foobar",
    "idCardType": "",
    "idCard": "",
    "homepage": "hello.com",
    "bio": "My bio",
    "tag": "hello there",
    "language": "",
    "gender": "",
    "birthday": "",
    "education": "",
    "score": 2000,
    "karma": 0,
    "ranking": 0,
    "balance": 0,
    "currency": "",
    "isDefaultAvatar": false,
    "isOnline": false,
    "isAdmin": false,
    "isForbidden": false,
    "isDeleted": false,
    "signupApplication": "",
    "hash": "",
    "preHash": "",
    "accessKey": "",
    "accessSecret": "",
    "accessToken": "",
    "createdIp": "",
    "lastSigninTime": "",
    "lastSigninIp": "",
    "github": "",
    "google": "",
    "qq": "",
    "wechat": "",
[...]

PR #3882, opened 2025-06-16
feat: get-user API respects org's account item's view rules now

dacongda:
Fix: https://github.com/casdoor/casdoor/issues/3880


karatekaneen:
@dacongda Thank you, awesome work!

PR #3894, opened 2025-06-24
feat: Supports smooth migration of password credentials

### LINKED ISSUE #3278: Switching the organization password type should re-create the passwords for the organization
In the following scenario, I have changed the default built-in organization's password type from plain to bycrypt.
The 1st user, Admin, is part of that Org.
After changing the password type, the plaintext password remained stored in the database.

Danger of locking oneself out.

In the case of the password being plain, it should be hased with the new password type.
There's a catch 22 situation when you switch from non-plain to other non-plain password types.
When switching password types, there should be a warning message, that the admin user should modify their password.
Of course it's logical to do so, but I don't believe this behavior is documented.

Of course how often does it happen that the password type changes? Probably once for the built-in org.
Speaking of, the default password type for the built-in org should be set to argon2d.
Storing passwords in plaintext is a liability in at least 1 country on the planet and can cost a lot of money in case of a breach.

So, TL;DR
if previous password type == plain, automatically generate new password hashes for all users of the org
if new password type == plain, since we don't know the unhashed password, do nothing.
Make argon2d the new default password type for the built-in organization

Vickko:
### Background
In the current implementation of Casdoor, there is a process flow issue concerning changes to password policies.

Currently, when a user is created, Casdoor records the password type used by their organization at that moment. However, it does not store a snapshot of the organization-level password salt. If the organization subsequently modifies its password policy, the user's recorded password type and hashed password remain unchanged unless the user actively changes their password.

This design, while allowing existing users to log in with the old password type after a policy change, introduces two significant issues:

- "Hard Break" on Salt Changes: When an organization changes its password salt, any user whose password type relies on this org-level salt is immediately unable to log in. The root cause is that, unlike the password type, the organization's salt is not snapshotted for each user. The system can only use the latest organization salt for hash verification, which will inevitably fail against a hash generated with the old salt. For instance, an organization switching from one salt-dependent password type to another would likely change the salt as well, triggering this failure scenario.

- Lack of a Seamless Migration Mechanism: The system currently lacks an "Update on Login" mechanism. Ideally, after a user successfully authenticates with legacy credentials (e.g., an old password type and salt), the system should immediately re-hash and store the password using the new organization policy and the provided plaintext password. This would enable a seamless migration of user credentials to the new policy without any user friction. This is particularly critical given that Casdoor's default password type is plain, which poses a potential security risk for administrators unfamiliar with this default behavior. The absence of a smooth migration path exacerbates this risk.

### Impact
The current password policy implementation has several negative consequences for both system administrators and end-users:

- Operational Difficulties: Administrators face significant challenges when attempting to update the organization's password salt if it has been used previously.
- Poor User Experience: To enforce a new password policy or update the salt, the only viable solution is to force all affected users to reset their passwords via the "Forgot Password" flow. This severely degrades the user experience and satisfaction.
- Potential Security Risks: The high operational cost and negative user impact of updating security policies may lead administrators to postpone or entirely avoid necessary security upgrades, leaving the system exposed to potential vulnerabilities for extended periods.
### Solution
This PR resolves the aforementioned issues by introducing an automatic and seamless credential migration mechanism. Upon each login or password change, the system will check the user's credentials against the current organization policy and, if necessary, automatically update the password hash, password type, and a snapshot of the organization's password salt.

### Implementation Details
This PR introduces the following changes:

- Added PasswordOrganizationSaltSnapshot Field: A new field has been added to the User struct (in object/user.go) to store a snapshot of the organization's salt that was used to hash the user's current password. This ensures that even if the organization's salt changes, the system can still validate the user's old password using the correct legacy salt.
- Modified CheckPassword Logic: The core logic in the CheckPassword function (in object/check.go) has been updated. It now considers a user's password policy "outdated" if it mismatches the current organization policy. In such cases, it first attempts to validate the password using the user's snapshotted salt. If successful, it immediately updates the user's password hash, password type, and salt snapshot to the new policy, achieving a seamless migration.
- Updated SetPassword Function: The SetPassword function (in controllers/user.go) has been modified to synchronize the organization salt snapshot whenever a user changes their password.

### Testing Completed
I have thoroughly tested multiple scenarios to ensure these changes function as expected and do not introduce regressions:

- Switching organization policies between different password types (plain, md5-salt, bcrypt).
- Users with legacy credentials successfully logging in and having their credentials automatically upgraded after a policy change.
- Users with a legacy salt successfully logging in and having their credentials automatically upgraded after the organization's salt was updated.
- The new user creation flow remains unaffected.
- The user-initiated password change flow functions correctly.

### Related Issues
Fixes #3278 : This PR addresses the problem outlined in this issue by the way. While the original issue's description may not be perfectly precise, the changes herein fully resolve its core concerns.

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -573,8 +573,13 @@ func (c *ApiController) SetPassword() {
 	targetUser.NeedUpdatePassword = false
 	targetUser.LastChangePasswordTime = util.GetCurrentTime()
 
+	// Set organization salt snapshot for password validation
+	if organization.PasswordSalt != "" {
+		targetUser.PasswordOrganizationSaltSnapshot = organization.PasswordSalt
+	}
+
 	if user.Ldap == "" {
-		_, err = object.UpdateUser(userId, targetUser, []string{"password", "need_update_password", "password_type", "last_change_password_time"}, false)
+		_, err = object.UpdateUser(userId, targetUser, []string{"password", "need_update_password", "password_type", "last_change_password_time", "password_organization_salt_snapshot"}, false)
 	} else {
 		if isAdmin {
 			err = object.ResetLdapPassword(targetUser, "", newPassword, c.GetAcceptLanguage())
--- a/object/check.go
+++ b/object/check.go
@@ -249,22 +249,35 @@ func CheckPassword(user *User, password string, lang string, options ...bool) er
 	if passwordType == "" {
 		passwordType = organization.PasswordType
 	}
-	credManager := cred.GetCredManager(passwordType)
-	if credManager != nil {
-		if organization.MasterPassword != "" {
-			if password == organization.MasterPassword || credManager.IsPasswordCorrect(password, organization.MasterPassword, "", organization.PasswordSalt) {
-				return resetUserSigninErrorTimes(user)
-			}
-		}
 
-		if credManager.IsPasswordCorrect(password, user.Password, user.PasswordSalt, organization.PasswordSalt) {
-			return resetUserSigninErrorTimes(user)
-		}
+	isOutdated := passwordType != organization.PasswordType || (organization.PasswordSalt != "" && user.PasswordOrganizationSaltSnapshot != organization.PasswordSalt)
 
-		return recordSigninErrorInfo(user, lang, enableCaptcha)
-	} else {
-		return fmt.Errorf(i18n.Translate(lang, "check:unsupported password type: %s"), organization.PasswordType)
+	orgSalt := organization.PasswordSalt
+	if isOutdated {
+		orgSalt = user.PasswordOrganizationSaltSnapshot
 	}
+
+	credManager := cred.GetCredManager(passwordType)
+	if credManager == nil {
+		return fmt.Errorf(i18n.Translate(lang, "check:Unsupported password type: %s"), passwordType)
+	}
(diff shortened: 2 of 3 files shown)


hsluoyz:
Hi @Vickko 

1. Create issue first before PR
2. You can use ChatGPT to polish your English, but don't use ChatGPT to generate all the texts. It makes my head hurt to read the ChatGPT-flavor long texts :)
3. Can you un-salt the user password and re-salt it with the new salt? Is it reversable?
4. Go to the community to discuss it first for a complicated problem

Vickko:
@hsluoyz Hi,
Sorry for inconvenience, I just worry about my English grammar, so I translate my words into English by chatGPT :(
In this case I will write it on my own, and try to use simple word and short sentence.

Later I will do following things, and when I finish, I will @ you to call you back for a review :)
1. Create a issue to describe the original problem, and we discuss it and its solution there
2. Edit this PR's description into a TLDR version, and use simple statements, and answer your question 3 then

I will leave this PR open temporarily, and if we come up with a conclusion that my solution in this PR works well, we come back and deal with it here again; If we decide to use another way to solve this problem, I will close it and come back with a new solution in a new PR. Is that OK?

hsluoyz:
This PR can be closed for now.

Issue #3896, opened 2025-06-24
Normal user cannot subscribe a plan via buying a product binds with a plan


Corgi14:
due to the current logic in [controllers/product.go#L185](https://github.com/casdoor/casdoor/blob/master/controllers/product.go#L185), user cannot buy a product which binds with a plan, and subscribe the specific plan after purchasing.

<img width="1604" alt="Image" src="https://github.com/user-attachments/assets/6070864c-1b0a-4230-9bb2-70a510f27785" />

I searched error message in repo and found out that 

```golang
if paidUserName != "" && !c.IsAdmin() {
	c.ResponseError(c.T("general:Only admin user can specify user"))
	return
}
```
code below was added in [Commit 7ae067e](https://github.com/casdoor/casdoor/commit/7ae067e369a5d9bf4299c72eaae5074405c6a265).

Plz check if I was using casdoor in a wrong way.

PR #3897, opened 2025-06-24
feat: fix check bug to allow logged-in users to buy product

Corgi14:
Fix: https://github.com/casdoor/casdoor/issues/3896

### DIFF
--- a/controllers/product.go
+++ b/controllers/product.go
@@ -183,7 +183,7 @@ func (c *ApiController) BuyProduct() {
 	owner, _ := util.GetOwnerAndNameFromId(id)
 	userId := util.GetId(owner, paidUserName)
 	if paidUserName != "" && paidUserName != c.GetSessionUsername() && !c.IsAdmin() {
-		c.ResponseError(c.T("general:Only admin user can specify another user"))
+		c.ResponseError(c.T("general:Only admin user can specify user"))
 		return
 	}
 	if paidUserName == "" {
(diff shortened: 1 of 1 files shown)


hsluoyz:
@Corgi14 you need to figure out why the initial author needs to add that check, what problem does it resolve? Not just remove it simply because it "blocked" your way

Corgi14:
> @Corgi14 you need to figure out why the initial author needs to add that check, what problem does it resolve? Not just remove it simply because it "blocked" your way

Background
The original check:
```golang
if paidUserName != "" && !c.IsAdmin() {
    c.ResponseError(c.T("general:Only admin user can specify user"))
    return
}
```
was introduced to prevent normal users from purchasing products on behalf of other users, which is a valid security safeguard.

### The Issue
However, when using Casdoor's built-in pricing page (which is the standard way for users to subscribe to a Plan), the frontend always passes the logged-in user’s userName to the /api/buy-product API — even when the user is buying a plan for themselves.

As a result, normal users are unable to subscribe to any plan, receiving the error:

> "general:Only admin user can specify user"

This behavior breaks the expected self-service subscription workflow described in Casdoor’s own documentation.

### The Fix
Instead of removing the check entirely, this PR adjusts the logic to:
```golang
if paidUserName != "" && paidUserName != c.GetSessionUsername() && !c.IsAdmin() {
    c.ResponseError(c.T("general:Only admin user can specify another user"))
    return
}
```
This ensures:

 - 🔒 Users can only specify themselves

 - 🚫 Users cannot purchase for others (unless they are admins)

 - ✅ Admins can still purchase for any user

### Summary
This fix keeps the original intention of preventing privilege escalation, but allows normal users to purchase subscriptions for themselves — which is required for the core functionality of Plans and Subscriptions to work.

PR #3907, opened 2025-06-26
fix: allow users to purchase plans immediately after registration

Corgi14:
>Apologies for the oversight in the previous [pr 3897](https://github.com/casdoor/casdoor/pull/3897). This updated PR correctly addresses the root cause by removing the admin-only check, allowing users to purchase plans immediately after registration as intended.

### Background
The original check:
```golang
if paidUserName != "" && !c.IsAdmin() {
    c.ResponseError(c.T("general:Only admin user can specify user"))
    return
}
```
was introduced to prevent normal users from purchasing products on behalf of other users, which is a valid security safeguard in session-based authentication flows.

### The Issue
However, in Casdoor’s default registration-to-subscription flow, users are allowed to register and immediately purchase a plan without logging in. After registration, users are redirected to the built-in pricing page, and their userName is passed via the URL query parameter to the /api/buy-product API.

At this point:

The user is not logged in, so session information (such as c.GetSessionUsername()) is unavailable

But the userName parameter is valid and intended to reflect the newly registered user

As a result, the original check treats all users as unauthorized and blocks them from purchasing their own plans — even though this is exactly what the default Casdoor workflow is designed to support.

### The Fix
This PR removes the check entirely:

```golang
diff
if paidUserName != "" && !c.IsAdmin() {
    c.ResponseError(c.T("general:Only admin user can specify user"))
    return
}
```
This change restores the expected behavior:

✅ Users can register and subscribe immediately (without logging in again)

✅ The userName parameter from the query string is treated as the user's identity

✅ Admins can still purchase for others

⚠️ Security remains under control, as the frontend controls the userName flow during registration

### Summary
Removing this check aligns the backend behavior with Casdoor’s intended self-service subscription model. It allows newly registered users to seamlessly proceed to plan purchase without requiring an additional login step, improving both UX and functional completeness of the subscription system.

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -309,6 +309,11 @@ func (c *ApiController) Signup() {
 	c.Ctx.Input.SetParam("recordUserId", user.GetId())
 	c.Ctx.Input.SetParam("recordSignup", "true")
 
+	// Store the registered user's ID into session to simulate login.
+	// This allows BuyProduct and other APIs to recognize the current user,
+	// even though the user hasn't gone through a separate login process.
+	c.SetSessionUsername(user.GetId())
+
 	userId := user.GetId()
 	util.LogInfo(c.Ctx, "API: [%s] is signed up as new user", userId)
 
--- a/controllers/product.go
+++ b/controllers/product.go
@@ -182,10 +182,6 @@ func (c *ApiController) BuyProduct() {
 	paidUserName := c.Input().Get("userName")
 	owner, _ := util.GetOwnerAndNameFromId(id)
 	userId := util.GetId(owner, paidUserName)
-	if paidUserName != "" && paidUserName != c.GetSessionUsername() && !c.IsAdmin() {
-		c.ResponseError(c.T("general:Only admin user can specify user"))
-		return
-	}
 	if paidUserName == "" {
 		userId = c.GetSessionUsername()
 	}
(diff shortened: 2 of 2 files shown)


hsluoyz:
@dacongda plz review

dacongda:
I don't think remove this check is a good idea, we can set session after register instead of remove this check

hsluoyz:
@dacongda I agree with you. @Corgi14 Allowing to pay as another user without restriction is just not a good idea (imagine that dirty money has purchased an item on behalf of your name, the police will find you!! So dangerous!!).

@dacongda do we support keeping the session right after user account registration?

dacongda:
> @dacongda I agree with you. @Corgi14 Allowing to pay as another user without restriction is just not a good idea (imagine that dirty money has purchased an item on behalf of your name, the police will find you!! So dangerous!!).
> 
> @dacongda do we support keeping the session right after user account registration?

nope, we just support auto sign in when application has prompt page

```
if application.HasPromptPage() && user.Type == "normal-user" {
	// The prompt page needs the user to be signed in
}
```

Corgi14:
> > @dacongda I agree with you. @Corgi14 Allowing to pay as another user without restriction is just not a good idea (imagine that dirty money has purchased an item on behalf of your name, the police will find you!! So dangerous!!).
> > @dacongda do we support keeping the session right after user account registration?
> 
> nope, we just support auto sign in when application has prompt page
> 
> ```
> if application.HasPromptPage() && user.Type == "normal-user" {
> 	// The prompt page needs the user to be signed in
> 	c.SetSessionUsername(user.GetId())
> }
> ```

Totally agree with u. I also agree with "set session after register". Do u have a plan setting user info to current session?

Corgi14:
@hsluoyz @dacongda 
Hello there,
I've `SetSessionUsername` when `Signup` succeed, plz check.

hsluoyz:
@Corgi14 no more work needed for this PR. We will work on it by ourselves

PR #3909, opened 2025-06-26
fix: update token cleanup logic to query all tokens and adjust cron schedule

Attack825:


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/3910

Issue #3914, opened 2025-06-28
[feature] Add support for per-account MFA validity period in org setting to reduce repeated prompts


hsluoyz:
Support setting MFA validity period per account. Within the validity period, the same account can log in again without being prompted for MFA verification, avoiding repeated code entry.

![Image](https://github.com/user-attachments/assets/f05aed5f-c30b-455e-bd26-a4c3a48b6776)

PR #3917, opened 2025-06-30
feat: add support for per-account MFA validity period in org setting to reduce repeated prompts

Attack825:
Fix: https://github.com/casdoor/casdoor/issues/3914

https://github.com/user-attachments/assets/0209eedc-ddc6-4ae7-be1d-ab9d37e9f39b

### DIFF
--- a/web/src/locales/id/data.json
+++ b/web/src/locales/id/data.json
@@ -83,7 +83,7 @@
     "Logged in successfully": "Berhasil masuk",
     "Logged out successfully": "Berhasil keluar dari sistem",
     "MFA remember time": "MFA remember time",
-    "MFA remember time - Tooltip": "MFA remember time - Tooltip",
+    "MFA remember time - Tooltip": "The duration for which the account is remembered as trusted after a successful MFA login",
     "Multiple Choices": "Multiple Choices",
     "New Application": "Aplikasi Baru",
     "No verification": "No verification",


dacongda:
I think this pr is not relevant to the issue you mentioned

Attack825:
> I think this pr is not relevant to the issue you mentioned

I sincerely apologize for the oversight. this PR should address a different issue. I'll correct the linked issue and update for review.

dacongda:
according the code, you set a session named `mfaVerifiedAt` to verify time, but if you login userA with mfa, and then logout, then you login UserB, which is required mfa too, it seems that you can jump mfa verify

controllers/auth.go
```diff
@@ -362,6 +362,13 @@ func checkMfaEnable(c *ApiController, user *object.User, organization *object.Or
 	}
 
 	if user.IsMfaEnabled() {
+		mfaVerifiedAtKey := "mfaVerifiedAt_" + user.GetId()
+		lastMfaVerifiedAt, ok := c.GetSession(mfaVerifiedAtKey).(int64)
+		if ok && lastMfaVerifiedAt > 0 {
+			if time.Now().Unix()-lastMfaVerifiedAt < 12*3600 {
```
dacongda: time shouldn't be fixed, you can add a external field or set `EnableMfaExpiry` to a int, if `EnableMfaExpiry` equal to zero, `EnableMfaExpiry` is disable

web/src/auth/mfa/MfaAuthVerifyForm.js
```diff
@@ -31,9 +31,9 @@ export function MfaAuthVerifyForm({formValues, authParams, mfaProps, application
   const [mfaType, setMfaType] = useState(mfaProps.mfaType);
   const [recoveryCode, setRecoveryCode] = useState("");
 
-  const verify = ({passcode}) => {
+  const verify = ({passcode, enableMfaExpiry}) => {
```
dacongda, line 34: `enableMfaExpiry` should never pass by user's request

dacongda (changes requested):

controllers/auth.go
```diff
@@ -355,13 +355,22 @@ func isProxyProviderType(providerType string) bool {
...
 	if object.IsNeedPromptMfa(organization, user) {
-		// The prompt page needs the user to be srigned in
+		// The prompt page needs the user to be signed in
 		c.SetSessionUsername(user.GetId())
 		c.ResponseOk(object.RequiredMfa)
 		return true
 	}
 
 	if user.IsMfaEnabled() {
+		mfaVerifiedAtKey := "mfaVerifiedAt_" + user.GetId()
```
dacongda: use `mfaVerifiedAtKey` only, admin can set expire time in application setting to determine how long will mfa expire(default 12 hour), then if user enable keep login, backend will get expire time from application and set expiredAt session base on this time, enableMfaExpiry could be a bool value for authForm.

dacongda: Because I initially mistook `EnableMfaExpiry ` for a field in the application, I gave incorrect advice.

Attack825: Could you share some advice on how to enable admins to set MFA expiration times effectively?

dacongda: add a field named mfaExpiredTime to application

dacongda: I will make a pr for this branch to assist you

Attack825:
https://github.com/user-attachments/assets/a5558461-68d8-4e62-8072-53097ca1dc1b

Attack825:
https://github.com/user-attachments/assets/9271b2fd-4a1e-4e67-b641-70967e759acf

web/src/auth/mfa/MfaVerifyTotpForm.js
```diff
@@ -54,6 +64,14 @@ export const MfaVerifyTotpForm = ({mfaProps, onFinish}) => {
           }}
         />
       </Form.Item>
+      <Form.Item
+        name="enableMfaExpiry"
+        valuePropName="checked"
+      >
+        <Checkbox>
+          {i18next.t("mfa:Remember this device for ") + (organization?.mfaRememberInHours || 12) + " hours"}
```
dacongda: i18n cannot handle this

dacongda:
have you test i18n?

Attack825:
> have you test i18n?

image

dacongda:
ok

web/src/locales/en/data.json
```diff
@@ -603,7 +603,8 @@
     "You have enabled Multi-Factor Authentication, please enter the TOTP code": "You have enabled Multi-Factor Authentication, please enter the TOTP code",
     "Your email is": "Your email is",
     "Your phone is": "Your phone is",
-    "preferred": "preferred"
+    "preferred": "preferred",
+    "Remember this device for {{hours}} hours": "Remember this device for {{hours}} hours"
```
dacongda: use `{{var}}` to support this is not a good idea, when you run `go test` in `i18n` folder to generate i18n translation, you will find that this will be remove, you can refer to `i18next.t("login:Sign in with {type}").replace("{type}", type);` to implement this.

dacongda (changes requested):

web/src/locales/en/data.json
dacongda: and you should run `go test` in `i18n` folder to generate i18n text

Attack825: > use `{{var}}` to support this is not a good idea, when you run `go test` in `i18n` folder to generate i18n translation, you will find that this will be remove, you can refer to `i18next.t("login:Sign in with {type}").replace("{type}", type);` to implement this.

I did think of using "replace" at the time, but GPT didn't recommend it. Then, after checking the i18next documentation, I went with {{}} instead. However, I overlooked go test, so I apologize for that.

dacongda (approved):

web/src/locales/en/data.json
```diff
@@ -82,6 +82,8 @@
     "Left": "Left",
     "Logged in successfully": "Logged in successfully",
     "Logged out successfully": "Logged out successfully",
+    "Mfa expire": "Mfa expire",
```
hsluoyz: Use this: https://casdoor.org/docs/contributing#how-to-update-i18n-data to update i18n

hsluoyz: UI should always be "MFA", not "Mfa"

controllers/auth.go
```diff
@@ -355,13 +355,20 @@ func isProxyProviderType(providerType string) bool {
...
-		// The prompt page needs the user to be srigned in
+		// The prompt page needs the user to be signed in
 		c.SetSessionUsername(user.GetId())
 		c.ResponseOk(object.RequiredMfa)
 		return true
 	}
 
 	if user.IsMfaEnabled() {
+		mfaVerifiedAtKey := object.MfaExpiredAt + user.GetId()
+		mfaExpiredAt, ok := c.GetSession(mfaVerifiedAtKey).(int64)
```
hsluoyz: Why don't save to DB's User table instead of session? DB is better than session

controllers/auth.go
```diff
@@ -355,13 +355,20 @@ func isProxyProviderType(providerType string) bool {
...
+		if ok && mfaExpiredAt > 0 {
+			if time.Now().Unix() <= mfaExpiredAt {
```
hsluoyz: We usually love to store RFC3339 timestamp instead of Unix timestamp, use `util.GetCurrentTime()`

controllers/auth.go
```diff
@@ -973,6 +980,28 @@ func (c *ApiController) Login() {
 			return
 		}
 
+		var application *object.Application
+		if authForm.ClientId == "" {
```
hsluoyz, line 984: Are if and else both needed?

controllers/auth.go
```diff
@@ -999,31 +1028,34 @@ func (c *ApiController) Login() {
 				}
 			}
 
+			if authForm.EnableMfaExpiry {
+				mfaExpiredAt := object.MfaExpiredAt + user.GetId()
+
+				mfaRememberInSeconds := organization.MfaRememberInHours * 3600
```
hsluoyz, line 1032: Can you just set the default value "12" in only one critical path like getOrganization? So many places are hard-coding this value 12

controllers/auth.go
```diff
@@ -999,31 +1028,34 @@ func (c *ApiController) Login() {
...
-		if err != nil {
-			c.ResponseError(err.Error())
-			return
-		}
+			if authForm.EnableMfaExpiry {
+				mfaVerifiedAtKey := object.MfaExpiredAt + user.GetId()
 
-		if application == nil {
-			c.ResponseError(fmt.Sprintf(c.T("auth:The application: %s does not exist"), authForm.Application))
+				mfaRememberInSeconds := organization.MfaRememberInHours * 3600
```
hsluoyz: Don't want to see this code again and again

web/src/OrganizationEditPage.js
```diff
@@ -603,6 +603,16 @@ class OrganizationEditPage extends React.Component {
             />
           </Col>
         </Row>
+        <Row style={{marginTop: "20px"}} >
+          <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 22 : 2}>
+            {Setting.getLabel(i18next.t("application:Mfa expire"), i18next.t("application:Mfa expire - Tooltip"))} :
+          </Col>
+          <Col span={22} >
+            <InputNumber style={{width: "150px"}} value={this.state.organization.mfaRememberInHours === 0 ? 12 : this.state.organization.mfaRememberInHours} min={1} step={1} precision={0} addonAfter="Hours" onChange={value => {
```
hsluoyz: Don't want to see this code again and again

web/src/auth/mfa/MfaVerifyTotpForm.js
```diff
@@ -54,6 +64,14 @@ export const MfaVerifyTotpForm = ({mfaProps, onFinish}) => {
           }}
         />
       </Form.Item>
+      <Form.Item
+        name="enableMfaExpiry"
+        valuePropName="checked"
+      >
+        <Checkbox>
+          {i18next.t("mfa:Remember this device for {hour} hours").replace("{hour}", organization?.mfaRememberInHours || 12)}
```
hsluoyz: Don't want to see this 12 value again

web/src/locales/zh/data.json
```diff
@@ -82,6 +82,8 @@
     "Left": "居左",
     "Logged in successfully": "登录成功",
     "Logged out successfully": "登出成功",
+    "Mfa expire": "Mfa expire",
```
hsluoyz: Zh needs to be translated

controllers/auth.go
dacongda: keep mfa valid should only valid for current device, so we must use session to store

controllers/auth.go
dacongda: if authForm.Passcode != "" and } else if authForm.RecoveryCode != "" both need

controllers/auth.go
dacongda: util.GetCurrentTime() return a string value, if we need to use it to compare with current time, we need to decode it first, so store unix timestamp seems like a better solution, same solution also used in `CheckVerificationCode` of `object/verification`

Attack825:
https://github.com/user-attachments/assets/c92d5432-4684-413f-aab4-9bef52d8b384

controllers/auth.go
hsluoyz: OK

controllers/auth.go
hsluoyz: OK

web/src/auth/mfa/MfaVerifySmsForm.js
```diff
@@ -12,6 +12,13 @@ export const MfaVerifySmsForm = ({mfaProps, application, onFinish, method, user}
   const [dest, setDest] = React.useState("");
   const [form] = Form.useForm();
 
+  const handleFinish = (values) => {
+    onFinish({
+      passcode: values.passcode,
+      enableMfaExpiry: values.enableMfaExpiry,
```
hsluoyz: enableMfaExpiry -> enableMfaRemember everywhere

web/src/OrganizationEditPage.js
```diff
@@ -603,6 +604,16 @@ class OrganizationEditPage extends React.Component {
             />
           </Col>
         </Row>
+        <Row style={{marginTop: "20px"}} >
+          <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 22 : 2}>
+            {Setting.getLabel(i18next.t("application:MFA validity hours"), i18next.t("application:Per-account MFA validity, re-login skips MFA"))} :
```
hsluoyz: "application: MFA remember time"
"application: MFA remember time - Tooltip"

"Per-account MFA validity, re-login skips MFA" is not native English. Improve it with ChatGPT

hsluoyz (changes requested):

object/mfa.go
```diff
@@ -49,6 +50,11 @@ const (
 	RequiredMfa      = "RequiredMfa"
 )
 
+const (
+	DefaultMfaRememberHours = 12
+	SecondsPerHour          = 3600
```
hsluoyz: Remove SecondsPerHour

web/src/OrganizationEditPage.js
```diff
@@ -603,6 +604,16 @@ class OrganizationEditPage extends React.Component {
             />
           </Col>
         </Row>
+        <Row style={{marginTop: "20px"}} >
+          <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 22 : 2}>
+            {Setting.getLabel(i18next.t("application:MFA remember time"), i18next.t("application:Configures the duration that a account is remembered as trusted after a successful MFA login"))} :
+          </Col>
+          <Col span={22} >
+            <InputNumber style={{width: "150px"}} value={this.state.organization.mfaRememberInHours === 0 ? DefaultMfaRememberInHours : this.state.organization.mfaRememberInHours} min={1} step={1} precision={0} addonAfter="Hours" onChange={value => {
```
hsluoyz: Remove all "DefaultMfaRememberInHours" in frontend JS

web/src/OrganizationEditPage.js
```diff
@@ -603,6 +604,16 @@ class OrganizationEditPage extends React.Component {
             />
           </Col>
         </Row>
+        <Row style={{marginTop: "20px"}} >
+          <Col style={{marginTop: "5px"}} span={(Setting.isMobile()) ? 22 : 2}>
+            {Setting.getLabel(i18next.t("application:MFA remember time"), i18next.t("application:Configures the duration that a account is remembered as trusted after a successful MFA login"))} :
```
hsluoyz: Tooltip

controllers/auth.go
```diff
@@ -999,6 +1028,20 @@ func (c *ApiController) Login() {
 				}
 			}
 
+			if authForm.EnableMfaRemember {
+				mfaRememberInSeconds := organization.MfaRememberInHours * object.SecondsPerHour
```
hsluoyz: Remove DefaultMfaRememberHours

web/src/OrganizationListPage.js
```diff
@@ -25,6 +25,7 @@ import PopconfirmModal from "./common/modal/PopconfirmModal";
 class OrganizationListPage extends BaseListPage {
   newOrganization() {
     const randomName = Setting.getRandomName();
+    const DefaultMfaRememberInHours = 12;
```
hsluoyz, line 28: Remove

web/src/OrganizationListPage.js
```diff
@@ -48,6 +49,7 @@ class OrganizationListPage extends BaseListPage {
       enableSoftDeletion: false,
       isProfilePublic: true,
       enableTour: true,
+      mfaRememberInHours: DefaultMfaRememberInHours,
```
hsluoyz, line 52: 12

hsluoyz (changes requested):

PR #3927, opened 2025-07-04
feat: Change the CredManager interface method, only input one salt based on different situations

raiki02:
f


hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/3928

Issue #3901, opened 2025-06-25
Change organization password salt will cause relevant users unable to log in


Vickko:
As is mentioned in the title, if an organization use a password type related to Org password salt（e.g. sha512-salt）and set org password  salt to "123", a user registered at this moment and set his password. After that, Org changes salt to "1234", this user will immediately unable to login.（regardless of what the password type is set to at this time.）

### How to reproduce

1. Create a new organization or randomly select an existing one.
2. Set its password type to sha512-salt, or other Org password salt related type, and set salt to some value
3. Create a user under this Org
4. Try to login the user in step 3, it should be successful.
5. Change Org password salt to another value
6. Try to login the user in step 3, and it’s locked out of the system.
 

Consider that this issue only relates to a few specific password types, I think this is unreasonable. Thus, It should be regarded as inconsistent with expectation rather than a feature.

Vickko:
My solution to this issue is to add a snapshot for "org_password_salt" in the user table in the database, it record’s the salt when the password created. 

This logic is consistent with the “password_type” that was already included in the user table, which allows existing users still able to login after a change of Org’s password type, an extra field of "org_password_salt_snapshot" in the user table can allow existing user still able to login after Org change its password hash as well.

hsluoyz:
This func `IsPasswordCorrect(password string, passwordHash string, userSalt string, organizationSalt string) bool` is strange.

@dacongda do you know why we have both userSalt and organizationSalt in this function?

hsluoyz:
This commit: https://github.com/casdoor/casdoor/commit/8a4311c85ce0ff02803e25545e9d6177300c766e#diff-e3a5733e379e9b811bb5e87ea034f83d4aa2846a3727245b54fea3131111bc5a added org's salt first

Then this commit: https://github.com/casdoor/casdoor/commit/7792f4589da42af0a617d35d6a7d27515f2f221d#diff-58d4e4fec0e2012442cfd045e2cdb775c6aeb362a139cdc95352132edd88419a added user's sault.

These two commits are too far away (2021) and I forgot the logic between them.

PR #3928, opened 2025-07-04
feat: CredManager.GetHashedPassword() only contains one salt arg now

raiki02:
Fix: https://github.com/casdoor/casdoor/issues/3901

### DIFF
--- a/object/organization.go
+++ b/object/organization.go
@@ -222,7 +222,7 @@ func UpdateOrganization(id string, organization *Organization, isGlobalAdmin boo
 	if organization.MasterPassword != "" && organization.MasterPassword != "***" {
 		credManager := cred.GetCredManager(organization.PasswordType)
 		if credManager != nil {
-			hashedPassword := credManager.GetHashedPassword(organization.MasterPassword, "", organization.PasswordSalt)
+			hashedPassword := credManager.GetHashedPassword(organization.MasterPassword, organization.PasswordSalt)
 			organization.MasterPassword = hashedPassword
 		}
 	}
--- a/object/user_cred.go
+++ b/object/user_cred.go
@@ -42,8 +42,9 @@ func (user *User) UpdateUserHash() error {
 func (user *User) UpdateUserPassword(organization *Organization) {
 	credManager := cred.GetCredManager(organization.PasswordType)
 	if credManager != nil {
-		hashedPassword := credManager.GetHashedPassword(user.Password, user.PasswordSalt, organization.PasswordSalt)
+		hashedPassword := credManager.GetHashedPassword(user.Password, organization.PasswordSalt)
 		user.Password = hashedPassword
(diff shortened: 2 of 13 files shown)


hsluoyz:
@dacongda plz review

object/check.go
```diff
@@ -21,11 +21,12 @@ import (
 	"time"
 	"unicode"
 
+	goldap "github.com/go-ldap/ldap/v3"
```
dacongda: don't change import list

object/user_cred.go
```diff
@@ -14,7 +14,9 @@
 
 package object
 
-import "github.com/casdoor/casdoor/cred"
+import (
```
dacongda: don't change import list

object/organization.go
```diff
@@ -18,12 +18,13 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/xorm-io/builder"
```
dacongda: don't change import list

object/check.go
```diff
@@ -252,12 +253,12 @@ func CheckPassword(user *User, password string, lang string, options ...bool) er
...
 	if credManager != nil {
 		if organization.MasterPassword != "" {
-			if password == organization.MasterPassword || credManager.IsPasswordCorrect(password, organization.MasterPassword, "", organization.PasswordSalt) {
+			if password == organization.MasterPassword || credManager.IsPasswordCorrect(password, organization.MasterPassword, organization.PasswordSalt) {
 				return resetUserSigninErrorTimes(user)
 			}
 		}
 
-		if credManager.IsPasswordCorrect(password, user.Password, user.PasswordSalt, organization.PasswordSalt) {
+		if credManager.IsPasswordCorrect(password, user.Password, user.PasswordSalt) {
```
dacongda: It seems that user.Password is empty in old version, have you tested?

dacongda (changes requested):

object/check.go
raiki02: I think the user.Password isn't empty but the user.PasswordSalt sometimes is. When checking password correct, the former function uses the organzation.PasswordSalt or user.PasswordSalt to encrypt the plain password the user input in the requet and compares with user.Password which is already encrypted.

dacongda: my fault, I type the wrong text, It should be `It seems that user.PasswordSalt is empty in old version`, I think we can check Organization.PasswordSalt first, if wrong, we can check user.PasswordSalt as fallback

raiki02: I see. that's to say, we use organization.PasswordSalt fisrt to encrypt the plain password to compare with the user.Password, if wrong, use the user.PasswordSalt to retry.

dacongda: yes

raiki02: I will remain the two salts input, and edit the inner logic.

PR #3934, opened 2025-07-06
feat: add phone number validation for user updates and remove random number generation

RooobinYe:
To ensure phone number initialization, there are two options:
1. Generate a phone number that complies with the area code
2. Let the new phone number default to empty

I chose the second option.


object/check.go
```diff
@@ -582,6 +582,8 @@ func CheckUpdateUser(oldUser, user *User, lang string) string {
 	if oldUser.Phone != user.Phone {
 		if HasUserByField(user.Owner, "phone", user.Phone) {
 			return i18n.Translate(lang, "check:Phone already exists")
+		} else if !util.IsPhoneValid(user.Phone, user.CountryCode) && user.Phone != "" {
```
hsluoyz: What if CountryCode is empty or Phone is empty?

web/src/Setting.js
```diff
@@ -1442,8 +1442,8 @@ export function getRandomName() {
   return Math.random().toString(36).slice(-6);
 }
 
-export function getRandomNumber() {
-  return Math.random().toString(10).slice(-11);
+export function getInitNumber() {
```
hsluoyz: Why make this change?

web/src/Setting.js
```diff
@@ -1442,8 +1442,8 @@ export function getRandomName() {
...
+  return "";
```
hsluoyz: Why keep getInitNumber func?

hsluoyz (changes requested):
I don't know why make this PR?

RooobinYe:
@hsluoyz  
Now the same country code always generates the same default phone number. This leads to a problem: if an administrator creates a user for the first time and doesn’t change that default phone number, they won’t be able to create a second user. The user-creation logic first saves a template user and then modifies it, but every template user with the same area code gets the identical default phone number, so the second attempt triggers a “duplicate phone number” error.

Do you think how to solve this problem?

图片

RooobinYe:
Remove `IsPhoneValid`, which has the same function with `GetE164Number`.

hsluoyz:
This PR has become too complicated to be understood

Issue #3902, opened 2025-06-25
Change Organization’s password type won’t refresh password and type for existing users


Vickko:
If an organization use a password type A at first, and several users registered and set their password. After that, organization change password type to another type B. Looking into relevant code logic and the user table in the database at this point, we can know that these users are still using old password type and old password hash to verify password. Only if these user change their password, the type and hash will update and align with its Org’s.

This behavior allows existing users still able to login after a change of Org’s password type, but user password policy should update as soon as possible with changes in Org’s password policy to ensure security. Considering that Casdoor uses plain text as default password type for Org, I think it is more necessary to deal with this issue.

This issue is similar to Issue #3278 to some extent, though I think the description there may not be 100% precise, but the core concern is the same.

Vickko:
My solution to this issue is to add a logic that, check if the password type of a user is "expired" when they successfully log in. If so, It means the plain password this user provide is its true password. At the same time we let him in as usual, we can use this plain text and the password type of the Org at this time to rehash the password and store it for the user. 

In this way, we can ensure that user’s password hash can be updated much sooner than the logic we have now. And more importantly, comparing to asking users to change their password on their own, this way makes it seamless to deal with the problem.

Vickko:
@hsluoyz hi, do you have any idea about this issue? can I PR to add this feature?

hsluoyz:
@Vickko your method only works for the plain password. Because you usually cannot decrypt the password with the old algorithm A and then re-encrypt it with the new algorithm B. If we are talking about only plain password, I think a much easier way is just that you, the admin changes the org password type first, then add your users. Therefore I don't think your proposal is very useful.

Vickko:
@hsluoyz no you get me wrong, it is not about decrypt password from old hash, it is when a user tries to log in again, we have his **unencrypted password text**, then we can hash this **unencrypted password** again using the new algorithm during this login process, just like how we hash a password in normal register process. Isn't it?

So this solution is **compatible with all algorithm migration situation**, not just about plain password type to other algorithm

hsluoyz:
@dacongda what do you think?

dacongda:
seems to be a good idea

hsluoyz:
@Vickko can you make a PR?

Vickko:
> [@Vickko](https://github.com/Vickko) can you make a PR?

sure, I will work on it

PR #3937, opened 2025-07-07
feat: Supports smooth migration of password hash

Vickko:
Fix: #3902 

and by the way, I notice function signature of `credManager.IsPasswordCorrect(password, user.Password, user.PasswordSalt, organization.PasswordSalt)` is changed in #3936, so these two PRs will potentially cause a conflict.
I will wait for its merge and resolve the conflict, or other further advice about this situation.


object/check.go
```diff
@@ -250,21 +250,33 @@ func CheckPassword(user *User, password string, lang string, options ...bool) er
...
-	} else {
-		return fmt.Errorf(i18n.Translate(lang, "check:unsupported password type: %s"), organization.PasswordType)
 	}
+
+	isOutdated := passwordType != organization.PasswordType
+
+	if isOutdated {
+		// password is correct, update user's password to new password type
+		user.Password = password
+		user.UpdateUserPassword(organization)
```
dacongda: PasswordSalt should be updated too

Issue #3935, opened 2025-07-07
OAuth 2.0 Token Introspection endpoint returns incorrect client_id


even425:
## Issue Description

The OAuth 2.0 Token Introspection endpoint (`/api/login/oauth/introspect`) is currently returning the client_id from the request authentication instead of the client_id associated with the token being introspected.

## Current Behavior

In `controllers/token.go:IntrospectToken()`, the `client_id` field in the `IntrospectionResponse` is set to the client_id from the Basic Auth request:

```go
introspectionResponse = object.IntrospectionResponse{
    Active:    true,
    Scope:     jwtToken.Scope,
    ClientId:  clientId,  // This is from Basic Auth, not from the token
    Username:  jwtToken.Name,
    // ...
}
```

## Expected Behavior

According to RFC 7662 (OAuth 2.0 Token Introspection), the `client_id` field should represent:
> "The client identifier for the OAuth 2.0 client that requested the token."

This means it should be the client_id of the application that originally generated the token, not the client_id of the application making the introspection request.

## Problem Analysis

1. **Token Structure**: The `Token` struct contains an `Application` field that stores the application name that generated the token
2. **Missing Logic**: The current implementation doesn't retrieve the original application's client_id from the token
3. **Security Concern**: Returning the requesting client_id may leak information about the calling service
4. **RFC Compliance**: This violates the RFC 7662 specification

## Proposed Solution

The `IntrospectToken` method should be modified to:

1. Retrieve the original application from the token's `Application` field
2. Get the application's `ClientId` 
3. Return that `ClientId` in the introspection response

Example fix:
```go
if token != nil {
    // Get the original application that generated the token
    tokenApplication, err := getApplication(token.Owner, token.Application)
    if err != nil {
        c.ResponseTokenError(err.Error())
        return
    }
    
    introspectionResponse = object.IntrospectionResponse{
        Active:    true,
        Scope:     jwtToken.Scope,
        ClientId:  tokenApplication.ClientId,  // Use the original application's client_id
        Username:  jwtToken.Name,
        // ...
    }
}
```

## Impact

This issue affects:
- OAuth 2.0 compliance
- Token introspection accuracy
- Security and privacy of client information
- Interoperability with other OAuth 2.0 implementations

## Steps to Reproduce

1. Create an OAuth 2.0 application (Application A)
2. Generate a token using Application A
3. Create another OAuth 2.0 application (Application B)
4. Use Application B's credentials to call the introspection endpoint for Application A's token
[...]

PR #3948, opened 2025-07-09
fix: token introspect endpoint returns correct client_id as rfc7662 defined

raiki02:
Fix: https://github.com/casdoor/casdoor/issues/3935

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -16,6 +16,7 @@ package controllers
 
 import (
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/beego/beego/utils/pagination"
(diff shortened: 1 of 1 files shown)


controllers/token.go
```diff
@@ -460,7 +461,13 @@ func (c *ApiController) IntrospectToken() {
 	}
 
 	if token != nil {
+		application, err = object.GetApplication(fmt.Sprintf("%s/%s", token.Owner, token.Application))
+		if err != nil {
+			c.ResponseTokenError(err.Error())
+			return
+		}
```
hsluoyz, line 468: Check application == nil, use existing error string

PR #3965, opened 2025-07-16
feat: fix tour bug about orgIsTourVisible settings

Attack825:


Attack825:
https://github.com/user-attachments/assets/e0318fe5-8c6f-41ea-a2aa-bbabe3ef0f66

PR #3975, opened 2025-07-20
feat: optimize the proxy create method

xtutu:
It support url with username & password:
socks5Proxy = "socks5://user:password@8.8.8.8:10011"


hsluoyz:
@xtutu create issue first. Your code without the issue justification and back compat will just break the existing system

Issue #3670, opened 2025-03-18
Auto Sign-Up Flow: User Not Automatically Logged In & Magic Link Clarification


4328031237927392:
Description:
When attempting a sign-up flow with email verification, the behavior does not match the expected seamless experience.

Expected Behavior:
• If a user does not already exist, the user uses the sign up page to register.
• Example, the system is configured to send a verification code via email without password.
• Upon confirming the code, the user is immediately logged in without additional steps.

Actual Behavior:
• On the sign-in page, entering an unregistered email returns an error (“user does not exist”).
• On the sign-up page (configured to not require a password), the account is created, but the user is not logged in automatically. Instead, a success page shows a link to sign in. Following this link, the user undergoes another email verification loop rather than being logged in directly.

Additional Note on Magic Link Behavior:
The functionality is very similar to a "Magic Link" experience. The current implementation includes an invitation code and an email verification code feature - the latter suffers from issue above. However, a Magic Link is slightly different and should allow the user to click a link in the email, which directly exchanges a stored secret token in the link for a valid session token, bypassing any manual verification code input or additional steps. Clarification is needed on whether this is the same or a separate feature request.

Steps to Reproduce:
1. Attempt to sign in with an email address that isn’t registered.
2. Notice the error message (“user does not exist”).
3. Alternatively, sign up using the sign-up page (with password not required).
4. Complete the email verification.
5. Observe that after verification, the user is not logged in but instead prompted to sign in again.

Expected Outcome:
Users should be able to complete the auto sign-up flow and be logged in immediately upon email code verification (or automatic Magic Link token exchange, if applicable), eliminating the need for a separate sign-in step.

Please let me know if you need any additional info!

buzaslan129:
+ 1

PR #3983, opened 2025-07-21
feat: fix auto sign-in flow on result page

Attack825:
Fix: https://github.com/casdoor/casdoor/issues/3670

https://github.com/user-attachments/assets/9b41b0ed-b425-4d21-a769-a8af8b9fb816

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -286,8 +286,7 @@ func (c *ApiController) Signup() {
 		}
 	}
 
-	if application.HasPromptPage() && user.Type == "normal-user" {
-		// The prompt page needs the user to be signed in
+	if user.Type == "normal-user" {
 		c.SetSessionUsername(user.GetId())
 	}
 
--- a/web/src/auth/ResultPage.js
+++ b/web/src/auth/ResultPage.js
@@ -18,6 +18,7 @@ import i18next from "i18next";
 import {authConfig} from "./Auth";
 import * as ApplicationBackend from "../backend/ApplicationBackend";
 import * as Setting from "../Setting";
+import * as AuthBackend from "./AuthBackend";
 
 class ResultPage extends React.Component {
   constructor(props) {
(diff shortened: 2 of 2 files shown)


hsluoyz:
@dacongda

controllers/account.go
```diff
@@ -286,10 +286,7 @@ func (c *ApiController) Signup() {
 		}
 	}
 
-	if application.HasPromptPage() && user.Type == "normal-user" {
```
dacongda, line 289: Why remove this if statement, how about the application has prompt page? have you test this case?

Attack825: > The prompt page needs the user to be signed in

The auto sign-in flow hopes that after email registration, users can log in immediately without any additional steps. So I set all new registered users with `c.SetSessionUsername(user.GetId())`. If the If statement is deleted, it will not affect the prompt page. Because it already requires `c.SetSessionUsername(user.GetId())`.

Attack825: Original logic: Set the login status (session user name) only when the application has a prompt page and the user is normal-user.

Modified logic: all normal-users will set the login status regardless of whether the application has a prompt page or not (so that all newly registered normal-users will be in the login status).

Impact:
-prompt page needs to be logged in, and normal-user can access it normally after modification;
-normal-user without prompt page also automatically logs in, which is in line with expectations;
-paid-user is not affected (its logic has not been modified), and still jumps to the payment page without logging in directly, which meets the requirements.

Attack825: Original logic: Set the login status (session user name) only when the application has a prompt page and the user is normal-user.

Modified logic: all normal-users will set the login status regardless of whether the application has a prompt page or not (so that all newly registered normal-users will be in the login status).

Impact:
- prompt page needs to be logged in, and normal-user can access it normally after modification;
- normal-user without prompt page also automatically logs in, which is in line with expectations;
- paid-user is not affected (its logic has not been modified), and still jumps to the payment page without logging in directly, which meets the requirements.

Attack825:
https://github.com/user-attachments/assets/2b90db8a-5e34-4c3c-9a93-45dc66fe933d

Issue #3977, opened 2025-07-20
Default value Auto-Sign in


HRNasr:
Is it possible to set the default value of َ (Auto Sign in) on the login page so that it is not always True?  and   It should be configurable.

hsluoyz:
@HRNasr "Auto Sign in" has no usage

hsluoyz:
@dacongda

dacongda:
we will try to add it later

PR #3987, opened 2025-07-22
feat: can set default value for "Auto sign in" in application edit page

dacongda:
Fix: https://github.com/casdoor/casdoor/issues/3977


dacongda:
demo

https://github.com/user-attachments/assets/723f81a0-0829-4a6a-a7bf-77be99abc4d9

PR #3995, opened 2025-07-24
fix: improve OAuth form_post code to id_token

RooobinYe:
According to [Final: OAuth 2.0 Form Post Response Mode](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html), I changed code to id_token to fix this bug.


web/src/auth/AuthCallback.js
```diff
@@ -35,16 +35,17 @@ class AuthCallback extends React.Component {
     };
   }
 
-  submitFormPost(redirectUri, code, state) {
+  submitFormPost(redirectUri, idToken, state) {
```
dacongda: id token is a example, response_type not always equal to `id_token`, it can be `"code", "code id_token", "id_token", "id_token token", "code id_token token"`, You need to return the corresponding value based on response_type

PR #4000, opened 2025-07-25
feat: fix issue that resource provider.PathPrefix is not used in GetDirectResources() API

IsAurora6:
I found that cloud storage (such as alicloud oss) considers the incoming fullFilePath and storage provider.PathPrefix to be concatenated as a path when uploading resources, as follows:
image
However, when querying (GetResources), it does not consider concatenating provider.PathPrefix, but only the incoming fullFilePath. This has led to the issue of “casibase being unable to retrieve any files when calling getresou


hsluoyz:
@dacongda plz review

PR #4020, opened 2025-07-30
feat: implement root DSE handling and schema query in LDAP server

GZYZhy:
Hello, I checked existing issues but didn't find relevant solutions for my needs, hence this enhancement proposal for LDAP support.

Here's what this PR changes:

### 1. Partial Root DSE Query Support  
The Root DSE is a virtual entry defined in the LDAP protocol (baseDN=""), serving to help clients quickly understand directory service capabilities and basic configuration. Supporting it helps Casdoor's LDAP server become more standards-compliant. This PR provides basic Root DSE query support.

**Feature examples:**  
- **Query namingContexts (organization list):**  
  ```bash
  ldapsearch -x -H ldap://<casdoor-host>:389 -D "cn=admin,ou=built-in" -w <passwd> -b "" -s base "(objectClass=*)" namingContexts
  ```  
  **Expected:** Returns an entry containing the `namingContexts` attribute listing all visible organization DNs (e.g., `ou=built-in`). Non-admin bindings will only return the current organization.  

- **Query subschemaSubentry:**  
  ```bash
  ldapsearch -x -H ldap://<casdoor-host>:389 -D "cn=admin,ou=built-in" -w <passwd> -b "" -s base "(objectClass=*)" subschemaSubentry
  ```  
  **Expected:** Returns `subschemaSubentry: cn=Subschema`.

### 2. Schema Query Support  
Some strict LDAP clients request objectClasses during connection to validate schema support. Previously Casdoor didn't support this, potentially causing connection failures. This PR adds objectClasses query capability.

**Feature example:**  
- **Query objectClasses:**  
  ```bash
  ldapsearch -x -H ldap://<casdoor-host>:389 -D "cn=admin,ou=built-in" -w <passwd> -b "cn=Subschema" -s base "(objectClass=*)" objectClasses
  ```  
  **Expected:** Returns standard definitions of posixAccount and posixGroup (e.g., `( 1.3.6.1.1.1.2.0 NAME 'posixAccount' ... )`).

---

This PR exclusively adds new functionality. Conditional checks (modified L163-167) ensure new logic only triggers under correct circumstances.I tested it and it doesn’t break existing functionality.


hsluoyz:
@dacongda plz review

ldap/server.go
```diff
@@ -159,6 +160,11 @@ func handleSearch(w ldap.ResponseWriter, m *ldap.Message) {
 	default:
 	}
 
+	if strings.EqualFold(r.FilterString(), "(objectClass=*)") && (string(r.BaseObject()) == "" || strings.EqualFold(string(r.BaseObject()), "cn=Subschema")) {
```
dacongda, line 162: why `cn=Subschema` is hardcoded, is it a part of standard?

ldap/server.go
```diff
@@ -229,6 +235,69 @@ func handleSearch(w ldap.ResponseWriter, m *ldap.Message) {
...
+			w.Write(e)
+		} else if strings.EqualFold(firstAttr, "subschemaSubentry") {
+			e := ldap.NewSearchResultEntry(string(r.BaseObject()))
+			e.AddAttribute("subschemaSubentry", message.AttributeValue("cn=Subschema"))
+			w.Write(e)
+		}
+	} else if strings.EqualFold(firstAttr, "objectclasses") && strings.EqualFold(string(r.BaseObject()), "cn=Subschema") {
+		e := ldap.NewSearchResultEntry(string(r.BaseObject()))
+		e.AddAttribute("objectClasses", []message.AttributeValue{
+			"( 1.3.6.1.1.1.2.0 NAME 'posixAccount' DESC 'Abstraction of an account with POSIX attributes' SUP top AUXILIARY MUST ( cn $ uid $ uidNumber $ gidNumber $ homeDirectory ) MAY ( userPassword $ loginShell $ gecos $ description ) )",
```
dacongda, line 268: Why AttributeValue must be this.

ldap/server.go
```diff
@@ -229,6 +235,69 @@ func handleSearch(w ldap.ResponseWriter, m *ldap.Message) {
...
+			"( 1.3.6.1.1.1.2.2 NAME 'posixGroup' DESC 'Abstraction of a group of accounts' SUP top STRUCTURAL MUST ( cn $ gidNumber ) MAY ( userPassword $ memberUid $ description ) )",
+		}...)
+		w.Write(e)
+	}
+
+	w.Write(res)
+}
+
+func GetFilteredOrganizations(m *ldap.Message) ([]*object.Organization, int) {
```
dacongda: put this func into `ldap/util.go` like `GetFilteredGroups` and `GetFilteredUser`

ldap/server.go
GZYZhy: Yes, as far as I know, the RFC 4512 specification states that subschema entries should be retrieved through a search operation with a base object scope, combined with the filter. This is a hard requirement.

dacongda: ok

ldap/server.go
GZYZhy: Because the value of the objectClasses attribute must comply with the object class description syntax defined in RFC 4512.

The attributes I've listed here are standard ones, and my tests show that they are supported by Casdoor. In fact, Casdoor supports more attributes than those listed here. For example, we can add "mail" to the MAY section to indicate that clients can request users' email addresses. However, I'm somewhat worried that this might cause unnecessary troubles.

Nevertheless, regardless of whether attributes like "mail" are listed here, clients can still query them. The purpose here is just for definition. Some LDAP clients (such as the one I'm using) will request this definition.

ldap/server.go
GZYZhy: OK, I will do this.

ldap/server.go
dacongda: ok

Issue #4028, opened 2025-08-02
[ui] show placeholder QR code with loading instead of "Loading" text in QR code login page


hsluoyz:
It's better to first display a QR code with a random string, and overlay it with a loading effect on top.

See:

<img width="1752" height="918" alt="Image" src="https://github.com/user-attachments/assets/cb85d528-2198-469d-9fe7-fbab90edefdb" />

PR #4031, opened 2025-08-02
feat: show placeholder QR code with loading instead of "Loading" text in QR code login page

dacongda:
Fix: https://github.com/casdoor/casdoor/issues/4028

### DIFF
--- a/idp/wechat.go
+++ b/idp/wechat.go
@@ -17,7 +17,6 @@ package idp
 import (
 	"bytes"
 	"crypto/sha1"
-	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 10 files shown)


dacongda:
demo: 


https://github.com/user-attachments/assets/e165e9f9-4c9d-4057-a732-15742ca165e0

idp/wechat.go
```diff
@@ -29,7 +28,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/skip2/go-qrcode"
```
hsluoyz, line 32: Update go.mod and go.sum

hsluoyz (changes requested):

PR #4040, opened 2025-08-06
Add mailbox to WoCom Enterprise

Rocky-Jinxing:
Enterprise WeChat scan code login cannot obtain the user's email address, only the UserID. By manually setting the email suffix, the UserID + email suffix is combined into the user's email address.


hsluoyz:
@Rocky-Jinxing 

You should create issue to discuss first. Closed here

PR #4046, opened 2025-08-08
feat: Add enterprise WeCom provider and generate email address through Userid #3801

### LINKED ISSUE #3801: Wecom scan code login will automatically connect you to existing users of casdoor.
The application does not open registration, and uses Wecom to scan the code to log in. If there is already a user with the same "name" in casdoor ("name" is not case-sensitive), it is hoped that Wecom can be automatically associated with the existing user.

If this can be achieved, for internal users of enterprises using Wecom, there is no need to manually bind Wecom.

VenusHou:
## What this PR does

- Generate the user's email address based on the UserID of the enterprise maintenance user.

## Related issue

Fixes #3801

## How to test

- Users can scan the QR code to log in and bind to existing users

https://github.com/user-attachments/assets/780fcf08-f2fe-4881-89e7-12dcf6a552ef


hsluoyz:
@VenusHou you should not create a new GitHub account to do spamming after you have sent three spam PRs:

image

PR #4068, opened 2025-08-12
revert-4067-fix-provider-captcha

nejiklol:
fix: allow CaptchaType="none" without provider type match in verification.go

### DIFF
--- a/controllers/verification.go
+++ b/controllers/verification.go
@@ -147,11 +147,6 @@ func (c *ApiController) SendVerificationCode() {
 	}
 
 	if provider != nil {
-		if vform.CaptchaType != provider.Type {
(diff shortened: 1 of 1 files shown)


hsluoyz:
@dacongda plz review

controllers/verification.go
```diff
@@ -147,11 +147,6 @@ func (c *ApiController) SendVerificationCode() {
 	}
 
 	if provider != nil {
-		if vform.CaptchaType != provider.Type {
```
hsluoyz, line 150: Why remove this check?

hsluoyz (changes requested):

controllers/verification.go
nejiklol: mb, thought there was an error, now i see that i was wrong

PR #4071, opened 2025-08-13
feat: add new language: "az"

KhanbalaRashidov:


hsluoyz:
@KhanbalaRashidov your code is not correct. There will be other contributor to work on this

KhanbalaRashidov:
> your code is not correct. There will be other contributor to work on this

Where is the wrong part?

Issue #3846, opened 2025-05-28
A user exists in two entity groups at the same time


Rocky-Jinxing:
The admin in the demo site exists in two entity groups. I tested that ordinary users can also exist in two entity groups.
https://casdoor.org/zh/docs/organization/organization-tree
Does not match the description in the document
![Image](https://github.com/user-attachments/assets/e9271ad5-4659-4473-82b3-65ca17d60f45)

![Image](https://github.com/user-attachments/assets/3b53b53f-fce9-45d2-b8dd-29aa50a85439)

hsluoyz:
@Rocky-Jinxing can you make a PR?

PR #4084, opened 2025-08-15
feat: fix issue that a user can belong to two physical groups at the same time

RooobinYe:
Fix: https://github.com/casdoor/casdoor/issues/3846


hsluoyz:
@Rocky-Jinxing is this PR what you want?

PR #4091, opened 2025-08-17
feat: fix bug that SelfLoginButton will re-render when username field updates

dacongda:
fix: fix SelfLoginButton will re-render when username field update


dacongda:
demo

https://github.com/user-attachments/assets/ca7e9a6d-9205-49a0-9e00-570196ba1ac5

PR #4105, opened 2025-08-20
feat: support inserting user's field to SAML attribute

dacongda:
feat: support insert user's field to saml attribute

### DIFF
--- a/object/saml_idp.go
+++ b/object/saml_idp.go
@@ -122,21 +122,11 @@ func NewSamlResponse(application *Application, user *User, host string, certific
 
 		valueList := []string{item.Value}
 		if strings.Contains(item.Value, "$user.roles") {
-			var roles []string
-			for _, userRole := range user.Roles {
-				roles = append(roles, userRole.Name)
-			}
-
-			valueList = replaceSamlAttributeValuesWithList("$user.roles", roles, valueList)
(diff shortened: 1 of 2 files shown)


dacongda:
for example:

`acs:ram::1879818006829152:role/$user.roles,acs:ram::1879818006829152:saml-provider/testa`
will generate 
```xml
      <saml:Attribute Name="https://www.aliyun.com/SAML-Role/Attributes/Role"
        NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified">
        <saml:AttributeValue xmlns:xs="http://www.w3.org/2001/XMLSchema"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="xs:string">
          acs:ram::1879818006829152:role/role1,acs:ram::1879818006829152:saml-provider/testa</saml:AttributeValue>
        <saml:AttributeValue xmlns:xs="http://www.w3.org/2001/XMLSchema"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="xs:string">
          acs:ram::1879818006829152:role/role2,acs:ram::1879818006829152:saml-provider/testa</saml:AttributeValue>
      </saml:Attribute>
```

Issue #4092, opened 2025-08-17
Sessions cannot be deleted across all applications in Casdoor


hamidrabedi:
Currently, Casdoor’s DeleteSession only deletes a session for a specific (owner, name, application) combination. This causes problems in setups where:

- Single sign-on (SSO) is enabled across multiple applications.
- A user logs in once and sessions are shared across apps.
- When the user logs out, Casdoor only deletes the session for one application, leaving other sessions active.

My working code:
```go
func DeleteSession(id string, onlyUseOrg bool) (bool, error) {
	owner, name, application := util.GetOwnerAndNameAndOtherFromId(id)
	session, err := GetSingleSession(id)

	if owner == CasdoorOrganization && application == CasdoorApplication {
		if err != nil {
			return false, err
		}
	}

	var affected int64
	if onlyUseOrg {
		// Get all sessions for this organization and user (without application filter)
		sessions := []*Session{}
		err = ormer.Engine.Where("owner = ? AND name = ?", owner, name).Find(&sessions)
		if err != nil {
			return false, err
		}

		// Delete all session IDs from beego session store
		for _, sess := range sessions {
			if sess.SessionId != nil {
				DeleteBeegoSession(sess.SessionId)
			}
		}

		// Delete all sessions from database
		affected, err = ormer.Engine.Where("owner = ? AND name = ?", owner, name).Delete(&Session{})
		if err != nil {
			return false, err
		}
	} else {
		affected, err = ormer.Engine.ID(core.PK{owner, name, application}).Delete(&Session{})
		if err != nil {
			return false, err
		}

		if session != nil {
			DeleteBeegoSession(session.SessionId)
		}
	}

	return affected != 0, nil
}

```

hsluoyz:
@hamidrabedi can you create PR?

hamidrabedi:
@hsluoyz can we continue working on this one?

another related issue that could be solved with this: #1281 

handling session deletion and tokens invalidation with some sort of admin customization would be really good to have.

hsluoyz:
@hamidrabedi Single Sign-Out (SSO Logout) has been supported, see docs: https://casdoor.org/docs/how-to-connect/single-sign-out

PR #4126, opened 2025-08-26
feat: make DeleteSession function more dynamic with flexible parameters

hamidrabedi:


hamidrabedi:
related to #4092

hsluoyz:
@dacongda plz review

object/session.go
```diff
@@ -137,20 +137,65 @@ func AddSession(session *Session) (bool, error) {
...
+		session, err := GetSingleSession(util.GetSessionId(org, user, application))
 		if err != nil {
 			return false, err
 		}
 
 		if session != nil {
 			DeleteBeegoSession(session.SessionId)
 		}
+	} else if user != "" {
+		query = query.Where("name = ?", user)
```
dacongda, line 159: since user.name can be same in different org, when user param is not empty, org param should not be empty too.

object/session.go
```diff
@@ -137,20 +137,65 @@ func AddSession(session *Session) (bool, error) {
...
+		err := query.Find(&sessions)
+		if err != nil {
+			return false, err
+		}
+
+		for _, session := range sessions {
+			DeleteBeegoSession(session.SessionId)
+		}
+	} else {
+		sessions := []*Session{}
```
dacongda, line 185: what will happen if org, user and application all equal to empty

object/session.go
```diff
@@ -137,20 +137,65 @@ func AddSession(session *Session) (bool, error) {
...
+		err := query.Find(&sessions)
+		if err != nil {
+			return false, err
+		}
+
+		for _, session := range sessions {
+			if session.Application == CasdoorApplication {
+				DeleteBeegoSession(session.SessionId)
```
dacongda, line 193: why only `session.Application == CasdoorApplication` that we can deleteDeegoSession

hsluoyz:
Closed because I'm worried with this PR after seeing dacongda's comments. PR authors usually don't take care of the backward compatibility. They only care about themselves, while we have to consider all people. That's the difference from angles.

Issue #4127, opened 2025-08-26
Issues with Casbin RBAC Model and Adapter Limitations


Alicenap:
### Dear Casbin/Casdoor Support Team,
(Using the docker image casbin/casdoor:v2.37.0 with postgreSQL)

I have implemented an RBAC with multiple roles model using Casbin, based on the standard configuration options. The policy_definition section of my model is defined as follows:

**[policy_definition]
p = sub, obj, act**

<img width="1640" height="1116" alt="Image" src="https://github.com/user-attachments/assets/3cb39e7d-cba4-4813-9cf8-64747b2278c8" />

Within my organization (my_org), I created two users: Alice and Bob. Alice is assigned to the reader group, while Bob is assigned to both the reader and writer groups. I have defined read_permission for the reader group and write_permission for the writer group.

When I inspect the enforcer (my_enforcer), it correctly displays the g entries but does not show the p entries. 

<img width="1640" height="1116" alt="Image" src="https://github.com/user-attachments/assets/1c0dc46d-c220-4153-b734-46cb0687deaf" />

However, when I extend the model by adding three additional (unused) arguments to the policy_definition...

<img width="1640" height="1116" alt="Image" src="https://github.com/user-attachments/assets/e11b6b79-d755-4b9b-bf38-450d59bd7d72" />

the enforcer begins to display both g and p entries. 

<img width="1640" height="1116" alt="Image" src="https://github.com/user-attachments/assets/f6ef7e4e-32ca-456e-a057-083d335bcddb" />

At this point, attempting to assign Alice to the writer group results in the following error:

**"Failed to save: the maximum policy_definition field number cannot exceed 5, got 6"**

https://github.com/user-attachments/assets/d80584d3-0d39-4972-9321-9b0efe39d00b

This issue also causes errors in other parts of the system, such as when creating new permissions.

According to the Casbin documentation  [Policy Storage](https://casbin.org/docs/policy-storage)

_v0-v5: The column names have no specific meaning and correspond to the values in the policy CSV from left to right. The number of columns depends on how many you define yourself. In theory, there can be an infinite number of columns, but generally only 6 columns are implemented in the adapter. If this is not enough for you, please submit an issue to the corresponding adapter repository._

This suggests that the number of columns in policy_definition should be flexible. However, I am **also** encountering this limitation when using the casbin_rule table via the casbin-pg-adapter library in my client application.

**Could you please clarify:**

[...]

PR #4130, opened 2025-08-29
feat: change builtInMaxFields to 6 in Casbin policy length

Attack825:
Fix: https://github.com/casdoor/casdoor/issues/4127

### DIFF
--- a/object/permission.go
+++ b/object/permission.go
@@ -49,7 +49,7 @@ type Permission struct {
 	State       string `xorm:"varchar(100)" json:"state"`
 }
 
-const builtInAvailableField = 5 // Casdoor built-in adapter, use V5 to filter permission, so has 5 available field
+const builtInMaxFields = 6 // Casdoor built-in adapter, use V5 to filter permission, so has 6 max field
 
 func GetPermissionCount(owner, field, value string) (int64, error) {
 	session := GetSession(owner, -1, -1, field, value, "", "")
--- a/object/permission_enforcer.go
+++ b/object/permission_enforcer.go
@@ -410,15 +410,23 @@ m = g(r.sub, p.sub) && r.obj == p.obj && r.act == p.act`
 		policyDefinition := strings.Split(cfg.String("policy_definition::p"), ",")
 
 		fieldsNum := len(policyDefinition)
-		if fieldsNum > builtInAvailableField {
-			return nil, fmt.Errorf("the maximum policy_definition field number cannot exceed %d, got %d", builtInAvailableField, fieldsNum)
+		if fieldsNum > builtInMaxFields {
+			return nil, fmt.Errorf("the maximum policy_definition field number cannot exceed %d, got %d", builtInMaxFields, fieldsNum)
 		}
 
 		// filled empty field with "" and V5 with "permissionId"
-		for i := builtInAvailableField - fieldsNum; i > 0; i-- {
-			policyDefinition = append(policyDefinition, "")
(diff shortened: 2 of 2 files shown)


eu-erwin:
just for my curiosity,

shouldn't we also update this line, so that the V5 column preserved? the casbin library expects exact number of column, if the last column is not empty. by changing the `builtInColumnNumber`, the policy definition in this code block has more than 6 column now, which was not a problem in my usecase. we needed this, because the casbin library filtered the "p" definitions and left only the "g" in the "/api/get-policies" Endpoint. To make sure that the permission-id didn't accidentally updated, i think the UI should also disabled this V5 Column in the editor (Enforcer). so we don't have data left in the adapter table in case the permission were deleted.

`https://github.com/casdoor/casdoor/blob/a50e89bf90673cb832c8d1b6e6c1201a31c64d56/object/permission_enforcer.go#L420C4-L421C62`

```
		// load [policy_definition]
		policyDefinition := strings.Split(cfg.String("policy_definition::p"), ",")

		fieldsNum := len(policyDefinition)
		if fieldsNum > builtInAvailableField {
			return nil, fmt.Errorf("the maximum policy_definition field number cannot exceed %d, got %d", builtInAvailableField, fieldsNum)
		}

		// filled empty field with "" and V5 with "permissionId"
		for i := builtInAvailableField - fieldsNum; i > 0; i-- {
			policyDefinition = append(policyDefinition, "")
		}
		policyDefinition = append(policyDefinition, "permissionId")
```

Attack825:
> just for my curiosity,
> 
> shouldn't we also update this line, so that the V5 column preserved? the casbin library expects exact number of column, if the last column is not empty. by changing the `builtInColumnNumber`, the policy definition in this code block has more than 6 column now, which was not a problem in my usecase. we needed this, because the casbin library filtered the "p" definitions and left only the "g" in the "/api/get-policies" Endpoint. To make sure that the permission-id didn't accidentally updated, i think the UI should also disabled this V5 Column in the editor (Enforcer). so we don't have data left in the adapter table in case the permission were deleted.
> 
> `https://github.com/casdoor/casdoor/blob/a50e89bf90673cb832c8d1b6e6c1201a31c64d56/object/permission_enforcer.go#L420C4-L421C62`
> 
> ```
> 		// load [policy_definition]
> 		policyDefinition := strings.Split(cfg.String("policy_definition::p"), ",")
> 
> 		fieldsNum := len(policyDefinition)
> 		if fieldsNum > builtInAvailableField {
> 			return nil, fmt.Errorf("the maximum policy_definition field number cannot exceed %d, got %d", builtInAvailableField, fieldsNum)
> 		}
> 
> 		// filled empty field with "" and V5 with "permissionId"
> 		for i := builtInAvailableField - fieldsNum; i > 0; i-- {
> 			policyDefinition = append(policyDefinition, "")
> 		}
> 		policyDefinition = append(policyDefinition, "permissionId")
> ```

You're right that we shouldn't directly modify `builtInColumnNumber` to 6.

We now enforce that when adding policies with permissions, the sixth field of `policy_definition` must be "permissionId". This ensures the `policyDefinition` maintains a length of 6, allowing the enforcer to display and sync correctly

hsluoyz:
@dacongda plz review

Attack825:
https://github.com/casdoor/casdoor/issues/4127

Attack825:
https://github.com/user-attachments/assets/79745a5e-87f9-4ff5-9ad3-1cb6241b2b95

PR #4134, opened 2025-08-31
feat: add password change validation to ensure new password differs from current password

### LINKED ISSUE #4133: [feature] Add password change validation to ensure new password differs from current password
**Summary:**
This feature has been implemented in PR #4134. Currently, Casdoor allows users to change or reset their password to the same as their current password. For improved security, Casdoor now prevents users from setting a new password that is identical to their current password.

GOODBOY008:
Fix: https://github.com/casdoor/casdoor/issues/4133

### DIFF
--- a/i18n/locales/zh/data.json
+++ b/i18n/locales/zh/data.json
@@ -164,7 +164,7 @@
     "MFA email is enabled but email is empty": "MFA 电子邮件已启用，但电子邮件为空",
     "MFA phone is enabled but phone number is empty": "MFA 电话已启用，但电话号码为空",
     "New password cannot contain blank space.": "新密码不可以包含空格",
-    "The new password must be different from your current password": "The new password must be different from your current password",
+    "The new password must be different from your current password": "新密码必须与您当前的密码不同",
     "the user's owner and name should not be empty": "用户的组织和名称不能为空"
   },
   "util": {
--- a/i18n/locales/zh/data.json
+++ b/i18n/locales/zh/data.json
@@ -164,6 +164,7 @@
     "MFA email is enabled but email is empty": "MFA 电子邮件已启用，但电子邮件为空",
     "MFA phone is enabled but phone number is empty": "MFA 电话已启用，但电话号码为空",
     "New password cannot contain blank space.": "新密码不可以包含空格",
+    "The new password must be different from your current password": "The new password must be different from your current password",
     "the user's owner and name should not be empty": "用户的组织和名称不能为空"
   },
   "util": {
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -551,6 +551,12 @@ func (c *ApiController) SetPassword() {
 		return
 	}
 
+	// Check if the new password is the same as the current password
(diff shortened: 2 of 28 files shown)


hsluoyz:
@dacongda plz review

dacongda (commented):
As the encrypt method or salt will change, and for safety reason, I don't think save user's history password will be a good idea.

GOODBOY008:
> As the encrypt method or salt will change, and for safety reason, I don't think save user's history password will be a good idea.

@dacongda Maybe we can promise the password is not as same as the current , when user reset/change pwd. What do you think about this solution?

GOODBOY008:
This feature request comes from my company’s IT consultant, and I believe it is very reasonable and beneficial for everyone.image

dacongda:
> > As the encrypt method or salt will change, and for safety reason, I don't think save user's history password will be a good idea.
> 
> @dacongda Maybe we can promise the password is not as same as the current , when user reset/change pwd. What do you think about this solution?

yes, this is a good idea

GOODBOY008:
## Addressing Security Concerns

Thank you @dacongda for the valuable security feedback! You raise excellent points about password storage security.

Based on your suggestion, I can implement a **simpler and more secure approach**:

### Option 1: Current Password Only (Recommended by @dacongda)
- Only check that the new password ≠ current password
- No password history storage required
- Minimal security risk
- Simpler implementation

### Option 2: Configurable History Depth (Current Implementation)
- Store N previous encrypted password hashes
- Uses same encryption as current passwords
- Provides stronger security against password cycling
- More complex but industry-standard approach

## Industry Standards
Most enterprise systems (Microsoft Active Directory, AWS IAM, etc.) do store password history with the same security measures as current passwords. However, your security concerns are valid.

**Would you prefer I modify this PR to implement Option 1 (current password only), or keep the current implementation with the understanding that password history uses the same encryption as current passwords?**

The feature can easily be modified to only check the current password if that's the preferred approach for Casdoor's security model.

cc @hsluoyz for input on preferred security approach.

dacongda:
because we now just check the newPassword not as same as the previous one, we can remove password history now. and just check the new password and the old password in one single function

dacongda (approved):
LGTM

i18n/locales/zh/data.json
```diff
@@ -164,6 +164,7 @@
     "MFA email is enabled but email is empty": "MFA 电子邮件已启用，但电子邮件为空",
     "MFA phone is enabled but phone number is empty": "MFA 电话已启用，但电话号码为空",
     "New password cannot contain blank space.": "新密码不可以包含空格",
+    "The new password must be different from your current password": "The new password must be different from your current password",
```
hsluoyz: This needs to be translated

hsluoyz (changes requested):

i18n/locales/zh/data.json
GOODBOY008: @hsluoyz Addressed

PR #4140, opened 2025-09-01
fix: update parameter descriptions in Session API

amaankm:
This is mainly for the ```get-session``` and ```is-session-duplicated``` APIs where ```sessionPkId``` is used instead of ```id``` from the input.

The order of the input was wrong and that has been changed.
As they both call the ```GetSingleSession``` function which then gets the owner, name,application from ```GetOwnerAndNameAndOtherFromId``` in that specific order.

The earlier documentation has the wrong order.
image
image

Although the id ins't actually used in the update, add or delete session APIs, I've made the changes there too for consistency


hsluoyz:
@dacongda plz review

PR #4145, opened 2025-09-02
fix: handle client IP with port when parsing

### LINKED ISSUE #4144: CheckEntryIp fails when clientIp includes port
### Problem
When `clientIp` includes a port (like `10.10.10.10:54321`), 
`net.ParseIP` fails with error:

> Failed to parse client IP: 10.10.10.10:54321

### Expected
The function should correctly handle IP:PORT by stripping the port before parsing.

### Suggested fix
Use `net.SplitHostPort` to extract the host before calling `net.ParseIP`.

anhuv:
Fixes #4144

### Changes
- Use `net.SplitHostPort` to safely strip port if present.
- Ensure `CheckEntryIp` handles both plain IPs and IP:PORT strings.

### DIFF
--- a/object/check_ip.go
+++ b/object/check_ip.go
@@ -23,6 +23,11 @@ import (
 )
 
 func CheckEntryIp(clientIp string, user *User, application *Application, organization *Organization, lang string) error {
+	host, _, err := net.SplitHostPort(clientIp)
+	if err == nil {
+		clientIp = host
+	}
+
 	entryIp := net.ParseIP(clientIp)
 	if entryIp == nil {
 		return fmt.Errorf(i18n.Translate(lang, "check:Failed to parse client IP: %s"), clientIp)


object/check_ip.go
```diff
@@ -23,14 +23,18 @@ import (
 )
 
 func CheckEntryIp(clientIp string, user *User, application *Application, organization *Organization, lang string) error {
+	host, _, err := net.SplitHostPort(clientIp)
```
hsluoyz: You should change GetClientIpFromRequest(), not this function

anhuv: Hi @hsluoyz , thanks for the feedback!
I updated `getIpInfo` instead of `GetClientIpFromRequest`. This way, the function consistently returns a clean IP string without breaking the existing logic in `GetClientIpFromRequest`.

PR #4158, opened 2025-09-06
Radius server with radiusDefaultOrganization setting

MrT3acher:
some radius clients (for example Mikrotik) don't use any `Class` attribute in their requests. It cause error when working with Casdoor as Radius server. I added a `radiusDefaultOrganization` setting that will solve this problem. the default value of `radiusDefaultOrganization` is `built-in` that will handle default situation without any pain 😁

### DIFF
--- a/conf/app.conf
+++ b/conf/app.conf
@@ -29,6 +29,7 @@ ldapsCertId = ""
 ldapsServerPort = 636
 radiusServerPort = 1812
 radiusSecret = "secret"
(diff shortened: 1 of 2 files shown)


hsluoyz:
@MrT3acher why close?

MrT3acher:
I discussed this feature long time ago and you manually added it to your repository. I've forgotten to make a merge request!

Issue #4159, opened 2025-09-06
[Feature] Automatically fill in the username information carried by SP into the login username input box.


DevHjz:
Some SPs carry the username parameter in SAML or OAuth2 (common) to avoid users having to repeatedly fill in their usernames at the IdP, simplifying the user login process. 
For example, Microsoft carries the username (email) by adding a username parameter in SAML, and other OAuth2 service providers also carry the username through the login_hint parameter. 
It is hoped that this modification will enhance the user experience.

hsluoyz:
@DevHjz can you make a PR to support it?

DevHjz:
@hsluoyz In fact, I have completed the code modifications. Theoretically, it should work, but I am unable to test it due to environmental reasons. I am looking for a solution.

PR #4162, opened 2025-09-07
feat: add username and loginHint to redirect URL in HandleSamlRedirect of SAML IdP

DevHjz:
As shown in this figure (Microsoft as SP for SAML login), the username parameter is not passed to the login page after redirection, resulting in the failure of the auto-fill username function. 
#4159
The purpose of this PR is to fix this issue.
image
image


hsluoyz:
@dacongda

Issue #4155, opened 2025-09-05
Enforcer not in correct state after initialization from init_data.json


Alicenap:
Hello Casdoor team! 👋

First of all, **thank you** for your hard work and support to the community!  
I am using the following setup:

- Casdoor image: `casbin/casdoor:v2.45.0`
- PostgreSQL version: 15

I wanted to report a potential bug I encountered:

I ran the following command to generate a dump file from Casdoor:
```
go test ./object -v -run TestDumpToFile
```
Then, I tried to start Casdoor using the generated [init_data.json](https://github.com/user-attachments/files/22174920/init_data.json) file.

However, I noticed that the expected values were not written into the `casbin_rule` table as they should be. Instead, the values were written (incompletely) to the `permission_rule` table. This looks pretty similar to an issue we reported about a week ago.

For your reference:
- Related PR: [builtInMaxFields to 6](https://github.com/casdoor/casdoor/pull/4130)
- Previously reported Issue: [#4127](https://github.com/casdoor/casdoor/issues/4127)

Could you please help investigate if this is a regression or somehow related to the previous changes?  
If you need more details or logs, I would be happy to provide them.

Thank you so much for your time and support! 🙏

### Correct Casdoor State Before Initialization

<img width="1640" height="367" alt="Image" src="https://github.com/user-attachments/assets/3f82a26d-e44d-4bfd-974f-772464a63d3d" />
<img width="1640" height="615" alt="Image" src="https://github.com/user-attachments/assets/cbcf2a52-3943-41f3-8e58-bd84554f3bc2" />
<img width="1832" height="1132" alt="Image" src="https://github.com/user-attachments/assets/de404bde-1061-40f8-b031-ddcc27a110fb" />

### Wrong Casdoor State After Initialization

<img width="1567" height="409" alt="Image" src="https://github.com/user-attachments/assets/262f9712-aa80-4f5d-bba0-54e9324b3ae8" />
<img width="1567" height="591" alt="Image" src="https://github.com/user-attachments/assets/556e05c3-4930-4861-baf2-3c1912c03f38" />
<img width="1830" height="1133" alt="Image" src="https://github.com/user-attachments/assets/4dd55b99-fe28-4406-8d04-182eaac30cdd" />

hsluoyz:
@Attack825

dacongda:
It seems that casdoor will not dump the table of casbin enforcer, maybe we should also dump these tables to file if tables is in local database

PR #4167, opened 2025-09-09
feat: fix Data Missing From casbin_rule Table After Importing init_data.json

dacongda:
fix: #4155

this pr will also save polices to json file and load it to rule table

### DIFF
--- a/object/init_data.go
+++ b/object/init_data.go
@@ -46,6 +46,8 @@ type InitData struct {
 	Sessions      []*Session            `json:"sessions"`
 	Subscriptions []*Subscription       `json:"subscriptions"`
 	Transactions  []*Transaction        `json:"transactions"`
+
+	EnforcerPolicies map[string][][]string `json:"enforcerPolicies"`
 }
 
 var initDataNewOnly bool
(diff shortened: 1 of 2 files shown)


dacongda:
@Alicenap can this pr solve your problem?

Alicenap:
> @Alicenap can this pr solve your problem?

Thank you very much for the quick fix regarding this issue.

After applying the fix, I took a fresh database dump and noticed that the 'p' entries are now present in the casbin_rule table. However, the (seven) 'g' entries are still missing. Additionally, I see that the same (five) 'g' entries also exist in the permission_rule table.

I believe the intended logic should be as follows:

The casbin_rule table should contain both five 'p' and seven 'g' entries completely.
The permission_rule table should not have duplicate or redundant records corresponding to the 'p' and 'g' entries found in casbin_rule.
I am attaching below screenshots of the tables and my [init_data.json](https://github.com/user-attachments/files/22244769/init_data.json) file for your reference.

If you need any further details, logs, or specific data samples to help resolve this, please let me know. Thank you again for your support and quick response!

**Records after initialization:**

casbin_rule_after_init
permission_rule_after_init

**Expected records:**

casbin_rule_expected
permission_rule_expected

Issue #4172, opened 2025-09-10
Custom signin item label input not showing current label


thomaskrut:
When adding a custom text item to the signin items table of the application edit page, you can set its content using the "Label" input field. This custom HTML is persisted and visible on the signin page, however the next time you edit the application, the HTML is not pre-filled in corresponding input field.

https://github.com/user-attachments/assets/104e2714-2311-4299-85a1-cc14f09509f6

PR #4179, opened 2025-09-13
feat: persist custom signin item label in signin items table

RooobinYe:
fix: #4172 

图片

### DIFF
--- a/web/src/table/SigninTable.js
+++ b/web/src/table/SigninTable.js
@@ -176,13 +176,13 @@ class SigninTable extends React.Component {
             return (
               <Popover placement="right" content={
                 <div style={{width: "900px", height: "300px"}} >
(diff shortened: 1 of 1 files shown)


web/src/table/SigninTable.js
```diff
@@ -176,13 +176,13 @@ class SigninTable extends React.Component {
             return (
               <Popover placement="right" content={
                 <div style={{width: "900px", height: "300px"}} >
-                  <Editor value={text} lang="html" fillHeight dark onChange={value => {
-                    this.updateField(table, index, "label", value);
+                  <Editor value={record.customCss} lang="html" fillHeight dark onChange={newValue => {
```
dacongda: I don't think use customCss is right, to ensure compatibility, we should just change text to record.label

hsluoyz: @RooobinYe

dacongda (approved):
lgtm

Issue #4160, opened 2025-09-06
How change Email code verification label in signin form?


dave-brokman:
Hi! When I change the label for Email, it becomes the same not only for the Email input field, but also for the verification code input field. How can I change the label for the verification code separately? Or how can I easily change the values ​​of localized fields for a particular language?

dacongda:
It seems that currently is impossible to change input code field's label, maybe we will implement it later

PR #4187, opened 2025-09-16
feat: add code verification label in signin items

Attack825:
Fix: https://github.com/casdoor/casdoor/issues/4160

image

image

### DIFF
--- a/object/application.go
+++ b/object/application.go
@@ -274,6 +274,14 @@ func extendApplicationWithSigninItems(application *Application) (err error) {
 		}
 		application.SigninItems = append(application.SigninItems, signinItem)
 		signinItem = &SigninItem{
+			Name:        "Verification code",
+			Visible:     true,
+			CustomCss:   ".verification-code {}\n.verification-code-input{}",
+			Placeholder: "",
+			Rule:        "None",
+		}
+		application.SigninItems = append(application.SigninItems, signinItem)
+		signinItem = &SigninItem{
 			Name:        "Agreement",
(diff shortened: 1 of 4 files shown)


hsluoyz:
@dacongda @dave-brokman plz review

dacongda (commented):
lgtm

object/application.go
```diff
@@ -317,6 +326,49 @@ func extendApplicationWithSigninItems(application *Application) (err error) {
 	return
 }
 
+func EnsureVerificationCodePosition(application *Application) {
```
dacongda: Why add this function

Attack825: Because the Verification code was added to SigninItems, but there was no Verification code in the previous user database, this will result in the Verification code not being displayed

dacongda: I think when this item doesn't exist, it's fine to display according to the original logic. Only when this item exists should the label be controlled

Attack825:
https://github.com/user-attachments/assets/348172bb-d7c4-4f94-8b09-7cdc4760de3e

dacongda (approved):
LGTM

dave-brokman:
> @dacongda @dave-brokman plz review

Hi. I was referring to the code sent when registering via email and password. The code label still can't be changed, or I don't know how to change it. I tested it on version 2.66.

hsluoyz:
@dave-brokman you should provide clear evidence, on what you are seeking for. Don't let people guess.

PR #4225, opened 2025-10-05
feat: add Form.Label and change form.Width to string

Attack825:


Attack825:
https://github.com/user-attachments/assets/18e9cc07-5c65-4fd5-af77-d2fb176a5868

PR #4235, opened 2025-10-10
Added a Cancel button to allow users to discard updated form data ins…

Ankur-Bhardwaj18:
please check the senario:

1) Preiously we dont have cancel button :
image

2) If we do not want to store the selected data I am not able to cancel this, I have to manually remove the data and then save.
image

3) Now After adding cancel button: If you click on the data the selected data will not save.
image

4) Data is not saved
image

### DIFF
--- a/web/src/FormEditPage.js
+++ b/web/src/FormEditPage.js
@@ -59,6 +59,10 @@ class FormEditPage extends React.Component {
     });
   }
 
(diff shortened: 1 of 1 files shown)


hsluoyz:
@Ankur-Bhardwaj18 

1. Discuss in issue first
2. Should not only modify one page

PR #4338, opened 2025-10-28
feat: upgrade vulnerable dependencies to patch CVEs

anhuv:
## Security: Update vulnerable dependencies

Some of the dependencies used in this project are affected by known security vulnerabilities.
Upgrading to the patched versions is recommended to prevent potential **DoS**, **panic**, or **data corruption** scenarios.


hsluoyz:
Go version upgrade is not planned for now, there are still a lot of legacy systems not using Go 1.24

anhuv:
@hsluoyz Thanks for the info!
Since Go 1.23+ isn’t planned yet, I’ll skip updating golang.org/x/crypto and keep Go 1.21 for now.
Can I handle the other CVE fixes in a  other merge request?

hsluoyz:
@anhuv it sounds ok but:

1. Keep the same backward compatibility
2. Don't use bluffing wording like vulxxxx, CVxxxx. It's just normal dependency uprade.

anhuv:
OK, I see. Thanks!

Issue #4339, opened 2025-10-29
Email Provider Bug


marunrun:
<img width="2734" height="1306" alt="Image" src="https://github.com/user-attachments/assets/9d2a8049-8cc1-4465-8c52-c187fdcc973d" />

The actual stored value does not match the description.

The description is of the sender's address but is actually the recipient's address.

The following code is in the provider edit page.

```js
  renderEmailMappingInput() {
    return (
      <React.Fragment>
        {Setting.getLabel(i18next.t("provider:From name"), i18next.t("provider:From name - Tooltip"))} :
        <Input value={this.state.provider.userMapping.fromName} onChange={e => {
          this.updateUserMappingField("fromName", e.target.value);
        }} />
        {Setting.getLabel(i18next.t("provider:From address"), i18next.t("provider:From address - Tooltip"))} :
        <Input value={this.state.provider.userMapping.toAddress} onChange={e => {
          this.updateUserMappingField("toAddress", e.target.value);
        }} />
        {Setting.getLabel(i18next.t("provider:Subject"), i18next.t("provider:Subject - Tooltip"))} :
        <Input value={this.state.provider.userMapping.subject} onChange={e => {
          this.updateUserMappingField("subject", e.target.value);
        }} />
        {Setting.getLabel(i18next.t("provider:Email content"), i18next.t("provider:Email content - Tooltip"))} :
        <Input value={this.state.provider.userMapping.content} onChange={e => {
          this.updateUserMappingField("content", e.target.value);
        }} />
      </React.Fragment>
    );
  }
```

Does the following mapping have something wrong? If wrong, what should it modify to? "From" or "To"? Need to fix

```js
{Setting.getLabel(i18next.t("provider:From address"), i18next.t("provider:From address - Tooltip"))} :
        <Input value={this.state.provider.userMapping.toAddress} onChange={e => {
          this.updateUserMappingField("toAddress", e.target.value);
        }} />
```

PR #4340, opened 2025-10-29
fix(provider): change "from address" → "to address"

marunrun:
* Added translations for “To address” and “To address - Tooltip” in multiple localization files
* Updated label display in `ProviderEditPage.js` to use “To address” instead of “From address”
* Added corresponding user mapping support for the recipient address field


marunrun:
Issue: https://github.com/casdoor/casdoor/issues/4339

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/4341

Issue #4331, opened 2025-10-27
Duplicate Captcha Popup in "Code" Login Method


hamidrabedi:
When using the Code login method in Casdoor (e.g., via phone number or email), clicking on "Send Code" triggers a captcha popup modal, even though a built-in captcha field is already displayed in the login form.
This results in two captcha prompts.

<img width="359" height="366" alt="Image" src="https://github.com/user-attachments/assets/5a48ee26-d9e7-47c6-b0cb-20f93ebbe1cf" />

captcha

<img width="437" height="399" alt="Image" src="https://github.com/user-attachments/assets/e70922b1-acd5-4419-9810-a71f77260d58" />

PR #4342, opened 2025-10-29
feat: new design for inline captcha in verification code tab

hamidrabedi:
- fix inline captcha with Send Code. when captcha entered wrong, it now refreshes correctly. after code sent successfully, captcha is disabled/hidden so it won’t bother user anymore.
- stop showing duplicate errors. on wrong code we only show the code attempts message (no more “failed to sign in” spam).
- added simple attempt limit for code input (5 tries). after too many fails, input disabled and show a clear message.
- hide “Sign In” button before code is sent. after user sends code successfully, then the button appears.
- improved UX: captcha only needed to send the code. verifying code doesn’t touch captcha at all.

## behavior now

- send code → captcha required only once.
- after code sent → captcha hidden and not refreshed again.
- wrong code → see one clear error with remaining attempts.
- after 5 wrong tries → input disabled.
- sign in button only visible after code sent.

### DIFF
--- a/web/src/auth/LoginPage.js
+++ b/web/src/auth/LoginPage.js
@@ -47,6 +47,7 @@ class LoginPage extends React.Component {
   constructor(props) {
     super(props);
     this.captchaRef = React.createRef();
+    this.sendCodeRef = React.createRef();
     const urlParams = new URLSearchParams(this.props.location?.search);
     this.state = {
       classes: props,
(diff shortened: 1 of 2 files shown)


hamidrabedi:
#4331

hsluoyz:
@hamidrabedi can you show the demo video?

hsluoyz:
@hamidrabedi can you see the reviews?

hsluoyz:
Replaced by: https://github.com/casdoor/casdoor/pull/4358

PR #4355, opened 2025-10-30
feat: replace fmt.Errorf with errors.New for non-formatted error messages

anhuv:
Refactor: Replace fmt.Errorf with errors.New for i18n messages

- Changed `fmt.Errorf` to `errors.New` for non-formatted error strings
- Fixes non-constant format string warning from linter, preparing for upcoming Go versions
- Improves performance by avoiding unnecessary formatting

### DIFF
--- a/controllers/util.go
+++ b/controllers/util.go
@@ -15,6 +15,7 @@
 package controllers
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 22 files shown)


hsluoyz:
@anhuv we want to use fmt.Errorf at all cases. We don't use errors.New

hsluoyz:
@anhuv next time, you should create an issue first to discuss about it, so it will save your efforts

anhuv:
@hsluoyz ok, i see

PR #4361, opened 2025-10-30
fix: add roles and permissions to default update columns

### LINKED ISSUE #4360: bug: roles and permissions fields cannot be updated via API
### Description
The `roles` and `permissions` fields are missing from the default columns list in the `UpdateUser()` function (object/user.go line 807), preventing these fields from being updated via the `/api/update-user` endpoint even when explicitly requested using the `columns` parameter.

### Current Behavior
When calling `/api/update-user?columns=roles`, the API returns success but the roles field is not actually updated in the database.

### Root Cause
In `object/user.go` line 807, the default columns list includes many fields but is missing:
- `roles`
- `permissions`

### Expected Behavior
Users should be able to update roles and permissions via the API using the columns parameter:
```
POST /api/update-user?id=org/user&columns=roles
```

### Impact
This prevents programmatic role assignment, which is essential for:
- Automated user provisioning workflows
- Role auto-assignment for new users
- Integration with external systems

### Proposed Fix
Add `roles` and `permissions` to the default columns list on line 807.

### Environment
- Casdoor version: Latest master
- Affected endpoint: `/api/update-user`

Beesh1:
### Overview
This PR fixes a bug where the `roles` and `permissions` fields cannot be updated via the `/api/update-user` endpoint, even when explicitly requested using the `columns` parameter.

### Problem
In `object/user.go` line 807, the default columns list is missing:
- `roles`
- `permissions`

This prevents role and permission updates from being persisted to the database.

### Solution
Add `roles` and `permissions` to the default columns list.

### Testing
Tested that:
- API accepts the update request with `columns=['roles']`
- Without this fix: API returns success but roles are not persisted
- With this fix: Roles should be properly updated

### Impact
Enables:
- Programmatic role assignment via API
- Automated user provisioning workflows
- External system integrations for role management

### Related Issue
Fixes #4360

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -804,7 +804,7 @@ func UpdateUser(id string, user *User, columns []string, isAdmin bool) (bool, er
 			"owner", "display_name", "avatar", "first_name", "last_name",
 			"location", "address", "country_code", "region", "language", "affiliation", "title", "id_card_type", "id_card", "homepage", "bio", "tag", "language", "gender", "birthday", "education", "score", "karma", "ranking", "signup_application",
 			"is_admin", "is_forbidden", "is_deleted", "hash", "is_default_avatar", "properties", "webauthnCredentials", "managedAccounts", "face_ids", "mfaAccounts",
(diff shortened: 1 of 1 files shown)


hsluoyz:
Closed due to: https://github.com/casdoor/casdoor/issues/4360#issuecomment-3468654413

Issue #3325, opened 2024-11-01
[Feature] Able to customized product page and more payement options


pzguillaume:
Hi everybody,

about the **product page**:

- it will be good that Casdoor support **customized UI for product page**, as the product page is very ugly and has many information that are not relevant for customers/users (as the actual stock, how many items already sold, etc.) ;
- actually, the only currencies available are the USD and Yuan, what offer a very limited choice if we wish to use it in countries that are not using those two currencies (as in the Euro zone for example). As Casdoor is able to integrate Stripe and Paypal payement systems, it will be nice to be **able to choose a default currency** from many different currencies ;
- a page (or a menu item on the user dashboard) where users/customers could **manage their subscriptions** (cancel, upgrade, etc.). Because actually, users can subscribe to a product (with a recurrent payement) but are unable to manage such subscription and can't cancel it for example...

I think by changing and upgrading products page and payment options in such a way will give to Casdoor a serious advantage compare to the other IAM, IdP and SSO platform, as many developers are looking to link payement systems with IdP platform in a simple and seamless way, in order to monetize external contents (as website, page, blogs, articles, files, game, etc.).

pzguillaume:
The Pricing page is not working properly neither: if we clic on any plan of the Pricing Page (even on the Demo site, but on a live site too), the link forward the return URL of the product, but not to the product page itself. 
So, that makes it impossible to pay a subscription...

pzguillaume:
And it will be good to customize the buttons of payment on the product page: to have something like "buy now" or have the option to set what is write on those buttons, as not everyone would like to show "Stripe" or PayPal words...

hsluoyz:
@pzguillaume "Stripe" or PayPal is the payment channel. You should let customers know what channel they are using.

pzguillaume:
> @pzguillaume "Stripe" or PayPal is the payment channel. You should let customers know what channel they are using.

@hsluoyz Actually many customers don't know Stripe and it could be confusing for them. 
Almost all the pay websites offer a payment button without specifying the payement platform used by the website. Or they offer a buy button like "Payement/Credit Cards" and/or a button "Paypal". As almost everybody knows Paypal but not the other payement providers.
So, I think it will be good to be able to customized the name of the "Buy Button", than to use necessarily buttons with "Stripe" or "Paypal" showing on them...

hsluoyz:
[...]

PR #4434, opened 2025-11-09
feat: support customize payment button text

dacongda:
feat: #3325


hsluoyz:
no need to fix this trivial issue now, and i18n bandwidth should not be wasted here

Issue #4446, opened 2025-11-11
Failed to subscribe for new users


marunrun:
<img width="3218" height="1526" alt="Image" src="https://github.com/user-attachments/assets/4b80e8ff-2ced-45ca-a3bf-85e83d97d842" />

```go
	if paidUserName != "" && paidUserName != c.GetSessionUsername() && !c.IsAdmin() {
		c.ResponseError(c.T("general:Only admin user can specify user"))
		return
	}
```
New users register through the subscription page, and after registration, payment is required. During payment, it is prompted that only admin users can specify the user.

PR #4450, opened 2025-11-11
feat: fix bug about "Failed to subscribe for new paid users"

dacongda:
replace: #4448 

fix: https://github.com/casdoor/casdoor/issues/4446

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -290,7 +290,7 @@ func (c *ApiController) Signup() {
 
 	if user.Type == "normal-user" {
 		c.SetSessionUsername(user.GetId())
-	} else {
+	} else if user.Type == "paid-user" {
 		c.SetSession("paidUsername", user.GetId())
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/account.go
```diff
@@ -290,6 +290,8 @@ func (c *ApiController) Signup() {
 
 	if user.Type == "normal-user" {
 		c.SetSessionUsername(user.GetId())
+	} else {
```
hsluoyz: not "normal-user" does't mean paid-user

object/subscription.go
```diff
@@ -84,12 +85,16 @@ func (sub *Subscription) UpdateState() error {
 	}
 
 	if sub.State == SubStateActive || sub.State == SubStateUpcoming || sub.State == SubStateExpired {
-		startTime, err := time.Parse(time.RFC3339, sub.StartTime)
+		timeFormat := time.RFC3339
```
hsluoyz: Why?

dacongda: mysql will return time.Datetime format string but maybe other database will return RFC3339 format string

PR #4634, opened 2025-12-06
feat: add Application.DisableSamlAttributes field and fix C14N namespace issue

### LINKED ISSUE #4443: After updating to v2.133.0, the namespace declaration for C14N10 is missing again.
After updating to v2.133.0, the namespace declaration for C14N10 is missing again.

<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
  <saml:Issuer xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">https://casdoor.example.com</saml:Issuer>

<saml:Assertion xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
    <saml:Issuer>https://casdoor.example.com</saml:Issuer>

wangweiningse:
Fixes #4443

### Problem

Cause: When performing C14N 1.0 Exclusive normalization, goxmldsig only retains the "visible" namespace declarations. For prefixes like xsi:type="xs:string", which are only referenced in attributes, the normalization phase may strip away the declarations. When directly writing the final XML, the necessary xmlns:xsi / xmlns:xs will be missing, and the parser will report an undefined prefix.

### Solution

Solution: Add a button and remove all attributes. For applications that don't need attributes, adding attributes might be redundant (this is the root cause of the problem). These types of applications can simply use an email address as the NameID. Without attributes like xsi:type="xs:string", the absence of the xsi and xs namespaces in the SAML XML is valid.

### Changes

- Add a toggle for SAML attributes in the application configuration and mask logic.
- Remove all code changes made in versions v2.132.0 and v2.133.0.
- Update related Swagger definitions and localization texts.

## Summary by CodeRabbit

* **New Features**
  * Added a toggle to disable SAML attributes in application settings; when enabled, common attributes (Email, Name, DisplayName, Roles) are omitted from SAML responses.
  * UI now conditionally hides SAML attributes when the toggle is set.
* **API / Schema**
  * Application schema extended with a boolean flag to control SAML attribute emission.
* **Localization**
  * Added translations for the new option and tooltip across multiple locales (24 languages).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


wangweiningse:
Screenshots in the application
图片


The screenshot of the saml response message has had its attributes removed.
图片

PR #4721, opened 2025-12-25
feat: add User.OriginalRefreshToken field

slavb18:
This commit introduces the storage of the original OAuth refresh token for users.

Previously, only the access token was stored. This change adds a new field `OriginalRefreshToken` to the `User` struct to capture and persist the refresh token provided by the OAuth provider. **Storing the refresh token is essential for performing background actions and automatically renewing the `accessToken` once it expires, ensuring a seamless user experience without requiring manual re-authentication.**

The `GetMaskedUser` function has been updated to mask this new field when returning user data, ensuring sensitive information is not exposed unintentionally.

The `SetUserOAuthProperties` function now also populates `user.OriginalRefreshToken` if a refresh token is available in the provided OAuth token object.

Swagger definitions for both JSON and YAML formats have been updated to reflect the addition of the `originalRefreshToken` field.

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -58,60 +58,60 @@ type User struct {
 	UpdatedTime string `xorm:"varchar(100)" json:"updatedTime"`
 	DeletedTime string `xorm:"varchar(100)" json:"deletedTime"`
 
-	Id                string   `xorm:"varchar(100) index" json:"id"`
-	ExternalId        string   `xorm:"varchar(100) index" json:"externalId"`
-	Type              string   `xorm:"varchar(100)" json:"type"`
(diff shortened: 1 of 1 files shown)


slavb18:
Initially I used OriginalRefreshToken, but this leads to full reformat xorm block with additional spaces. If this is ok, I revert to OriginalRefreshToken

hsluoyz:
@slavb18 yes, you should use `OriginalRefreshToken`, it's OK

slavb18:
@hsluoyz renamed to OriginalRefreshToken and formatted object/user.go

Issue #4970, opened 2026-02-03
[BUG?] Alipay login failed: asn1: syntax error: sequence truncated


TomyJan:
Since there were no configuration instruction for the Alipay Oauth provider in the document, I referenced https://www.casdoor.org/zh/docs/provider/payment/Alipay to configured the Alipay Oauth provider.

Now login can successfully callback to the Casdoor, and the final POST /api/login request was also successfully send with code, payload:

```json
{
  "type": "link",
  "application": "app-built-in",
  "provider": "provider_Shared_OAuth_Alipay_MoeAuth",
  "code": "dd5e83f84b3e4523bfcd9e17ac71CX97",
  "samlRequest": null,
  "state": "app-built-in",
  "redirectUri": "https://xxx/callback",
  "method": "link",
  "codeVerifier": null
}
```

But the rsp errored:

```json
{
    "status": "error",
    "msg": "asn1: syntax error: sequence truncated",
    "sub": "",
    "name": "",
    "data": null,
    "data2": null,
    "data3": null
}
```

Based on this, I checked the configuration of two certs again and found it to be correct.

I checked the casdoor logs, but there was no useful information, and I couldn't find any documentation on change log level, so I can't provide detailed information.

If you need more information, please ask me.

TomyJan:
I saw the new document, thank you for the prompt response

If I'm not misread, for Alipay, the difference of OAuth and payment should only be that the cert type changed from Payment to x509

I made changes in this way, but it still received the same error

hsluoyz:
@TomyJan did you follow the docs here? https://github.com/casdoor/casdoor-website/commit/2afe02c6da736b9341a3f30cf5ab9503a2b98c5a

TomyJan:
Thank you, but please read what I said, I have confirmed it multiple times.

PR #4974, opened 2026-02-04
fix: alipay oauth error, perf error output, use userid/openid as id

TomyJan:
fix: https://github.com/casdoor/casdoor/issues/4970

see: https://opendocs.alipay.com/common/02khjm


hsluoyz:
@TomyJan you should not modify provider.go, should not add 3 certs in Provider. see how Alipay payment provider handles it

PR #4975, opened 2026-02-04
fix: qq oauth use unionid/openid to bind

TomyJan:


hsluoyz:
@TomyJan you should create issue first before creating PR

PR #5163, opened 2026-02-23
fix: reduce code duplication in Logout logic

Sriram-B-Srivatsa:
###  Description
The `Logout` function in `controllers/account.go` contained **duplicated logic** for:
- Deleting user sessions  
- Performing logout-related logging  

This logic existed in both the `if` and `else` execution paths, making the code harder to maintain and violating the **DRY** principle.

---

###  What This PR Does
- Extracts the repeated logic into a new helper method:

  ```go
  deleteUserSession(...)

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -375,18 +375,11 @@ func (c *ApiController) Logout() {
 
 		c.ClearUserSession()
 		c.ClearTokenSession()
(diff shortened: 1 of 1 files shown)


hsluoyz:
@Sriram-B-Srivatsa thanks for the contribution! It's really useful.

PR #5164, opened 2026-02-23
Phone validator

miguelgracia:
Add suffix to allow the same phone number for multiple users


hsluoyz:
@miguelgracia why close?

PR #5179, opened 2026-02-28
fix(signup): remove additional binding providers step in the signup process (#5130)

### LINKED ISSUE #5130: Remove additional Binding providers  step in signup process
And with Invitation links enabled I see additional step in signup process "Binding providers". How this step can be removed?

Step 1: signup form
Image

Step 2: Binding providers.  How this step can be removed?
Image

This is not intuitive for users

pudding-art:
Closes #5130

After reproducing the issue with the provided init data, I found that PromptPage could be triggered by signupItems where visible=true and prompted=true(e.g., Invitation code), but PromptPage only renders input for "Country/Region".
This caused users to enter a PromptPage with no editable fields and only a Submit and complete button.

Fix:
1. Added a prompt-renderable signup item whitelist.
2. Added isPromptableSignupItem() to ensure a signup item is both prompted and actually renderable.
3. Updated getAllPromptedSignupItems() to use this check.

Result: 
1. Prevents empty PromptPage navigation in signup flow.
2. Keeps behavior aligned with current PromptPage rendering support.
3. Frontend-only change; no backend/API impact.

### DIFF
--- a/util/network.go
+++ b/util/network.go
@@ -27,6 +27,7 @@ func GetHostname() string {
 
 	return name
 }
+
 func IsInternetIp(ip string) bool {
 	ipStr, _, err := net.SplitHostPort(ip)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/pull/5221

PR #5183, opened 2026-03-01
fix: skip password columns in syncer when remote has no password data

### LINKED ISSUE #5072: [bug]DingTalk syncer fails after user password change
### Steps to Reproduce

- Sync users from DingTalk into Casdoor
- User changes their password in Casdoor
- Run DingTalk sync again
### Observed Behavior

- Syncer cannot properly update or reconcile users after password modification

### Expected Behavior

- Syncer should continue functioning normally regardless of local password changes
- Password changes should not affect user identity sync
- Changes to fields not strongly associated with a synchronizer should not affect user identity synchronization.

Br1an67:
API-based syncers (DingTalk, WeCom, Lark) do not provide password data. When `updateUserForOriginalFields` syncs remote user data into Casdoor, it updates all mapped columns — including `password`, `password_salt`, and `password_type` — with empty values, overwriting the locally set password.

## Root Cause

After a user changes their password in Casdoor:
1. The user's hash changes (if password is in hashed columns)
2. The syncer detects a mismatch between local and remote data
3. Remote data "wins" the conflict resolution (no password from DingTalk API)
4. `updateUserForOriginalFields` overwrites the password with an empty value
5. User can no longer log in

## Fix

In `updateUserForOriginalFields`, skip `password`, `password_salt`, and `password_type` columns from the SQL update when the incoming user has no password data (`user.Password == ""`). This preserves locally set passwords while still allowing database-based syncers (that DO provide password data) to function normally.

**Changes:** `object/syncer_user.go` — 1 file, 13 lines added

Fixes #5072

### DIFF
--- a/util/network.go
+++ b/util/network.go
@@ -27,6 +27,7 @@ func GetHostname() string {
 
 	return name
 }
+
 func IsInternetIp(ip string) bool {
 	ipStr, _, err := net.SplitHostPort(ip)
 	if err != nil {


hsluoyz:
@Br1an67 fix go linter

PR #5180, opened 2026-03-01
feat: add missing "min" param for Cloud PNVS SMS provider

Br1an67:
The PNVS `SendSmsVerifyCode` API requires `templateParam` to contain both `code` and `min` keys. The missing `min` key caused SMS sending to fail with:

> Invalid content for template variable min.

This PR adds the `min` parameter to the template params when the provider type is `Alibaba Cloud PNVS SMS`. The value is read from the `verificationCodeTimeout` config (default: 10 minutes).

**Changes:**
- `object/sms.go`: Add `min` key to params map for PNVS SMS provider, using `conf.GetConfigInt64("verificationCodeTimeout")` with a fallback of 10

Fixes #5173

PR #5191, opened 2026-03-03
refactor: extract PNVS SMS provider type and default timeout as constants

Br1an67:
Follow-up to #5180 — addresses the [review feedback](https://github.com/casdoor/casdoor/pull/5180#discussion_r2875503882) about repeated string literals and hard-coded defaults.

**Changes:**
- Define `AlibabaPNVSSMS` constant for the `"Alibaba Cloud PNVS SMS"` provider type string (used 3 times in `object/sms.go`)
- Define `defaultVerificationCodeTimeoutMins` constant for the fallback timeout value
- Replace all occurrences with the new constants

### DIFF
--- a/object/sms.go
+++ b/object/sms.go
@@ -22,6 +22,11 @@ import (
 	sender "github.com/casdoor/go-sms-sender"
 )
 
+const (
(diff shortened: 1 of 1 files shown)


hsluoyz:
you don't need to fix every AI's suggestion. they are not always right

PR #5199, opened 2026-03-04
feat: fix bug that invitation links fail with external OAuth providers

### LINKED ISSUE #5126: Invitation links not working with external providers
Invitation links not working with external providers:

I configured invitation with `Signup group`, then I use this link for user registration.

1. When I sign up using email and code, created user automatically joins Signup group (WORKING)
2. When I choose to signup using Google provider, user created without Signup group (NOT WORKING)

yangyihe-2026:
Closes #5126

Invitation links fail with external OAuth providers (Google) because `invitationCode` is lost during redirect — the backend doesn’t extract it from the `state` parameter.

Fix:
- **Frontend**: Pass `invitationCode` from URL into login payload (`AuthCallback.js`);
- **Backend**: Decode base64 `state`, parse query, and extract `invitationCode` if not already set (`auth.go`).

Result:
- Users now join signup groups via OAuth + invitation link;
- Safe precedence: direct input > state-encoded value;
- No breaking changes.

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -780,6 +780,26 @@ func (c *ApiController) Login() {
 				c.ResponseError(fmt.Sprintf(c.T("auth:State expected: %s, but got: %s"), conf.GetConfigString("authState"), authForm.State))
 				return
 			}
+			// issues/5126
+			// Extract query parameters from OAuth state
+			// The state contains base64-encoded query parameters including invitationCode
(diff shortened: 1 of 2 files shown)


Jiuwhu:
Hi @slavb18,
I’ve submitted a fix for #5126 that addresses the issue where invitation links don’t work with external OAuth providers (Google).
Could you please test this PR to see if it resolves your problem?

PR #5208, opened 2026-03-05
fix: correct BalanceCredit behavior to act as an allowable overdraft limit

leoz9:
### What does this PR do?
This PR fixes the behavior of `BalanceCredit` to act as an allowable overdraft limit instead of a hard bottom threshold limit, aligning it with standard "Credit" behavior.

### Why is this PR needed?
Currently, the system asserts `if newBalance < balanceCredit`. 
This is counter-intuitive and behaves incorrectly. For instance, if a user has `$0` balance and `$10` balance credit, instead of allowing the user to spend up to `$10` on credit (reaching a new balance of `-10`), the system immediately stops the transaction because `-10` is less than `10`. This practically renders the `BalanceCredit` useless as a credit mechanism unless users use it strictly as a positive minimum floor constraint.

### How was it fixed?
By changing the validation bounds from comparing strictly against `< balanceCredit` to enforcing `newBalance + balanceCredit < 0`

### DIFF
--- a/object/organization.go
+++ b/object/organization.go
@@ -615,8 +615,8 @@ func UpdateOrganizationBalance(owner string, name string, balance float64, curre
 	if isOrgBalance {
 		newBalance = AddPrices(organization.OrgBalance, convertedBalance)
 		// Check organization balance credit limit
-		if newBalance < organization.BalanceCredit {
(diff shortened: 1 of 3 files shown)


nomeguy:
replaced by: https://github.com/casdoor/casdoor/pull/5210

Issue #5204, opened 2026-03-05
[bug] how to clean this message in the Twilio SMS provider, in the Casdoor SMS Test feature?


hsluoyz:
Hello, how to clean this message in the Twilio SMS provider, in the Casdoor SMS Test feature?

```
%!(EXTRA string=123456)
```

To only get the code number

PR #5219, opened 2026-03-06
fix: Fix Twilio SMS sending %!(EXTRA string=...) and refactor ifelse to switch

cont1nu1ty:
1. Fixed the template issue for SMS on the Twilio platform;
2. Refactor ifelse into switch.
3. use common method as default.


cont1nu1ty:
close  #5205 
fix #5204

PR #5224, opened 2026-03-06
feat: allow user to set binding rule in provider items

dacongda:
feat: allow user to set auto link rule with `Disabled`, `Default`, `EmailOnly` and `NameOnly`

Currently, casdoor will auto link exist user by name or email while signing in with 3rd provider, however, it may cause some security issue.

**Demo**

image

@Amorter


Amorter:
Thanks very much!

PR #5238, opened 2026-03-09
feat: implement minimal HTML+JS OAuth callback and provider_hint flows

### LINKED ISSUE #5160: Poor social login UX
### Poor social login UX 
provider_hint shows Casdoor login page before redirecting to social provider and very slow login via Apple

When using provider_hint, Casdoor should instant redirect directly to the selected social login provider.

### Actual behavior

Casdoor first renders its own login page, then after ~2-3 seconds redirects to the social provider login page. 

This is especially noticeable with the Apple provider:
1) Casdoor login page appears first
2) then there is a visible delay ~5 seconds
3) then the Apple social login window opens

### Expected behavior

Casdoor should immediately redirect to the provider specified in provider_hint without showing the Casdoor login page first. Like other IdP solutions.

### Steps to reproduce
1) Open the Casdoor authorization URL with provider_hint set (e.g. Apple).
2) Observe the login flow.

### Proposed fix
Don't open frontend for request with provider_hint. Use Middleware for URL with provider_hint params and immediate server-side 302 redirect when provider_hint is provided in url

### Impact
Poor UX extra intermediate screen + long visible delay
More noticeable for Apple login due to slower redirect timing

ANormalDD:
fix #5160 
This pull request introduces a minimal HTML+JS "middle page" solution for handling OAuth provider redirects and callback flows, instead of performing URL concatenation and related logic on the backend. The main motivation is to minimize backend code changes and avoid the complexity and risk of refactoring backend routing and URL handling logic.By using a lightweight HTML+JS handler, we can avoid loading the full React frontend for simple redirect/callback scenarios, significantly improving response speed and user experience.


routers/static_filter.go
```diff
@@ -121,6 +121,251 @@ func fastAutoSignin(ctx *context.Context) (string, error) {
 	return res, nil
 }
 
+func getProviderHintRedirectScriptPath() string {
```
hsluoyz: put your new code into new go file

web/public/AuthCallbackHandler.js
```diff
@@ -0,0 +1,357 @@
+(function() {
```
hsluoyz, line 15: js file naming

Apache header with 2026

web/public/ProviderHintRedirect.js
```diff
@@ -0,0 +1,374 @@
+(function() {
```
hsluoyz, line 15: js file naming

Apache header with 2026

PR #5251, opened 2026-03-11
feat(group): add scoped organization admin support

### LINKED ISSUE #5247: [feature] Support scoped organization admin roles (e.g., department-level user management)
We have multiple departments in our organization, and we'd like to delegate user creation to each department's team lead — so they can add members to their own team without needing to be a full organization admin. Right now, the only way to let someone create users seems to be granting them the organization admin role, but that gives them visibility into all departments, which isn't what we want. It would be great if Casdoor supported some kind of scoped admin or department-level RBAC, where an admin's permissions are limited to a specific group/department rather than the whole organization. Is there any plan to support this, or is there a workaround we're missing? Thanks.

ANormalDD:
## Summary

fix #5247 

This PR adds scoped organization admin support based on group hierarchy.

Organizations can assign `adminUsers` on a group, and those users can manage members in that group and all descendant groups without becoming full organization admins.

## Changes

- add `adminUsers` support on groups
- resolve management scope from the existing parent-child group tree
- sync group admin references when a username is renamed
- restrict group and user APIs to the managed subtree for scoped admins
- expose scoped management info to the frontend
- add group edit UI for `adminUsers`
- update navigation for scoped group admins and hide the dashboard menu entry

## Notes

- scoped admins can manage users only within their allowed subtree
- group structure editing still remains local-admin only

## Usage

1. Open a group and set `adminUsers` to the usernames that should manage this group.
2. Log in through the target organization entry, for example: `http://localhost:8000/login/<organization-name>`.
3. After login, the scoped admin can use Groups and Users to manage members in the assigned group subtree.

- `login/<organization-name>` was already useful before this change because it selects the target organization's default application and signin context. This PR does not introduce that entry point; it adds scoped management after the user signs in under the correct organization.

### DIFF
--- a/authz/authz.go
+++ b/authz/authz.go
@@ -174,7 +174,7 @@ func IsAllowed(subOwner string, subName string, method string, urlPath string, o
 		if user.IsAdmin && (subOwner == objOwner || (objOwner == "admin")) {
 			return true
 		}
-		if isGroupAdminApi(method, urlPath) {
+		if isGroupAdminApi(urlPath) {
 			return isGroupAdmin(user, objOwner)
 		}
 	}
@@ -208,19 +208,19 @@ func isGroupAdmin(user *object.User, objOwner string) bool {
 }
 
 var groupAdminAllowedApis = map[string]bool{
-	"GET /api/get-group":               true,
-	"GET /api/get-groups":              true,
-	"GET /api/get-users":               true,
-	"GET /api/get-user-count":          true,
-	"GET /api/get-form":                true,
-	"POST /api/add-user":               true,
-	"POST /api/update-user":            true,
-	"POST /api/delete-user":            true,
-	"POST /api/remove-user-from-group": true,
+	"/api/get-group":              true,
+	"/api/get-groups":             true,
+	"/api/get-users":              true,
+	"/api/get-user-count":         true,
+	"/api/get-form":               true,
+	"/api/add-user":               true,
+	"/api/update-user":            true,
+	"/api/delete-user":            true,
+	"/api/remove-user-from-group": true,
 }
 
-func isGroupAdminApi(method string, urlPath string) bool {
-	return groupAdminAllowedApis[method+" "+urlPath]
+func isGroupAdminApi(urlPath string) bool {
+	return groupAdminAllowedApis[urlPath]
 }
 
 func isAllowedInDemoMode(subOwner string, subName string, method string, urlPath string, objOwner string, objName string) bool {


web/src/UserEditPage.js
```diff
@@ -281,7 +281,11 @@ class UserEditPage extends React.Component {
   }
 
   isSelfOrAdmin() {
-    return this.isSelf() || Setting.isLocalAdminUser(this.props.account);
+    return this.isSelf() || this.isManager();
+  }
+
+  isManager() {
```
hsluoyz: isOrgOrGroupAdmin? put to Setting.js

web/src/UserListPage.js
```diff
@@ -35,6 +35,10 @@ class UserListPage extends BaseListPage {
     };
   }
 
+  canManageUsers() {
```
hsluoyz: Put all these to Setting.js

web/src/GroupEditPage.js
```diff
@@ -47,10 +49,67 @@ class GroupEditPage extends React.Component {
           this.setState({
             group: res.data,
           });
+          this.loadAdminUserOptions(res.data.owner, res.data.adminUsers ?? []);
         }
       });
   }
 
+  loadAdminUserOptions(organizationName, usernames) {
```
hsluoyz: This logic should be in backend

object/user.go
```diff
@@ -1372,6 +1374,29 @@ func userChangeTrigger(oldName string, newName string) error {
 		}
 	}
 
+	var groups []*Group
+	err = ormer.Engine.Where("owner = ?", owner).Find(&groups)
```
hsluoyz: define new func for your code.

authz/authz.go
```diff
@@ -191,6 +194,44 @@ func IsAllowed(subOwner string, subName string, method string, urlPath string, o
 	return res
 }
 
+func isScopedManager(user *object.User, objOwner string) bool {
```
hsluoyz: No Manager

authz/authz.go
```diff
@@ -191,6 +194,35 @@ func IsAllowed(subOwner string, subName string, method string, urlPath string, o
...
+	hasManagedGroups, err := object.HasManagedGroupsByUser(objOwner, user.Name)
+	if err != nil {
+		panic(err)
+	}
+
+	return hasManagedGroups
+}
+
+var groupAdminAllowedApis = map[string]bool{
+	"GET /api/get-group":               true,
```
dacongda: you can only judge the url by path, make the url too long is not a good idea

object/user.go
dacongda: it seems that we should also filter owner here

authz/authz.go
ANormalDD: this is to maintain consistency with the casbin ruleText format

dacongda:
1. You need to resolve the review when you solve it or comment and resolve it if you don't want to solve it
2. It will be better if you provide a demo video.

authz/authz.go
dacongda: I think there is no need to ensure consistency with the casbin ruleText format.

dacongda: We are just making a very simple judgment

PR #5311, opened 2026-03-23
fix: #5308 read request body directly instead of relying on Beego's RequestBody

### LINKED ISSUE #5308: API update endpoints silently fail to persist fields when Content-Type header is missing
When API update endpoints (e.g. `update-role`) receive a valid JSON body **without** a `Content-Type: application/json` header, certain fields are silently not persisted. The API returns `{"status": "ok", "data": "Affected"}` but the data is not updated.

This affects all official SDK clients that don't explicitly set the Content-Type header (confirmed with the Python SDK which uses `requests.post(data=json_string)`).

## Reproduction

```bash
# WITHOUT Content-Type - groups NOT persisted despite "ok" response:
curl -X POST ".../api/update-role?id=org/role&clientId=...&clientSecret=..." \
  -d '{"owner":"org","name":"role","groups":["org/some-group"],"isEnabled":true}'
# Response: {"status": "ok", "data": "Affected"}
# Re-fetch: groups is still null

# WITH Content-Type - groups persisted:
curl -X POST ".../api/update-role?id=org/role&clientId=...&clientSecret=..." \
  -H "Content-Type: application/json" \
  -d '{"owner":"org","name":"role","groups":["org/some-group"],"isEnabled":true}'
# Response: {"status": "ok", "data": "Affected"}
# Re-fetch: groups is ["org/some-group"] ✓
```

## Root Cause

All controllers use `c.Ctx.Input.RequestBody` for JSON unmarshaling:

```go
err := json.Unmarshal(c.Ctx.Input.RequestBody, &role)
```

Beego's `c.Ctx.Input.RequestBody` is not reliably populated when the request lacks a `Content-Type: application/json` header. The `json.Unmarshal` then operates on an empty or partial body, producing a zero-valued struct. Since the database update uses `.AllCols().Update()`, all fields — including the ones meant to be updated - are overwritten with zero values.

Some fields (like `users` on roles) appear unaffected because they are managed through dedicated endpoints (`assign-role`, `remove-role-from-user`) rather than through `update-role`.

## Suggested Fix

Replace `c.Ctx.Input.RequestBody` with a direct read from the raw HTTP body in all controllers:

```go
// Before:
err := json.Unmarshal(c.Ctx.Input.RequestBody, &role)

// After:
body, err := io.ReadAll(c.Ctx.Request.Body)
if err != nil {
    c.ResponseError(err.Error())
    return
}
err = json.Unmarshal(body, &role)
```

This affects all `controllers/*.go` files that follow this pattern.

Alternatively, a Beego middleware/filter could ensure `RequestBody` is always populated from the raw body regardless of Content-Type.

## Impact

This bug silently corrupts data - the API returns success but writes zero values. Any SDK client that doesn't set `Content-Type: application/json` is affected.

dhilipsiva:
Beego's c.Ctx.Input.RequestBody is not populated when the HTTP request lacks a Content-Type: application/json header. This causes json.Unmarshal to silently produce zero-valued structs, which then get written to the database via .AllCols().Update(), corrupting data.

Reading directly from c.Ctx.Request.Body via io.ReadAll avoids this issue
by not depending on Beego's input parsing behavior.

Fixes #5308

### DIFF
--- a/controllers/cas.go
+++ b/controllers/cas.go
@@ -16,8 +16,8 @@ package controllers
 
 import (
 	"encoding/xml"
-	"io"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"strings"
--- a/controllers/enforcer.go
+++ b/controllers/enforcer.go
@@ -16,8 +16,8 @@ package controllers
 
 import (
 	"encoding/json"
-	"io"
 	"fmt"
+	"io"
 
 	"github.com/beego/beego/v2/core/utils/pagination"
 	"github.com/casdoor/casdoor/object"
--- a/controllers/group.go
+++ b/controllers/group.go
@@ -15,8 +15,8 @@ package controllers
 
 import (
 	"encoding/json"
-	"io"
 	"fmt"
+	"io"
 
 	"github.com/beego/beego/v2/core/utils/pagination"
 	"github.com/casdoor/casdoor/object"
--- a/controllers/order_pay.go
+++ b/controllers/order_pay.go
@@ -16,8 +16,8 @@ package controllers
 
 import (
 	"encoding/json"
-	"io"
 	"fmt"
+	"io"
 
 	"github.com/casdoor/casdoor/object"
 	"github.com/casdoor/casdoor/util"


hsluoyz:
this is a so small and little issue, not worth modifying so many lines of code

PR #5347, opened 2026-03-31
fix: stop auto-filling empty signupApplication with first app

HMBSbige:
When a user's `signupApplication` is empty, the user edit page incorrectly auto-fills it with the first application in the organization, making it appear as if the user registered through that app.

Now empty values are preserved and no longer trigger the auto-fill.

### DIFF
--- a/web/src/UserEditPage.js
+++ b/web/src/UserEditPage.js
@@ -149,18 +149,14 @@ class UserEditPage extends React.Component {
   getApplicationsByOrganization(organizationName) {
     ApplicationBackend.getApplicationsByOrganization("admin", organizationName)
       .then((res) => {
+        const applications = res.data || [];
         this.setState({
-          applications: res.data || [],
+          applications: applications,
         });
 
-        const applications = res.data;
         if (this.state.user) {
-          if (this.state.user.signupApplication === "" || applications.filter(application => application.name === this.state.user.signupApplication).length === 0) {
-            if (applications.length > 0) {
-              this.updateUserField("signupApplication", applications[0].name);
-            } else {
-              this.updateUserField("signupApplication", "");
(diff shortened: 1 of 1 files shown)


HMBSbige:
I've reviewed Copilot's suggestions and don't think either needs to be addressed in this PR — the first contradicts the fix intent, and the second is a pre-existing concern outside this scope. Any suggestions on this PR?

hsluoyz:
@dacongda

web/src/UserEditPage.js
```diff
@@ -149,18 +149,14 @@ class UserEditPage extends React.Component {
...
-        const applications = res.data;
         if (this.state.user) {
-          if (this.state.user.signupApplication === "" || applications.filter(application => application.name === this.state.user.signupApplication).length === 0) {
-            if (applications.length > 0) {
-              this.updateUserField("signupApplication", applications[0].name);
-            } else {
-              this.updateUserField("signupApplication", "");
-            }
+          if (this.state.user.signupApplication && !applications.some(application => application.name === this.state.user.signupApplication)) {
+            this.updateUserField("signupApplication", applications.length > 0 ? applications[0].name : "");
```
dacongda: I think we should fill signupApplication field to empty if the application is not exist

HMBSbige: done

HMBSbige:
Hi @hsluoyz, I noticed this PR was closed without a comment. Could you share the reason? Is the current approach not what you'd like, or is there something I should adjust? Happy to rework it if needed. Thanks!

hsluoyz:
@HMBSbige you can create a github issue to make sure it's a real problem, suitable community member will be assigned to fix it

PR #5385, opened 2026-04-11
fix: prevent non-global-admin from tampering entity owner field

deluxebear:
Add enforceOwnerFromId() helper in controllers that forces the entity's owner to match the id query parameter — but only for non-global-admin users. Global admins can still reassign entities across organizations.

This prevents org admins from transferring entities to other orgs by crafting a request body with a different owner field, while preserving the ability for built-in admins to manage cross-org operations.

### DIFF
--- a/controllers/adapter.go
+++ b/controllers/adapter.go
@@ -102,6 +102,8 @@ func (c *ApiController) UpdateAdapter() {
 		return
 	}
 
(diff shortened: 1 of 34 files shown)


nomeguy:
don't create PR with 34 file changes without getting agreed in the issue for a 2nd time. We also have AI. so if that issue makes sense, we will make code changes by selves

PR #5390, opened 2026-04-12
fix: load dashboard stats correctly when tableNamePrefix is set

KKRainbow:
## Summary
- fix dashboard map lookups to use the prefixed table name key
- add regression coverage for prefixed and unprefixed dashboard map loads

## Root cause
`GetDashboard()` stores query results in `dashboardMap` using `tableNamePrefix + tableName`, but later reads them back using only `tableName`.

When `tableNamePrefix` is set, the lookups miss and all dashboard counts stay at zero even though the underlying data exists.

## Verification
- added regression tests for prefixed and unprefixed dashboard map lookups
- note: `go test ./object -run 'TestGetDashboardMapItem'` currently hits unrelated existing compile failures in `object/permission_rbac_dedup_test.go` on current `master`, so I verified the change with focused logic inspection and the new unit test file

### DIFF
--- a/object/get-dashboard.go
+++ b/object/get-dashboard.go
@@ -96,14 +96,28 @@ func GetDashboard(owner string) (*map[string][]int64, error) {
 	for i := 30; i >= 0; i-- {
 		cutTime := nowTime.AddDate(0, 0, -i)
 		for _, tableName := range tableNames {
-			item, exist := dashboardMap.Load(tableName)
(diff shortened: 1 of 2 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/710f1e1ef843fa14eb2185a90c4c9aee4298ecbf

Issue #5421, opened 2026-04-19
[bug] SendCodeInput does nothing with Dynamic / Internet-Only captcha when 4th arg is missing


HMBSbige:
### Bug

This is a follow-up regression after #5407.

When captcha rule is `Dynamic` or `Internet-Only`, clicking "Send Code" can do nothing at all in pages that call `SendCodeInput` with only 3 `onButtonClickArgs`.

### Reproduction

1. Configure a captcha provider with rule `Dynamic` or `Internet-Only`
2. Open a page that uses `SendCodeInput` with 3 args, for example verification-code login on `LoginPage`
3. Click "Send Code"

### Current behavior

Nothing happens:
- no `get-captcha-status` request
- no `send-verification-code` request
- no error message

### Cause

`SendCodeInput` checks `onButtonClickArgs[3]` before captcha-status and returns early if it is missing, but some callers such as `LoginPage` pass only 3 args again.

This also affects other 3-arg callers such as `SignupPage` and `ResetModal`.

### Suggested fix

Do not early-return when the 4th arg is missing.

For `Dynamic`, captcha-status lookup can use `onButtonClickArgs[3] ?? onButtonClickArgs[0]`.
For `Internet-Only`, the flow should continue without depending on the 4th arg.

hsluoyz:
fixed in: https://github.com/casdoor/casdoor/commit/0682c36190570e6615d85e21867244895d4449d0

PR #5424, opened 2026-04-19
fix: fix check captcha error cause by error if statement

### LINKED ISSUE #5422: [bug] SendCodeInput layout is broken after replacing Input.Search with Input + addonAfter
### Bug

This is a follow-up regression after #5408.

The Enter-submit race seems fixed, but `SendCodeInput` now has a UI regression after replacing `Input.Search` with plain `Input + addonAfter={<Button />}`.

### Current behavior

The "Send Code" control looks visually broken:
- nested / double-border effect
- broken integrated input-button layout
- disabled button looks misaligned inside the input group

### Cause

`SendCodeInput` switched from `Input.Search + enterButton` to plain `Input + addonAfter={<Button />}`.

### Suggested fix

Use `Space.Compact` with `Input + Button` instead of plain `Input + addonAfter`.

dacongda:
fix: #5422

### DIFF
--- a/web/src/common/SendCodeInput.js
+++ b/web/src/common/SendCodeInput.js
@@ -90,12 +90,9 @@ export const SendCodeInput = ({value, disabled, captchaValue, useInlineCaptcha, 
     };
 
     const checkCaptchaStatusAndSend = () => {
-      if (!onButtonClickArgs?.[3]) {
-        return;
-      }
       const values = {
         organization: application?.organization,
-        username: onButtonClickArgs?.[3],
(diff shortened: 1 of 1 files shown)


HMBSbige (commented):
This looks like a partial fix for #5421.

It should fix the login 3-arg callers, but `ResetModal` may still be inconsistent: the frontend falls back to `onButtonClickArgs[0]`, while the backend checks reset CAPTCHA status against the current logged-in user.

hsluoyz:
let me fix this issue

PR #5436, opened 2026-04-22
fix: use exclusive C14N for SAML assertion signatures

### LINKED ISSUE #5433: Casdoor SAML C14N Regression: Assertion Signature Broken for SPs Requiring Exclusive C14N
## Environment

| | Version |
|---|---|
| **Known good** | v2.285.0 |
| **Broken** | v2.400.0 → latest (v3.36.0) |
| **Service Provider** | HashiCorp Cloud Platform (HCP), Auth0-backed SAML consumer |

---

### With `EnableSamlC14n10 = true`
```
Signature check errors: invalid signature: for uri #_32cd4317-859f-4296-81c5-96509598fb9a
calculated digest is ZbUeVLROG6SD6r4QBR5lDmSirlozWIlhZ9R7wLKHoRA=
but the xml to validate supplies digest AFSd/3h9iDbrud/skwGWZDa4pCwkV52B83+W3kqyMMU=
```

### With `EnableSamlC14n10 = false`
```
canonicalization algorithm 'http://www.w3.org/2006/12/xml-c14n11' is not supported
```

---

## Root Cause

The bug is in the **assertion signing code** in `object/saml_idp.go`. The `EnableSamlC14n10` toggle — introduced somewhere between v2.285.0 and v2.400.0 — offers only two algorithm options, neither of which is correct for broad SP compatibility:

| `EnableSamlC14n10` | Algorithm used | URI | Result |
|---|---|---|---|
| `true` | Inclusive C14N 1.0 | `http://www.w3.org/TR/2001/REC-xml-c14n-20010315` | ❌ Digest mismatch |
| `false` | C14N 1.1 | `http://www.w3.org/2006/12/xml-c14n11` | ❌ Algorithm not supported |

The correct algorithm for SAML assertion signing — as specified in the [OASIS SAML 2.0 metadata spec](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) and required by virtually all major SPs — is **Exclusive C14N**:

```
http://www.w3.org/2001/10/xml-exc-c14n#
```

This is already what Casdoor's own dependency `goxmldsig` uses for response-level signing. It is also what Auth0, Okta, Azure AD, and AWS all require.

### Why Inclusive C14N Causes a Digest Mismatch

Inclusive C14N propagates **all ancestor namespace declarations** into the canonicalized form of the signed element. When the SP (HCP/Auth0) extracts the `<Assertion>` element from the `<Response>` to verify its signature in isolation, the namespace context changes — so re-canonicalizing it produces a different byte sequence than what Casdoor originally signed.

Exclusive C14N was designed specifically to solve this: it only includes namespaces **actually used** within the element being signed, making the digest stable regardless of where the element appears in the document.

---

## Why v2.285.0 Worked

In v2.285.0, the assertion signing code very likely used `goxmldsig`'s `SigningContext` directly (which hardcodes Exclusive C14N), or the C14N selection logic did not yet exist. When the `EnableSamlC14n10` toggle was added, the `false` branch was incorrectly assigned C14N 1.1 instead of preserving Exclusive C14N.

---

## Proposed Fix

In `object/saml_idp.go`, the canonicalization algorithm for assertion signing should be hardcoded to Exclusive C14N, regardless of the toggle state:

```go
// Both <CanonicalizationMethod> on <SignedInfo>
// and <Transform> on <Reference> must use this URI:
const canonicalizationAlgorithm = "http://www.w3.org/2001/10/xml-exc-c14n#"
```

The `EnableSamlC14n10` toggle should either be **removed**, or if retained for legacy reasons, both branches should resolve to Exclusive C14N. C14N 1.1 (`http://www.w3.org/2006/12/xml-c14n11`) is not widely supported by SAML SPs and should not be used as a default.

---

## Reproduction Steps

1. Deploy Casdoor v2.400.0+ as a SAML IdP
2. Enable `EnableSamlAssertionSignature = true` on the application
3. Configure any SP that validates signatures using Exclusive C14N (HCP, Okta, Auth0-backed SPs, Azure AD)
4. Attempt SSO login

**Result with `EnableSamlC14n10 = true`:** digest mismatch error on ACS callback  
**Result with `EnableSamlC14n10 = false`:** "canonicalization algorithm not supported" error

---

## References

- [OASIS SAML 2.0 Metadata Spec](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) — signatures SHOULD use `http://www.w3.org/2001/10/xml-exc-c14n#` or `http://www.w3.org/2001/10/xml-exc-c14n#WithComments`
- [`goxmldsig`](https://github.com/russellhaering/goxmldsig) — Casdoor's own XML signing library, uses Exclusive C14N by default
- [Casdoor `object/saml_idp.go`](https://github.com/casdoor/casdoor/blob/master/object/saml_idp.go)
- Casdoor docs: [SAML IdP overview](https://casdoor.org/docs/how-to-connect/saml/overview/) — notes that `EnableSamlAssertionSignature` was added in v2.81.0

## Notes

I am currently running version 2.400.0 in production, and cannot upgrade to version 3.x at the moment; would it be possible to apply a fix in both versions 2 and 3?

Thanks!

spphashicorp:
Fixes #5433.

Summary
- Keep the `EnableSamlC14n10` toggle in place.
- Restore the `EnableSamlC14n10=true` branch to the known-good v2.285.0 behavior: exclusive C14N using `dsig.MakeC14N10ExclusiveCanonicalizerWithPrefixList("")`.
- Preserve the `EnableSamlC14n10=false` path by continuing to use goxmldsig's direct `SigningContext` default, which is C14N11.
- Add a regression test that verifies both branches explicitly: disabled emits C14N11, enabled emits exclusive C14N, and the enabled branch does not emit `InclusiveNamespaces` or `PrefixList="xs"`.

Why
- The regression reported in #5433 is specifically in the enabled branch: newer releases changed the exclusive C14N branch shape from the v2.285.0 implementation to an incompatible variant.
- Preserving the flag avoids changing existing configuration semantics while still restoring the working exclusive-C14N path for users who enable it.

Validation
- Ran a standalone probe against the production helper and confirmed:
  - `EnableSamlC14n10=false` emits `http://www.w3.org/2006/12/xml-c14n11`
  - `EnableSamlC14n10=true` emits `http://www.w3.org/2001/10/xml-exc-c14n#`
  - the enabled branch does not emit `InclusiveNamespaces` or `PrefixList="xs"`
- Confirmed the edited files have no local compile errors.

Notes
- This PR branch is pushed from a fork because the upstream repository does not grant direct push access to this account.

### DIFF
--- a/object/saml_idp.go
+++ b/object/saml_idp.go
@@ -161,7 +161,9 @@ func NewSamlSigningContext(application *Application, randomKeyStore *X509Key) *d
 		ctx.Hash = crypto.SHA512
 	}
 
-	ctx.Canonicalizer = dsig.MakeC14N10ExclusiveCanonicalizerWithPrefixList("")
+	if application.EnableSamlC14n10 {
+		ctx.Canonicalizer = dsig.MakeC14N10ExclusiveCanonicalizerWithPrefixList("")
+	}
 
 	return ctx
 }
--- a/object/saml_idp.go
+++ b/object/saml_idp.go
@@ -151,6 +151,21 @@ func NewSamlResponse(application *Application, user *User, host string, certific
 	return samlResponse, nil
 }
 
+func NewSamlSigningContext(application *Application, randomKeyStore *X509Key) *dsig.SigningContext {
+	ctx := dsig.NewDefaultSigningContext(randomKeyStore)
+	if application.SamlHashAlgorithm == "" || application.SamlHashAlgorithm == "SHA1" {
+		ctx.Hash = crypto.SHA1
+	} else if application.SamlHashAlgorithm == "SHA256" {
+		ctx.Hash = crypto.SHA256
+	} else if application.SamlHashAlgorithm == "SHA512" {
+		ctx.Hash = crypto.SHA512
+	}
+
+	ctx.Canonicalizer = dsig.MakeC14N10ExclusiveCanonicalizerWithPrefixList("")
+
+	return ctx
+}
+
 type X509Key struct {
 	X509Certificate string
 	PrivateKey      string
@@ -375,18 +390,7 @@ func GetSamlResponse(application *Application, user *User, samlRequest string, h
 		PrivateKey:      cert.PrivateKey,
 		X509Certificate: certificate,
 	}
-	ctx := dsig.NewDefaultSigningContext(randomKeyStore)
-	if application.SamlHashAlgorithm == "" || application.SamlHashAlgorithm == "SHA1" {
-		ctx.Hash = crypto.SHA1
-	} else if application.SamlHashAlgorithm == "SHA256" {
-		ctx.Hash = crypto.SHA256
-	} else if application.SamlHashAlgorithm == "SHA512" {
-		ctx.Hash = crypto.SHA512
-	}
-
-	if application.EnableSamlC14n10 {
-		ctx.Canonicalizer = dsig.MakeC14N10ExclusiveCanonicalizerWithPrefixList("xs")
-	}
+	ctx := NewSamlSigningContext(application, randomKeyStore)
 
 	// signedXML, err := ctx.SignEnvelopedLimix(samlResponse)
 	// if err != nil {
--- a/object/saml_idp_test.go
+++ b/object/saml_idp_test.go
@@ -1,0 +1,81 @@
+// Copyright 2026 The Casdoor Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package object
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/pem"
+	"testing"
+
+	"github.com/beevik/etree"
+	"github.com/stretchr/testify/require"
+)
+
+func testSamlSigningKeyStore(t *testing.T) *X509Key {
+	t.Helper()
+
+	privateKey, err := rsa.GenerateKey(rand.Reader, 1024)
+	require.NoError(t, err)
+
+	privateKeyPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(privateKey),
+	})
+
+	return &X509Key{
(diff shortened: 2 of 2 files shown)


object/saml_idp.go
```diff
@@ -151,6 +151,21 @@ func NewSamlResponse(application *Application, user *User, host string, certific
...
+	ctx := dsig.NewDefaultSigningContext(randomKeyStore)
+	if application.SamlHashAlgorithm == "" || application.SamlHashAlgorithm == "SHA1" {
+		ctx.Hash = crypto.SHA1
+	} else if application.SamlHashAlgorithm == "SHA256" {
+		ctx.Hash = crypto.SHA256
+	} else if application.SamlHashAlgorithm == "SHA512" {
+		ctx.Hash = crypto.SHA512
+	}
+
+	ctx.Canonicalizer = dsig.MakeC14N10ExclusiveCanonicalizerWithPrefixList("")
```
dacongda: remove application.EnableSamlC14n10 may cause bug, this is not a good idea.

dacongda (changes requested):

object/saml_idp.go
spphashicorp: As I highlighted in the issue I raised: https://github.com/casdoor/casdoor/issues/5433 no major provider is using C14N11, so i'm a little confused why that architectural choice was made in the first place.

But that's a fair point; I will take another look today, and push an update.

spphashicorp: Brought back the toggle in https://github.com/casdoor/casdoor/pull/5436/commits/d0376a25a25767832057c818e7c815b2e3ac58b9 ; also updated the tests to check both.

Would it be possible to also backport this fix to the 2.x version, and create a 2.405.0 tag? I am currently using 2.400.0 in my production environment, and cannot upgrade to version 3.x...

dacongda: I can't decide this

dacongda: In fact, v3 and v2 hasn't too much different, you can switch v2 to v3 smoothly.

dacongda: It seems that you only change the MakeC14N10ExclusiveCanonicalizerWithPrefixList, I think it is not essential to separate these to a new function, just keep the original style and replace "xs" to ""

spphashicorp: @dacongda regarding v3, are there any significant database schema changes? this is what I am concerned with. Otherwise, I would be happy to upgrade :) 

Also, regarding the refactor, the reason I created a separate function is to be able to add a test for it, is that ok?

dacongda: no。database changes，and no need to add。test

hsluoyz:
@spphashicorp don't do refactoring, and inside it, you changed `dsig.MakeC14N10ExclusiveCanonicalizerWithPrefixList("xs")`, removed the "xs", you should know that we added it deliberately before. This is the key issue. Don't try to hide this behavior with refactoring and some new go tests

spphashicorp:
@hsluoyz so what is the solution ? version 2.400.0 and every version after that simply no longer works for consumers requiring C14N10. I tested 3.36.0 and the issue is still there.

PR #5460, opened 2026-05-11
fix: global admin exit impersonation fails for cross-org users

### LINKED ISSUE #5459: Global admin cannot exit impersonation when impersonating a user in a different organization
## Describe the bug

When a global admin (`built-in/admin`) impersonates a user in a different organization (e.g., `myorg/someone`), the `/api/exit-impersonate-user` endpoint fails with `"Unauthorized operation"`.

## Root Cause

There are two issues in `getImpersonateUser` in `routers/authz_filter.go`:

1. **Overly restrictive exit condition** — The condition `user.IsAdmin && impUserOwner == user.Owner` prevents global admins from exiting impersonation when the impersonated user belongs to a different organization, even though global admins should have cross-org access.

2. **Wrong identity used for authz on exit** — The impersonated user's identity is always substituted as the request subject. For `/api/exit-impersonate-user`, the authz check should use the **real admin's** identity, not the impersonated user's. Otherwise, the check fails because a regular user lacks permission to exit impersonation.

## Steps to reproduce

1. Log in as `built-in/admin`
2. Impersonate a user in a non-built-in organization (e.g., `myorg/testuser`)
3. Call `POST /api/exit-impersonate-user`
4. Observe: `result = deny` and response `"Unauthorized operation"`

## Expected behavior

Global admins should be able to exit impersonation regardless of which organization the impersonated user belongs to.

Blackcbears:
fix: https://github.com/casdoor/casdoor/issues/5459

## Summary

Fixes two issues that prevent global admins from exiting impersonation when the impersonated user belongs to a different organization.

### 1. Allow global admins to exit cross-org impersonation

The exit condition in `getImpersonateUser` now checks `user.IsGlobalAdmin()` in addition to `impUserOwner == user.Owner`. This allows global admins (`built-in/*`) to exit impersonation regardless of the impersonated user's organization.

### 2. Use real admin identity for authz on exit endpoint

When calling `/api/exit-impersonate-user`, the authz filter now uses the real admin's identity as the request subject instead of the impersonated user's identity. The `"impersonating"` flag is still set so the controller can verify the impersonation state.

### Implementation

- `getImpersonateUser` accepts a `urlPath` parameter
- For `/api/exit-impersonate-user`, returns the original (admin) subject identity while still setting `"impersonating": true`
- `ApiFilter` computes `urlPath` earlier and passes it through

## Affected files

- `routers/authz_filter.go`

## Before / After

| Scenario | Before | After |
|---|---|---|
| Global admin exits impersonation (same org) | ✅ Allow | ✅ Allow |
| Global admin exits impersonation (cross-org) | ❌ Deny | ✅ Allow |
| Org admin exits impersonation (same org) | ✅ Allow | ✅ Allow |

### DIFF
--- a/routers/authz_filter.go
+++ b/routers/authz_filter.go
@@ -294,7 +294,7 @@ func getExtraInfo(ctx *context.Context, urlPath string) map[string]interface{} {
 	return extra
 }
 
-func getImpersonateUser(ctx *context.Context, subOwner, subName, username string) (string, string, string) {
+func getImpersonateUser(ctx *context.Context, subOwner, subName, username, urlPath string) (string, string, string) {
 	impersonateUser, ok := ctx.Input.Session("impersonateUser").(string)
 	impersonateUserCookie := ctx.GetCookie("impersonateUser")
 	if ok && impersonateUser != "" && impersonateUserCookie != "" {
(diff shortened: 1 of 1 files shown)


Blackcbears:
#5459

PR #5485, opened 2026-05-18
fix: remove localhost and `127.0.0.1` from hardcoded CORS allowlist

Danipulok:
**Previous PR:** [#5463](https://github.com/casdoor/casdoor/pull/5463) (closed with "not planned" — no technical reason given, resubmitting with full analysis)

## Summary

`util/validation.go:130` unconditionally trusts `http://localhost`, `https://localhost`, and `http://127.0.0.1` as valid CORS origins. The CORS filter at `cors_filter.go:55-63` checks this **before** any other validation — if it passes, the origin is reflected with `Access-Control-Allow-Credentials: true`.

In production, any process listening on a user's localhost (browser extension, local dev server, malware, Electron app) can make **authenticated cross-origin requests** to the Casdoor API on behalf of that user — read their profile, change their password, enumerate other users — because the browser attaches session cookies.

**CWE-942** (Permissive Cross-domain Policy with Untrusted Domains) / **OWASP A05:2021** (Security Misconfiguration)

### Attack scenario

1. User is logged into Casdoor (session cookie set for `auth.example.com`)
2. User visits `http://localhost:3000` (any local app — dev tool, Electron app, malicious page)
3. Page sends `fetch("https://auth.example.com/api/get-account", {credentials: "include"})` with `Origin: http://localhost:3000`
4. Casdoor reflects the origin with `Access-Control-Allow-Credentials: true`
5. Browser allows the response — attacker reads the user's full account data

### `util/validation.go` — `IsValidOrigin()`

Removed `http://localhost`, `https://localhost`, and `http://127.0.0.1` from the hardcoded allowlist. Kept `http://casdoor-authenticator` and `*.chromiumapp.org` (browser extension — legitimate use case).

### Why dev environments still work

`IsValidOrigin()` is called in 3 places. In all three, removing localhost is safe because subsequent checks already handle development scenarios:

1. **CORS filter** (`cors_filter.go:86-105`) — falls through to hostname matching (`originHostname == host`) and intranet detection (`IsHostIntranet`). Dev servers on localhost pass both checks.
2. **OAuth redirect URI** (`application_util.go:368`) — falls through to the application's configured `RedirectUris` list. Developers who need `http://localhost:PORT` as a redirect URI add it to their app config explicitly.
3. **Application origin** (`application_util.go:515`) — same fallthrough to configured redirect URIs.

### Frontend impact

None. The frontend uses relative URLs in production (same-origin, no CORS needed). In dev, the craco proxy handles API requests with `changeOrigin: true` — the backend never sees a cross-origin localhost request.

## Verification

- `go build ./...` — passes
- Dev environment CORS works via hostname matching and intranet detection (no regression)
- Production deployments that explicitly need localhost CORS can add `http://localhost:PORT` to the application's redirect URIs

### DIFF
--- a/util/validation.go
+++ b/util/validation.go
@@ -127,6 +127,6 @@ func IsValidOrigin(origin string) (bool, error) {
 		originHostOnly = fmt.Sprintf("%s://%s", urlObj.Scheme, urlObj.Hostname())
 	}
 
-	res := originHostOnly == "http://localhost" || originHostOnly == "https://localhost" || originHostOnly == "http://127.0.0.1" || originHostOnly == "http://casdoor-authenticator" || strings.HasSuffix(originHostOnly, ".chromiumapp.org")
+	res := originHostOnly == "http://casdoor-authenticator" || strings.HasSuffix(originHostOnly, ".chromiumapp.org")
 	return res, nil
 }


nomeguy:
closed as false positive

Danipulok:
@nomeguy hey!
Could you please clarify how exactly is this false positive?

PR #5495, opened 2026-05-21
fix: show user group membership from group side

### LINKED ISSUE #5467: [bug] Group members are not visible from Group page after assigning user to group
When a user is assigned to a group, the relationship is only visible from the User page, but not from the Group page.

Currently:

On the User page, the assigned groups can be seen correctly.
On the Group page, the corresponding users are not displayed.

This makes the relationship appear one-way in the UI.

Expected Behavior

The user-group relationship should be visible from both directions:

User → Groups
Group → Users

LegendPei:
## What

Fixes #5467.

This PR makes the user-group relationship visible from the Group side even when the relationship exists in `User.Groups` but is missing from the Casbin user-group policy table.

## How to Reproduce

1. Assign a user to a group from the User edit page.
2. Open the same user page.
3. Confirm the assigned group is displayed correctly in the user's Groups field.
4. Open the corresponding Group page or Group list.
5. The assigned user may not be displayed under that group.

This makes the relationship appear one-way in the UI:

- User -> Groups works
- Group -> Users does not always work

## Root Cause

The User page reads group membership from the `User.Groups` field.

The Group page reads users through the Casbin user-group enforcer. If `User.Groups` and the Casbin policy table are out of sync, the Group page cannot find users that are still visible from the User page.

## Solution

Group-side membership lookup now merges both relationship sources:

- users returned by the Casbin user-group enforcer
- users whose `User.Groups` field contains the target group

The merged result is de-duplicated and used by:

- `GetGroupUsers()`
- `GetGroupUserCount()`
- `GetPaginationGroupUsers()`
- `ExtendGroupWithUsers()`
- `ExtendGroupsWithUsers()`

## Tests

Added regression tests covering users assigned through `User.Groups` without a corresponding Casbin policy.

```bash
go test ./object -run "Test(GetGroupUsersIncludesUsersAssignedByUserGroups|ExtendGroupWithUsersIncludesUsersAssignedByUserGroups)" -count=1

### DIFF
--- a/object/group.go
+++ b/object/group.go
@@ -315,9 +315,12 @@ func GetGroupUserCount(groupId string, field, value string) (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	names, err := userEnforcer.GetUserNamesByGroupName(groupId)
+	names, err := getGroupUserNames(groupId)
 	if err != nil {
 		return 0, err
+	}
+	if len(names) == 0 {
+		return 0, nil
 	}
 
 	if field == "" && value == "" {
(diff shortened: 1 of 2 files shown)


LegendPei:
@hsluoyz why do you close the PR?Did I do something wrong?

Issue #3691, opened 2025-03-30
[CORS Issue] Using casdoor-js-sdk meet CORS issues


1834423612:
Hi, 
While I was trying to test my front-end connection with the Casdoor server, I encountered a CORS issue when logging in.
The website front-end was running on my local (Vue3+TS), and the Casdoor server was running on a remote server.

Below is the details & version.
```
Browser: Chrome 134.0.0.0
OS: Windows 11 24H2 x86_64 26100.3476
```
Thank you so much!!!

----  
**OPTION works:**

![Image](https://github.com/user-attachments/assets/40eadbba-2416-408e-b464-ce395fbbb96a)

**fetch Error:** 
Cross-domain resource sharing error: PreflightMissingAllowOriginHeader
![Image](https://github.com/user-attachments/assets/53c17400-1ff8-48d2-8612-af87a202dff9)

Nginx Reverse proxy Config:

![Image](https://github.com/user-attachments/assets/b15e97e2-2186-4b92-b491-8e127628efd2)
```
# Reverse proxy Config
    location ^~ / {
        proxy_pass http://localhost:8001;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header REMOTE-HOST $remote_addr;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection $connection_upgrade;
        proxy_http_version 1.1;
    
        # CORS config
        add_header 'Access-Control-Allow-Origin' '*';
        add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS, PUT, DELETE, PATCH';
        add_header 'Access-Control-Allow-Headers' 'DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization';
        if ($request_method = 'OPTIONS') {
            add_header 'Access-Control-Max-Age' 1728000;
            add_header 'Content-Type' 'text/plain charset=UTF-8';
            add_header 'Content-Length' 0;
            return 204;
        }
    }
```

1834423612:
When I access this casdoor api, I find that it opens a page like the following. Do I also need to add CORS rules to the casdoor sever config file separately?

Thanks

![Image](https://github.com/user-attachments/assets/fe508642-da56-4487-8eef-89ca83022f7f)

[...]

PR #5509, opened 2026-05-24
feat: make CORS `Access-Control-Allow-Headers` configurable

### LINKED ISSUE #5508: [Feature] Make CORS `Access-Control-Allow-Headers` configurable via `app.conf`
# [Feature] Make CORS `Access-Control-Allow-Headers` configurable via `app.conf`

## Problem

`Access-Control-Allow-Headers` in `routers/cors_filter.go:37` is hardcoded:

```go
ctx.Output.Header(headerAllowHeaders, "Content-Type, Authorization")
```

Any cross-origin request that includes a non-safelisted header outside `Content-Type` / `Authorization` triggers a CORS preflight that the browser rejects, because Casdoor's `OPTIONS` response doesn't list the header in `Access-Control-Allow-Headers`.

This blocks legitimate integrations and forces every team to choose between:
- Wrapping Casdoor behind a reverse proxy that rewrites the headers (the workaround in #3691).
- Patching client code to strip the headers (the workaround we ended up using for Swagger UI).
- Forking Casdoor.

## Reproduction

The clearest case is **Swagger UI calling the OIDC token endpoint** during an `authorizationCode` flow:

**Stack:** FastAPI backend, mounted Swagger UI with `OAuth2AuthorizationCodeBearer` + `usePkceWithAuthorizationCodeGrant: true`, Casdoor as OIDC provider.

**Setup:**

1. `init_data.json` has an application with `clientId: "demo"` and `redirectUris: ["http://localhost:8000/api/v1/docs/oauth2-redirect"]`.
2. The FastAPI security scheme is configured from Casdoor's `.well-known/openid-configuration`:
   - `authorization_endpoint: http://localhost:8001/login/oauth/authorize`
   - `token_endpoint: http://localhost:8001/api/login/oauth/access_token`
3. Browser is on `http://localhost:8000/api/v1/docs`.

**Steps:**

1. Click **Authorize** in Swagger UI → enter `clientId` → submit.
2. Browser redirects to Casdoor `/login/oauth/authorize`, you sign in.
3. Casdoor redirects back to `…/docs/oauth2-redirect?code=…`.
4. Swagger UI tries to `POST http://localhost:8001/api/login/oauth/access_token` with PKCE `code_verifier`.

**Expected:** Token returned, Swagger UI saves Bearer token.

**Actual:**

Browser console:
```
Access to fetch at 'http://localhost:8001/api/login/oauth/access_token' from origin
'http://localhost:8000' has been blocked by CORS policy: Request header field
x-requested-with is not allowed by Access-Control-Allow-Headers in preflight response.
```

The `OPTIONS` preflight succeeds with HTTP 200 (verified via curl), Casdoor returns:
```
Access-Control-Allow-Origin: http://localhost:8000
Access-Control-Allow-Headers: Content-Type, Authorization
Access-Control-Allow-Credentials: true
```

`X-Requested-With: XMLHttpRequest` is added by Swagger UI's bundled fetch client (a long-standing legacy behavior that cannot be disabled from configuration). The browser blocks the actual POST because `X-Requested-With` isn't in the allow-list.

## Why this is broader than Swagger UI

Many HTTP clients send `X-Requested-With` automatically:

- Swagger UI 3.x and 5.x (bundled inside FastAPI, Drf-Spectacular, Spring Boot's Swagger integration, etc.).
- jQuery `$.ajax`.
- Axios with certain interceptors.
- Older `XMLHttpRequest`-based code.

Other common non-safelisted headers that integrators want to add:
- `X-CSRF-Token` (when fronting Casdoor with apps that use CSRF tokens).
- `X-Trace-Id` / `X-Request-Id` (observability stacks).
- `If-Modified-Since`, `Cache-Control` (caching layers — see the nginx workaround in #3691).

Today every integrator has to either run Casdoor behind a reverse proxy or patch their client. A one-line config would solve it for everyone.

## Proposed solution

Add an optional `corsAllowHeaders` key to `app.conf` with the current value as default. Read it in `setCorsHeaders`.

**`conf/app.conf`** — add:
```ini
# Comma-separated list of headers added to Access-Control-Allow-Headers.
# Default: "Content-Type, Authorization"
corsAllowHeaders = "Content-Type, Authorization"
```

**`routers/cors_filter.go`** — change `setCorsHeaders`:

```diff
 func setCorsHeaders(ctx *context.Context, origin string) {
     ctx.Output.Header(headerAllowOrigin, origin)
     ctx.Output.Header(headerAllowMethods, "POST, GET, OPTIONS, DELETE")
-    ctx.Output.Header(headerAllowHeaders, "Content-Type, Authorization")
+    allowHeaders := conf.GetConfigString("corsAllowHeaders")
+    if allowHeaders == "" {
+        allowHeaders = "Content-Type, Authorization"
+    }
+    ctx.Output.Header(headerAllowHeaders, allowHeaders)
     ctx.Output.Header(headerAllowCredentials, "true")

     if ctx.Input.Method() == "OPTIONS" {
         ctx.ResponseWriter.WriteHeader(http.StatusOK)
     }
 }
```

That preserves backward compatibility (existing deployments behave exactly as before) and lets operators opt in to whatever set of headers their stack needs.

## Same idea, optionally bigger

If the maintainers prefer a richer config surface, the same pattern could expose:
- `corsAllowMethods` (also hardcoded to `POST, GET, OPTIONS, DELETE`).
- `corsAllowCredentials` (`true` hardcoded).
- `corsExposeHeaders` (currently not sent at all).

Happy to send a PR for the minimal `corsAllowHeaders` change first if the direction is acceptable.

## Related

- #3691 — nginx-level workaround that adds `X-Requested-With` and others to the allow-list.
- #1408 — earlier CORS report; closed without addressing allow-headers.
- swagger-api/swagger-ui#6081 — long-standing Swagger UI CORS issue, downstream of every OAuth provider that doesn't accept `X-Requested-With`.

## Environment

- Casdoor `latest` Docker image (verified against current `master` `routers/cors_filter.go`).
- Browsers: Chrome 134+, Firefox 124+, Safari 17+ (all enforce preflight strictly per CORS spec).
- Use case: any OIDC integration whose client library adds non-safelisted headers — confirmed with FastAPI's bundled Swagger UI but applies to any HTTP client emitting `X-Requested-With` or similar.

Neko1313:
## What

Adds an optional `corsAllowHeaders` key to `app.conf`. When set, its value replaces the previously hardcoded `Content-Type, Authorization` advertised in `Access-Control-Allow-Headers`. When unset, the original default is preserved — existing deployments behave identically.

## Why

Several common HTTP clients add headers that browsers treat as non-CORS-safelisted, triggering preflight `OPTIONS` requests. Casdoor's response currently lists only `Content-Type, Authorization`, so the browser rejects the actual request. Today operators have to either fork or wrap Casdoor behind a reverse proxy that rewrites the headers.

Concrete cases reported:

- **Swagger UI (FastAPI / Spring Boot)** adds `X-Requested-With: XMLHttpRequest` to every fetch. The OIDC `authorization_code` token POST fails with:
  ```
  Access to fetch at 'http://localhost:8001/api/login/oauth/access_token'
  has been blocked by CORS policy: Request header field x-requested-with
  is not allowed by Access-Control-Allow-Headers in preflight response.
  ```
  Full repro and context in #5508.
- **CSRF-protected frontends** want to forward `X-CSRF-Token`.
- **Observability stacks** want `X-Trace-Id` / `X-Request-Id` echoed through.
- **#3691** documents a community nginx workaround that adds a much wider header set.

## How

`routers/cors_filter.go` gets a tiny helper that reads `corsAllowHeaders` from config and falls back to the historical default:

```go
const defaultCorsAllowHeaders = \"Content-Type, Authorization\"

func getCorsAllowHeaders() string {
    configured := conf.GetConfigString(\"corsAllowHeaders\")
    if configured == \"\" {
        return defaultCorsAllowHeaders
    }
    return configured
}
```

`setCorsHeaders` calls it instead of using the literal. `conf/app.conf` gets an empty `corsAllowHeaders =` placeholder so operators discover the option.

## Backward compatibility

Drop-in. Without `corsAllowHeaders` in `app.conf`, Casdoor emits exactly the same `Access-Control-Allow-Headers` value as before this change.

## Example usage

Operators wanting to allow Swagger UI's OIDC flow plus CSRF tokens add:

\`\`\`ini
corsAllowHeaders = \"Content-Type, Authorization, X-Requested-With, X-CSRF-Token\"
\`\`\`

## Notes

- No new dependencies.
- Same pattern could later expose `corsAllowMethods` and `corsAllowCredentials`; happy to follow up in a separate PR if maintainers want those too.
- No new tests — \`routers/\` package currently ships without unit tests; adding test harness for one helper felt out of scope. Manual verification: \`go build ./routers/...\` and \`go vet ./routers/...\` pass.

Closes #5508

### DIFF
--- a/conf/app.conf
+++ b/conf/app.conf
@@ -18,6 +18,7 @@ logPostOnly = true
 isUsernameLowered = false
 origin =
 originFrontend =
+corsAllowHeaders =
 staticBaseUrl = "https://cdn.casbin.org"
 isDemoMode = false
 batchSize = 100
--- a/routers/cors_filter.go
+++ b/routers/cors_filter.go
@@ -29,12 +29,28 @@ const (
 	headerAllowMethods     = "Access-Control-Allow-Methods"
 	headerAllowHeaders     = "Access-Control-Allow-Headers"
 	headerAllowCredentials = "Access-Control-Allow-Credentials"
+
+	defaultCorsAllowHeaders = "Content-Type, Authorization"
 )
+
+// getCorsAllowHeaders returns the value to advertise in the
+// Access-Control-Allow-Headers response header. The list is read from the
+// optional "corsAllowHeaders" key in app.conf so deployments can permit
+// additional non-safelisted headers (e.g. X-Requested-With sent by Swagger
+// UI, X-CSRF-Token, X-Trace-Id) without forking. Empty / missing config
+// preserves the historical default.
+func getCorsAllowHeaders() string {
+	configured := conf.GetConfigString("corsAllowHeaders")
+	if configured == "" {
+		return defaultCorsAllowHeaders
+	}
+	return configured
+}
 
 func setCorsHeaders(ctx *context.Context, origin string) {
 	ctx.Output.Header(headerAllowOrigin, origin)
 	ctx.Output.Header(headerAllowMethods, "POST, GET, OPTIONS, DELETE")
-	ctx.Output.Header(headerAllowHeaders, "Content-Type, Authorization")
+	ctx.Output.Header(headerAllowHeaders, getCorsAllowHeaders())
 	ctx.Output.Header(headerAllowCredentials, "true")
 
 	if ctx.Input.Method() == "OPTIONS" {


Neko1313:
@hsluoyz Can you check this out?

nomeguy:
won't fix, modify source code by yourself is much easier

PR #5556, opened 2026-06-05
feat: respect Provider.customLogo for all types, not just Custom*

### LINKED ISSUE #5555: [bug] customLogo on Provider is silently ignored for non-Custom OAuth types (Google, MicrosoftOnline, etc.)
The `Provider.customLogo` field can be set via the admin UI and the REST API for any provider type, but the login-page renderer only respects it when `provider.type` starts with `"Custom"` (e.g. `CustomHTTP`, `CustomOAuth2`). For all standard OAuth providers (`Google`, `MicrosoftOnline`, `GitHub`, `Apple`, ...) the field is silently dropped and the icon URL is always built as `${StaticBaseUrl}/img/social_<type>.png` — pulling from `cdn.casbin.org` by default.

This has two practical consequences:

**1. GDPR / DSGVO problem in the EU.** Every social-signin-button render on a self-hosted Casdoor login page triggers a request from the end-user's browser to `cdn.casbin.org`. Same Munich Regional Court reasoning (LG München I, 3 O 17493/20, 2022) that motivated #5547 / PR #5550 (Inter font from Google Fonts) applies here. Setting `customLogo` is the natural opt-out — and the admin UI / docs imply it works — but it doesn't.

**2. Inconsistent escape-hatch semantics.** The field is named `customLogo`, not `customLogoForCustomTypeOnly`. The admin UI offers it on every provider edit page (including Google/Microsoft). Current behaviour surprises operators who set the field expecting it to take effect.

**Reproduction:**

1. Edit an existing Google provider (`type=Google`, `category=OAuth`) in the admin UI.
2. Set `customLogo` to any URL, e.g. `https://example.org/google.svg`. Save.
3. Open the login page of any application that exposes the Google IdP.
4. Browser network tab: still loads `${StaticBaseUrl}/img/social_google.png`, **not** the value of `customLogo`.

**Source — `web/src/Setting.js` (current `master`):**

```javascript
export function getProviderLogoURL(provider) {
  if (provider.type.startsWith("Custom") && provider.customLogo) {
    return provider.customLogo;
  }
  if (provider.category === "OAuth") {
    const type = provider.type.startsWith("Custom") ? "Custom" : provider.type;
    return `${StaticBaseUrl}/img/social_${type.toLowerCase()}.png`;
  } else {
    const info = OtherProviderInfo[provider.category][provider.type];
    if (info) {
      return info.logo;
    }
    return "";
  }
}
```

The `provider.type.startsWith("Custom")` guard on the first `if` makes `customLogo` work only for Custom* types. Removing the guard makes `customLogo` a universal override; fallback chain stays intact for providers without it.

**Suggested fix:**

```diff
 export function getProviderLogoURL(provider) {
-  if (provider.type.startsWith("Custom") && provider.customLogo) {
+  if (provider.customLogo) {
     return provider.customLogo;
   }
   if (provider.category === "OAuth") {
```

One-line change, fully backward-compatible: behaviour for providers without `customLogo` is identical (still falls back to `${StaticBaseUrl}/img/social_<type>.png`). Providers that set `customLogo` now get it honoured regardless of type.

PR follows.

nocode-ananas:
Closes #5555.

`Setting.getProviderLogoURL` currently honours `provider.customLogo` only when `provider.type` starts with `"Custom"`. Every other OAuth provider (Google, MicrosoftOnline, GitHub, Apple, etc.) falls through to the hardcoded `${StaticBaseUrl}/img/social_<type>.png` path, dragging visitors of self-hosted Casdoor instances to `cdn.casbin.org` on every login render — same GDPR/DSGVO problem class as the Inter-font hot-link fixed in #5550.

Removing the type guard turns `customLogo` into a universal opt-out, matching its field name and the admin-UI affordance (the field is shown on every provider, regardless of type). Fallback chain (`${StaticBaseUrl}` for OAuth providers without `customLogo`, `OtherProviderInfo` table for non-OAuth categories) is preserved unchanged.

## Changes

- `web/src/Setting.js`: drop the `provider.type.startsWith("Custom")` guard on the `customLogo` branch.

Total diff: **+1 / -1** in 1 file.

## Verification

Tested on a self-hosted instance (Casdoor v3.83.0):

- Provider `Google` (`type=Google`, `category=OAuth`), `customLogo` set to a same-origin SVG URL.
- Before: login page network tab shows `${StaticBaseUrl}/img/social_google.png` (third-party hop to `cdn.casbin.org`).
- After: login page network tab shows the `customLogo` URL only; no `${StaticBaseUrl}/img/social_*.png` request.
- Providers without `customLogo` continue to use the `${StaticBaseUrl}` fallback unchanged.

### DIFF
--- a/web/src/Setting.js
+++ b/web/src/Setting.js
@@ -1243,7 +1243,7 @@ export function getClickable(text) {
 }
 
 export function getProviderLogoURL(provider) {
-  if (provider.type.startsWith("Custom") && provider.customLogo) {
+  if (provider.customLogo) {
     return provider.customLogo;
   }
   if (provider.category === "OAuth") {


hsluoyz:
@nocode-ananas

The `customLogo` input field in the UI (`OAuthProviderFields.js`) is already gated behind `provider.type.startsWith("Custom")`, so for standard providers like Google or GitHub, **the field is never shown and always remains empty**. This means the logic change here has no practical effect — `provider.customLogo` will always be falsy for standard providers, so the CDN fallback path is always taken regardless.

If the real concern is avoiding outbound requests to `cdn.casbin.org` in an air-gapped or self-hosted environment, the recommended solution is to deploy Casdoor CDN resources in intranet mode: https://casdoor.ai/docs/deployment/deploy-intranet/

nocode-ananas:
@hsluoyz Apologies — wasn't able to respond before close.

Quick clarification: the field is gated in the UI but not in the data model or API. On a self-hosted v3.83.0 instance, with the value set via `/api/update-provider`:

```
$ curl -s 'https://<host>/api/get-application?id=admin/<app>' | jq '.data.providers[] | select(.name=="provider_google") | .provider | {type, customLogo}'
{
"type": "Google",
"customLogo": "https://<host>/brand/icons/brands/google.svg"
}

`customLogo` round-trips through the backend just fine — `getProviderLogoURL` discards it at render time. So "the field always remains empty" doesn't hold when the value is set via API / CLI / DB, which is the only path available today for standard providers.
If UI symmetry matters, I'd be happy to also drop the `Custom`-gate on the `customLogo` row in `OAuthProviderFields.js` in this PR — same one-line idea, different file. Would you reconsider with that bundled in?

PR #5560, opened 2026-06-09
fix: prevent Casdoor restart caused by concurrent LDAP group sync

### LINKED ISSUE #5559: LDAP synchronization caused Casdoor to restart.
LDAP synchronization caused Casdoor to restart. 

Testing revealed that this issue begins with version v3.83.0.

Image

wangweiningse:
I tried to fix it using Opus 4.8, and it hasn't restarted since being packaged. You can have AI check it again.

### Fixes

Fixes casdoor#5559

### Problem

Since **v3.83.0** (commit `adf5f5a`, "Fixed LDAP groups synchronization routine"), Casdoor crashes and restarts whenever LDAP auto-sync runs, on instances configured with multiple LDAP servers.

As shown in #5559, all LDAP `autoSync` goroutines start almost simultaneously:

```
2026/06/09 20:06:43.330 [I] [ldap_autosync.go:58] autoSync started for 994c67b3-...
2026/06/09 20:06:43.332 [I] [ldap_autosync.go:58] autoSync started for 35ba0c59-...
... (~25 servers) ...
The old instance with pid: 1 has been stopped
2026/06/09 20:06:45 ... http server Running on http://:8000
```

### Root cause

The global `userEnforcer` (`UserGroupEnforcer`) wraps a plain `casbin.NewEnforcer`, which is **not safe for concurrent use**, and it had no locking.

`adf5f5a` changed `SyncLdapUsers` so that for **every existing user, on every sync**, it now calls `UpdateUser` → `UpdateGroupsForUser` (a `Delete` + `Add` mutation of the enforcer). With many LDAP servers syncing concurrently — plus web requests reading group membership via `LoadPolicy` — multiple goroutines read and write casbin's internal maps at the same time.

This triggers Go's **`fatal error: concurrent map read and map write`**, which is a runtime fatal error that **`recover()` cannot catch**. As a result the `SafeGoroutine` recovery is bypassed and the whole process crashes and restarts (pid 1 in the container).

This is why the crash only began in v3.83.0: the change amplified concurrent writes to the shared enforcer by orders of magnitude.

### Changes

- **`object/user_enforcer.go`** (core fix): add a `sync.Mutex` to `UserGroupEnforcer` and serialize every access to the underlying casbin enforcer. Internal re-entrant calls use new unlocked helpers (`addGroupsForUser`, `deleteGroupsForUser`, `getAllUsersByGroup`) to avoid deadlock.
- **`object/ldap_conn.go`** (hardening): `SyncLdapUsers` now checks the error/nil result of `GetLdap` before dereferencing `ldap.BaseDn`; `SyncLdapGroups` guards against cyclic/self-referential parent chains so a group can't recurse infinitely.
- **`object/group.go`** (hardening): `ConvertToTreeData` now tracks visited nodes so a self-referential group (`ParentId == Name`) or a parent cycle can't cause an infinite-recursion stack overflow when rendering the group tree.

### Testing

- `go build ./...` passes.
- Verified that concurrent LDAP auto-sync across multiple servers no longer crashes the process; enforcer access is now serialized.

### DIFF
--- a/object/group.go
+++ b/object/group.go
@@ -290,17 +290,31 @@ func (group *Group) GetId() string {
 }
 
 func ConvertToTreeData(groups []*Group, parentId string) []*Group {
+	return convertToTreeData(groups, parentId, map[string]bool{})
+}
+
+// convertToTreeData builds the group tree recursively. visited tracks the
+// parentId values already descended into, so a self-referential group
+// (ParentId == Name) or a cyclic parent chain (A -> B -> A) can't cause
+// infinite recursion and crash the whole process with a stack overflow
+// (a fatal error that recover() cannot catch). Such bad hierarchies can be
+// produced by LDAP sync when distinct DNs collapse to the same group name.
+func convertToTreeData(groups []*Group, parentId string, visited map[string]bool) []*Group {
 	treeData := []*Group{}
+	if visited[parentId] {
+		return treeData
+	}
+	visited[parentId] = true
 
 	for _, group := range groups {
-		if group.ParentId == parentId {
+		if group.ParentId == parentId && group.Name != parentId {
 			node := &Group{
 				Title: group.DisplayName,
 				Key:   group.Name,
 				Type:  group.Type,
 				Owner: group.Owner,
 			}
-			children := ConvertToTreeData(groups, group.Name)
+			children := convertToTreeData(groups, group.Name, visited)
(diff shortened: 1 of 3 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/9f90aa4a34585e059efe3099eb2cd8dea62923b7

PR #5577, opened 2026-06-14
fix: expand OAuth token scope length

monkey92t:
This PR fixes a PostgreSQL error that can occur when calling `/api/grant-consent` with longer OAuth scope values:

  ```text
  pq: value too long for type character varying(100)
  ```

  The consent flow updates the user’s application_scopes and then calls GetOAuthCode() to generate an OAuth authorization code. During that process, the requested OAuth scope string is stored in token.scope. That column was previously defined as varchar(100), which can be too short when an application uses multiple custom scopes or longer scope names.

  This change increases Token.Scope from varchar(100) to varchar(300) so longer valid OAuth scope strings can be stored without failing token creation.

### DIFF
--- a/idp/wecom_internal.go
+++ b/idp/wecom_internal.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"bytes"
 	"time"
 
 	"golang.org/x/oauth2"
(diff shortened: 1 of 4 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/9fb1b0022ab1961e4addfa898ef8c3406c82d507

Issue #5586, opened 2026-06-17
logout ignores id_token_hint when it is an id token


hamidrabedi:
I'm running Keycloak as an OIDC broker in front of Casdoor. When you log out of Keycloak it calls Casdoor's logout URL with id_token_hint set to the id token, but the token stays active on Casdoor's side.

Steps: broker Keycloak to Casdoor over OIDC, log in to Keycloak, then log out of Keycloak. The Casdoor token is not expired.

hsluoyz:
@hamidrabedi In Casdoor, `id_token` and `access_token` are the same JWT string (see `token_oauth.go`), so `ExpireTokenByAccessToken` already handles `id_token_hint` correctly. Could you share more details about your setup — specifically, are you sure the logout request is actually reaching Casdoor? The token not being expired might be a symptom of #5585 (session ID ending up in `jti` instead of `sid`), which would cause Keycloak's backchannel logout to silently fail. Try again after that fix lands.

PR #5587, opened 2026-06-17
fix: handle id_token_hint id tokens and set sid claim in oidc logout

### LINKED ISSUE #5585: Back-channel logout token has no sid claim
When Casdoor sends a back-channel logout token, the session id ends up in the jti claim instead of a sid claim. LogoutTokenClaims doesn't have a sid field at all, and generateLogoutToken does claims.RegisteredClaims.ID = sessionId, which is the jti

The back-channel logout spec uses sid to identify the session to end, so an RP that matches by sid (Keycloak's OIDC broker does this) can't find the session, and jti is no longer a unique token id either

hamidrabedi:
fixes #5585 #5586

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -442,7 +442,8 @@ func (c *ApiController) Logout() {
 			return
 		}
 
-		_, application, token, err := object.ExpireTokenByAccessToken(accessToken)
+		// id_token_hint can be an access token or an id token, so resolve both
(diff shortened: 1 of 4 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/c09015f40f70d0195c0b4fff60c944d86ae07791

PR #5597, opened 2026-06-19
fix: preserve safe adapter policy tables

buzaslan129:
## Summary

This fixes `SafeAdapter` policy persistence for adapters that use a custom policy table, such as the built-in API adapter table `casbin_api_rule`.

The current safe adapter embeds `xorm-adapter`'s `CasbinRule`, whose `TableName()` falls back to `casbin_rule` unless its private `tableName` field is set inside the adapter package. As a result, table-specific safe adapter operations can target the default policy table instead of the adapter's configured table.

The patch adds an internal policy row type that carries the configured table name through Xorm's `TableName()` hook and implements `SavePolicy()` without dropping and recreating the policy table.

## Why

With Casdoor master (`6e443ad5`) and CockroachDB v25.3 configured via the PostgreSQL wire-compatible driver:

```ini
driverName = postgres
dataSourceName = user=root host=casdoor-crdb port=26257 sslmode=disable dbname=casdoor serial_normalization=virtual_sequence
dbName = casdoor
```

Casdoor initialized the schema on a fresh database, then failed during built-in API policy initialization:

```text
panic: pq: unexpected transaction status idle
github.com/casdoor/casdoor/authz.InitApi()
    /src/authz/authz.go:153
```

The failing boundary is `Enforcer.SavePolicy()` for the built-in API enforcer.

## Changes

- Adds `SafeAdapter.SavePolicy()` that replaces policy rows transactionally without dropping the policy table.
- Uses an internal policy row type so custom adapter tables are honored by Xorm.
- Keeps `GetRules`, `RemovePolicy`, `RemovePolicies`, `UpdatePolicy`, and `UpdatePolicies` aligned with the configured adapter table.
- Adds a regression test that verifies a custom table is preserved and not dropped/recreated.

## Validation

Focused unit test:

```text
go test ./object -run TestSafeAdapterSavePolicyReplacesRulesWithoutDroppingTable -count=1
ok github.com/casdoor/casdoor/object
```

Local CockroachDB repro:

- Baseline master + CockroachDB v25.3 + `driverName=postgres` failed with `pq: unexpected transaction status idle` in `authz.InitApi()`.
- Patched branch + fresh CockroachDB v25.3 started successfully:

```text
http server Running on http://:8000
```

- Policy table check after patched startup:

```text
api_rules      98
_default_rules 0
```

(`api_rules` is `SELECT count(*) FROM casbin_api_rule`; `_default_rules` is `SELECT count(*) FROM casbin_rule`.)

## Notes

`driverName = CockroachDB` was also checked for the top-level Casdoor DB path and currently fails earlier with `panic: unsupported driver name: CockroachDB`, so this PR targets the documented PostgreSQL-compatible CockroachDB configuration path.

### DIFF
--- a/object/adapter_safe.go
+++ b/object/adapter_safe.go
@@ -1,6 +1,7 @@
 package object
 
 import (
+	"github.com/casbin/casbin/v2/model"
 	xormadapter "github.com/casdoor/xorm-adapter/v3"
 	"github.com/xorm-io/xorm"
 )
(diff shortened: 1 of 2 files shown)


buzaslan129:
@codex review

hsluoyz:
casbin adapter is the one needs to change. Not here. Closed

buzaslan129:
i see

PR #5610, opened 2026-06-25
fix: back DeviceAuthMap with Redis for multi-instance device authorization grant

### LINKED ISSUE #5609: Device Authorization Grant (RFC 8628) breaks with multiple replicas — DeviceAuthMap is in-process sync.Map
## Summary

OAuth 2.0 Device Authorization Grant (RFC 8628) breaks when Casdoor runs with more than one replica behind a load balancer. The device authorization state is kept in an in-process `sync.Map` and is not shared across instances, so the three steps of the device flow frequently land on different pods and fail with `authorization_pending` (or the approval step silently fails).

## Root cause

Device auth state is stored in a process-local map:

`object/token_oauth_util.go`
```go
var DeviceAuthMap = sync.Map{}
```

All three steps of the device flow read/write this local map:

| Step | Code | Where state lives |
| --- | --- | --- |
| 1. Client requests device code | `controllers/auth.go:1592` `DeviceAuthMap.Store(deviceCode, ...)` / `Store(userCode, ...)` | memory of the pod that served this request |
| 2. User approves in browser | `controllers/auth.go:213` `DeviceAuthMap.LoadAndDelete(form.UserCode)` | load-balanced to a different pod → `ok = false` → fails |
| 3. Client polls `/api/oauth/token` with `device_code` | `controllers/token.go:273` `DeviceAuthMap.Load(deviceCode)` | load-balanced to a different pod → not found → stuck on `authorization_pending` |

`InitCleanupDeviceAuthMap()` only expires entries from the *local* map, so there is no cross-instance visibility at all.

This is different from the login session, which already goes through Redis (set via `SessionProvider = "redis"` + `redisEndpoint` in `main.go`). Device auth state is the only piece that is still purely in-memory.

## Steps to reproduce

1. Deploy Casdoor with `replicas: 2` behind a load balancer (Envoy / nginx / etc.).
2. Run an RFC 8628 device flow against any OAuth application (e.g. `gh`-style device login, or any CLI using device code grant).
3. Repeat a few times. A meaningful fraction of flows hang on `authorization_pending` until timeout, or the browser approval page reports the code as invalid, because steps 1/2/3 were served by different pods.

Setting `replicas: 1` makes the problem disappear (all steps hit the same pod).

## Environment

- Casdoor version: `v3.99.0` (still present on `master`)
- Deployment: 2 replicas behind Envoy Gateway
- `redisEndpoint` is configured and reachable; sessions already use Redis

## Suggestion

Casdoor already has a Redis client (configured via `redisEndpoint`, already used for sessions). It would be great to back `DeviceAuthMap` by Redis (with TTL) so the device authorization grant works correctly in multi-instance / HA deployments. Sticky routing at the LB is not really viable here because the three requests originate from different actors (the polling client, the user's browser, etc.) and cannot share a routing key.

Happy to help with a PR if there's interest.

Sweetnow:
## Description

The OAuth 2.0 Device Authorization Grant (RFC 8628) currently keeps its state in an in-process `sync.Map`:

```go
// object/token_oauth_util.go
var DeviceAuthMap = sync.Map{}
```

All three steps of the device flow read/write this local map (requesting the device code in `controllers/auth.go`, the user approving it in a browser via `controllers/auth.go`, and the client polling `/api/oauth/token` in `controllers/token.go`). When Casdoor runs with more than one replica behind a load balancer, these requests are spread across instances, so the state is not shared and the flow fails (`authorization_pending` until timeout, or "DeviceCode Invalid" on the approval step). This is different from the login session, which already uses Redis via `SessionProvider = "redis"` + `redisEndpoint`.

This was reported in #5609, including a multi-replica reproduction.

## What this changes

- Introduce `object/device_auth_map.go` with a small interface mirroring the subset of `sync.Map` methods used by the device flow (`Load` / `Store` / `Delete` / `LoadAndDelete` / `Range`), backed by two implementations:
  - `redisDeviceAuthMap` — stores `DeviceAuthCache` entries as JSON under a `casdoor:device_auth:` prefix with a TTL derived from `ExpiresIn` (falling back to `DeviceAuthExpiresIn`), so entries expire automatically and no background cleanup is needed.
  - `memoryDeviceAuthMap` — wraps a plain `sync.Map`, preserving the exact current behavior for single-instance deployments.
- `var DeviceAuthMap` defaults to the in-memory implementation; `object.InitDeviceAuthStore()` (called from `main.go`) switches it to Redis when `redisEndpoint` is configured. The option reuses the same `redisEndpoint` already used by the session store and follows the same `host:port,db[,password]` format.
- The existing `InitCleanupDeviceAuthMap` sweep is left untouched: it still cleans up the in-memory backend, and is a no-op for the Redis backend (whose keys expire via TTL).
- The `// indirect` on `github.com/redis/go-redis/v9` is promoted to a direct dependency (no version change).

The interface keeps the same `any`-typed method signatures that the call sites already use, so `controllers/auth.go` and `controllers/token.go` require no changes.

## Tests

- `go build ./...` ✅
- `go vet ./object/ ./controllers/` ✅
- No upstream unit tests exist for the device flow today. Verified end-to-end behaviorally: a deployment with `replicas: 2` + `redisEndpoint` set previously failed the device flow whenever requests landed on different pods; after this change all three steps resolve consistently because the state lives in Redis.

I did not add unit tests since the device flow has no existing test scaffolding in the repo — happy to add tests (e.g. against a `miniredis` for the Redis backend) if the maintainers prefer a specific pattern.

Fixes #5609.

## Backwards compatibility

- No `redisEndpoint` ⇒ identical to today (in-process map).
- With `redisEndpoint` ⇒ device flow now works across replicas. Existing single-instance users with Redis configured also keep working; the only new footprint is `casdoor:device_auth:*` keys with short TTLs.

## Notes

- `LoadAndDelete` uses Redis `GETDEL` (atomic, Redis ≥ 6.2). If support for older Redis is required I can switch to a non-atomic `GET` + `DEL`.
- `Range` is a no-op for the Redis backend by design; the periodic cleanup is unnecessary because of the TTL.


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/98b6011b632ba7fcbe1b2a63344071af474be14c

PR #5641, opened 2026-07-11
feat: fix cramped toolbar buttons and unnecessary scrollbars in admin UI

oxsean:
## Problems

Four long-standing layout annoyances in the admin console:

1. **Forced horizontal scrollbars on list pages**: every list page passes `scroll={{x: "max-content"}}`, which makes the table *always* as wide as the sum of the column `width`s — the design widths act as hard floors even when the viewport has room. E.g. the Certs table forced a scrollbar for 75 extra pixels while cells held values like `4096` in a 130px column.
2. **A permanent vertical scrollbar on every page**: `ManagementPage` wraps its content in a `minHeight: 100vh` div, but that div lives inside `#parent-area` (also `min-height: 100vh`) **together with the 68px footer** — so every page overflows by exactly the footer height, even a page with one table row, and the footer is permanently pushed below the fold.
3. **Cramped list page toolbars**: the buttons in every list page's table title bar (`Add`, `Upload (.xlsx)`, ...) are `size="small"` (24px high, 7px inline padding) inside a title bar whose padding is crushed to `5px 8px`, right next to the full-size row action buttons (`Edit`, `Delete`, 32px), which makes the mismatch obvious.
4. **Cramped edit pages**: the card title bar is 38px tall but holds 32px `Save` / `Save & Exit` buttons (3px of air), and the bottom button row has no bottom margin, so it sits **1px** away from the card's bottom border.

## Fix

All column `width` definitions are kept untouched — only the scroll mode changes:

- `web/src/*ListPage.js` (36 files, one line each): `scroll={{x: "max-content"}}` → `scroll={{x: true}}`. The table now renders `width: auto; min-width: 100%`, so the column widths act as *preferred* widths: honored when the viewport has room, shrunk proportionally (never below content) when it doesn't. Tables only scroll when the actual content genuinely exceeds the viewport (e.g. Users with its 20+ columns), and `fixed` (sticky) columns keep working.
- `web/src/index.css`: table cells get `white-space: nowrap` so shrinking only ever absorbs column padding, never wraps content (under `x: "max-content"` nothing could wrap anyway, so this is the status quo made explicit). Also: table title bars get `8px 12px` padding, their small toolbar buttons are sized like default buttons (32px high, 15px inline padding) to match the row action buttons, and small card title bars get `min-height: 46px`.
- `web/src/ManagementPage.js` (one line): content wrapper `minHeight: "100vh"` → `flex: 1`, so content + footer fill exactly one viewport. Short pages no longer scroll and the footer is visible; long pages scroll exactly as before.
- `web/src/*EditPage.js` (25 files, one line each): bottom button row `{marginTop: "20px", marginLeft: "40px"}` → `{margin: "20px 40px"}` for symmetric breathing room above the card's bottom edge.

- `web/cypress/support/commands.js`: the shared `login` command now disables the page tour (`isTourVisible=false`), so the tour popover can never cover elements the tests click — the tokens spec only passed by geometric accident before.

Verified locally on organizations / users / applications / providers / certs / roles / permissions / tokens / webhooks / syncers / sessions / invitations plus the token edit page, a long edit page (`/organizations/built-in`), and a 900px-wide viewport for the sticky-column / genuine-overflow cases.

(`index.css` also picks up the repo's own `stylelint --fix` corrections, a wrapped over-long line, and a file-level `no-descending-specificity` disable for pre-existing selector ordering — the pre-commit hook refuses any commit touching the file otherwise; same approach as the existing disables in `App.less`.)

### Certs list — scrollbar gone, columns keep their design proportions, footer on screen, toolbar matches row buttons

| Before | After |
|---|---|
| before | after |

### Organizations list — wide table still scrolls (legitimately), no vertical scrollbar

| Before | After |
|---|---|
| before | after |

### Token edit page — title bar and bottom buttons breathe

| Before (top) | After (top) |
|---|---|
| before | after |

| Before (bottom) | After (bottom) |
|---|---|
| before | after |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ server
 
 # include helm-chart
 !manifests/casdoor
+
+web/cypress/videos/
+web/cypress/screenshots/
--- a/web/cypress/support/commands.js
+++ b/web/cypress/support/commands.js
@@ -29,7 +29,12 @@ const selector = {
   loginButton: ".ant-btn",
 };
 Cypress.Commands.add('login', ()=>{
-  cy.visit("http://localhost:7001");
+  cy.visit("http://localhost:7001", {
+    onBeforeLoad(win) {
+      // Disable the page tour so its popover never covers elements the tests click
+      win.localStorage.setItem("isTourVisible", "false");
+    },
+  });
   cy.get(selector.username).type("admin");
   cy.get(selector.password).type("123");
   cy.get(selector.loginButton).click();
--- a/web/src/AdapterEditPage.js
+++ b/web/src/AdapterEditPage.js
@@ -319,7 +319,7 @@ class AdapterEditPage extends React.Component {
         {
           this.state.adapter !== null ? this.renderAdapter() : <Loading type="page" tip={i18next.t("login:Loading")} />
         }
-        <div style={{marginTop: "20px", marginLeft: "40px"}}>
+        <div style={{margin: "20px 40px"}}>
           <Button size="large" onClick={() => this.submitAdapterEdit(false)}>{i18next.t("general:Save")}</Button>
           <Button style={{marginLeft: "20px"}} type="primary" size="large" onClick={() => this.submitAdapterEdit(true)}>{i18next.t("general:Save & Exit")}</Button>
           {this.state.mode === "add" ? <Button style={{marginLeft: "20px"}} size="large" onClick={() => this.deleteAdapter()}>{i18next.t("general:Cancel")}</Button> : null}
--- a/web/src/AdapterListPage.js
+++ b/web/src/AdapterListPage.js
@@ -227,7 +227,7 @@ class AdapterListPage extends BaseListPage {
 
     return (
       <div>
-        <Table scroll={{x: "max-content"}} columns={columns} dataSource={adapters} rowKey={(record) => `${record.owner}/${record.name}`} size="middle" bordered pagination={paginationProps}
+        <Table scroll={{x: true}} columns={columns} dataSource={adapters} rowKey={(record) => `${record.owner}/${record.name}`} size="middle" bordered pagination={paginationProps}
           title={() => (
             <div>
               {i18next.t("general:Adapters")}&nbsp;&nbsp;&nbsp;&nbsp;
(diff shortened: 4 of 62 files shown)


hsluoyz:
@oxsean Nice cleanup — these are all real, long-standing annoyances (the forced horizontal scrollbar from `x: "max-content"` and the footer-height vertical overflow are genuine bugs, not cosmetic nitpicks), and the per-file changes are consistent and well explained. A couple of things before merge:

1. **`web/cypress/videos/tokens.cy.js.mp4` looks accidentally committed.** Cypress video artifacts are usually gitignored — could you drop this binary from the PR?

2. **The global `white-space: nowrap` on table cells has a wide blast radius.** I agree it's a no-op under the old `x: "max-content"` in most cases, but any column that previously relied on wrapping (e.g. long `description` text) will now be forced single-line + scroll. Could you double-check the content-heavy tables (long text columns, and the fixed/sticky-column cases) on a narrow viewport so we don't regress readability there?

Otherwise the approach (`x: true` + keeping the design widths as preferred widths, `flex: 1` on the content wrapper) is the right call. Also worth noting the PR bundles in some unrelated `stylelint --fix` churn and the Cypress tour fix — understandable given the pre-commit hook, just flagging it for reviewers.

oxsean:
@hsluoyz Thanks for the review! Both points addressed:

**1. Cypress video** — good catch, that was a local test-run artifact swept in by accident. Removed in the latest push, and I added `web/cypress/videos/` / `web/cypress/screenshots/` to `.gitignore` so it can't happen again.

**2. `white-space: nowrap` blast radius** — I went through this carefully:

- **Code sweep**: no list page renders a wrap-reliant long-text column. Every unbounded-text column in the repo (`requestUri` / `response` / `object` on Records, `accessToken` on Tokens) already has `ellipsis` + a fixed width, so it truncates instead of wrapping. Everything else is short values (names, dates, enums, switches).
- **Narrow viewport test**: Records (the most content-heavy table, with real long-JSON cells) at a 900px viewport — single-line rows, ellipsis truncation intact, all three sticky right columns working, no cell overflow:

records at 900px

The key invariant: under the old `x: "max-content"` the table was always sized to fit every cell on one line, so no column could ever actually wrap — `nowrap` just makes that existing behavior explicit so that shrinking under `x: true` only absorbs column padding, never squeezes real content into wrapped lines.

Also rebased onto latest master (no conflicts with #5640).

PR #5642, opened 2026-07-12
feat: add org option to output LDAP memberOf as valid DN

oxsean:
### Problem

The built-in LDAP server returns `memberOf` as the raw Casdoor group id `owner/name` (e.g. `devops/dev`), which is **not a valid LDAP DN**. Standard LDAP clients that parse `memberOf` as a DN — e.g. Nexus Repository's dynamic group mapping (`userMemberOfAttribute`) — extract the RDN and crash with `IndexOutOfBounds` on `devops/dev`, taking down the Nexus user-list UI.

### Fix

Add an opt-in per-organization option **`ldapMemberOfDn` (default off)**. When enabled, `memberOf` is output as a valid DN `cn=<group>,ou=<owner>,<dc-suffix>` — matching the `posixGroup` entry DN already returned by group searches (`ldap/server.go`) and the format Casdoor's own LDAP client parses (`dnToGroupName` in `object/ldap_conn.go`). Clients can then extract the group name (`dev`) for role mapping.

The reverse `(memberOf=...)` filter also accepts the DN form, for round-trip consistency.

**Backward compatible:** default-off, existing setups return `owner/name` unchanged.

### Verification

- `go build ./...`
- Unit test `TestGroupToDN` (normal, cross-org, no-prefix, no-dc)
- Live `ldapsearch` against the built-in server:
  - off → `memberOf: devops/dev`
  - on  → `memberOf: cn=dev,ou=devops,dc=example,dc=com`
- UI round-trip: org-edit switch (default off) → toggle on → Save → persisted

### UI

A new "LDAP memberOf as DN" switch on the organization edit page, between "LDAP attributes" and "LDAPs". i18n added for en and zh.

### DIFF
--- a/ldap/server.go
+++ b/ldap/server.go
@@ -288,7 +288,11 @@ func buildUserSearchEntry(user *object.User, baseDN string, attrs []string, org 
 	e.AddAttribute("objectClass", "posixAccount")
 	if IsLdapAttrAllowed(org, ldapMemberOfAttr) {
 		for _, group := range user.Groups {
-			e.AddAttribute(ldapMemberOfAttr, message.AttributeValue(group))
+			value := group
+			if org != nil && org.LdapMemberOfDn {
+				value = groupToDN(group, baseDN)
+			}
+			e.AddAttribute(ldapMemberOfAttr, message.AttributeValue(value))
 		}
 	}
 	for _, attr := range attrs {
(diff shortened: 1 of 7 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/4b1915e02cca2f26e3166287182a12bfb7736c9f

PR #5663, opened 2026-07-24
fix: remove session row on logout after SessionRegenerateID

### LINKED ISSUE #5662: Logout leaves stale session row in Sessions admin UI
## Bug
After calling `/api/logout`, the user's session row remains visible in the Sessions admin UI even though the user is logged out.

## Root cause
1. `Logout()` calls `ClearUserSession()`
2. `ClearUserSession()` calls `SessionRegenerateID()`
3. `deleteUserSession()` then reads the *new* Beego session id and passes it to `DeleteSessionId`
4. The *old* session id stays in `Session.SessionId`
5. Because the array is non-empty, the Session DB row is not deleted

## Expected
Logging out should remove the current Beego session id from `Session.SessionId`. If the array becomes empty, the Session row should be deleted.

## Actual
The Session row remains with the pre-regenerate Beego session id.

VikaLvovich:
## Summary
- Capture the current Beego session id before `ClearUserSession()` / `SessionRegenerateID()`
- Pass that id into `deleteUserSession` / `DeleteSessionId` so the DB entry is cleared
- Apply the same capture ordering in `SsoLogout` without changing its broader flow
- Add a small `util.DeleteVal` regression test documenting empty-list cleanup

Fixes #5662

## Test plan
- [ ] `go test ./util/ -run TestDeleteVal`
- [ ] Manual: login → Sessions UI shows row → `/api/logout` → row disappears


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/8bb6c391480b2ac9ea04d3e2b5d3c7ab12115b1f

PR #5671, opened 2026-07-27
fix: logout session cleanup for non-built-in applications

### LINKED ISSUE #5669: Logout does not remove the session row for non-built-in applications
### Description

When a user signs in through an application other than `app-built-in`, Casdoor stores the Beego session ID in a `Session` row associated with that application's name.

During logout, Casdoor attempts to remove the session ID only from the `app-built-in` session row. As a result, the actual application session remains visible in the Sessions admin UI even though the user has logged out.

### Steps to reproduce

1. Create or use an application whose name is not `app-built-in`.
2. Sign in to that application.
3. Confirm that a session row exists for the application.
4. Log out.
5. Open the Sessions page in the admin UI.

### Current behavior

The session row for the application remains after logout.

### Expected behavior

Logout should remove the current Beego session ID from the `Session` row associated with the application used during login. If no session IDs remain, the row should be deleted.

VikaLvovich:
Fixes #5669

### Summary
Session records are stored under the application used during login, but logout previously attempted to remove the Beego session ID only from the `app-built-in` session row.

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -559,8 +559,9 @@ func (c *ApiController) SsoLogout() {
 		return
 	}
 
-	_, err = object.DeleteSessionId(util.GetSessionId(owner, username, object.CasdoorApplication), currentSessionId)
-	if err != nil {
+	// Remove the Beego session id from the Session row for whatever application
+	// was used at login (not only app-built-in).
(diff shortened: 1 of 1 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/3f676725cecdf245254834a8f7e34a8eb79e381d

PR #3628, opened 2025-03-01
feat: update user language when the language changed on login page

WindSpiritSR:
[Demo](https://casdoor.dev.leviatan.cn)
Username: `admin`
Password: `123`

fixed #3548

PR #5672, opened 2026-07-27
feat: save the language chosen on the signup page to the new user

markwellis:
Signup copies every field from the signup form onto the new user, except for the selected language. 

This patch stores the language of the signup request on the new user, the same as the login page already does.

The same issue existed on the login page, and was fixed in #3628

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -246,6 +246,7 @@ func (c *ApiController) Signup() {
 		Email:             strings.ToLower(authForm.Email),
 		Phone:             authForm.Phone,
 		CountryCode:       authForm.CountryCode,
+		Language:          authForm.Language,
 		Address:           []string{},
 		Affiliation:       authForm.Affiliation,
 		IdCard:            authForm.IdCard,
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -1050,6 +1050,7 @@ func (c *ApiController) Login() {
 						Phone:             userInfo.Phone,
 						CountryCode:       userInfo.CountryCode,
 						Region:            userInfo.CountryCode,
+						Language:          authForm.Language,
 						Score:             initScore,
 						IsAdmin:           false,
 						IsForbidden:       false,
--- a/web/src/auth/AuthCallback.js
+++ b/web/src/auth/AuthCallback.js
@@ -310,6 +310,8 @@ class AuthCallback extends React.Component {
       method: method,
       userCode: innerParams.get("userCode") || "",
       codeVerifier: codeVerifier, // Include PKCE code verifier
+      // The selector on the login page is gone by now, so use the language it stored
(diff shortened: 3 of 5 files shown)


hsluoyz:
Thanks for the patch, and for spotting the inconsistency — the signup path really does
copy almost every form field onto the new user while dropping the language, so the gap
is real. A couple of things I'd like to sort out before merging, though.

### 1. This fixes it differently from #3628

#3628 persists the language from the **form field** (`authForm.Language`,
`controllers/auth.go` L1287-1297). On the frontend, `values["language"]` is set from
`this.state.userLang` (`web/src/auth/LoginPage.js` L510), and `userLang` is only
populated when the user *actively* clicks the language selector (L724). So the login
path writes the language **only when the user explicitly picked one**.

This PR instead reads the `Accept-Language` header, and `GetAcceptLanguage()` never
returns an empty string — `conf.GetLanguage()` falls back to `"en"`
(`conf/conf.go` L99-109). So every new signup gets a language stamped on it, whether
the user chose one or not. Two different mechanisms and two different semantics for the
same feature.

Could we make signup send `values["language"]` from the selector the same way
`LoginPage` does, and have `Signup()` use `authForm.Language` only when it's non-empty?
That keeps both paths identical and avoids the points below.

### 2. Behaviour change for non-UI callers of `/api/signup`

Integrations that POST to `/api/signup` directly don't send Casdoor's synthetic
`Accept-Language`. They either send nothing (→ stored as `"en"`) or a real browser
header like `de-DE,de;q=0.9` (→ stored as `"de"`). Previously the field stayed empty,
meaning "follow the browser / default"; now it becomes a hard override that switches
the UI on every later login (`web/src/App.js` L382-387).

Related: `GetAcceptLanguage()` just truncates to the first two characters
(`controllers/util.go` L67), so `zh-TW` is stored as `zh` (Simplified). That lossiness
was previously scoped to translating a single response; this PR persists it to the user
record.

### 3. Possible bypass of the organization's language list

In `renderLanguageSelect` (`web/src/auth/SignupPage.js` L263-266) we return early when
the `Languages` signup item is hidden — *before* the
`organizationObj.languages.length <= 1` normalization runs. In that case
`i18next.language` is whatever the browser detected, so a signup can store a language
the organization doesn't allow (e.g. `de` for a zh-only org). The login path can't do
this because the value comes from the selector, which is already constrained to the
org's list.

### 4. The other signup path still misses it

Auto-signup via a social/OIDC provider builds its `User` at `controllers/auth.go`
L1040-1063 and also leaves `Language` unset. Not a defect of this PR, but it would be
good to cover it here so the behaviour is uniform.

---

Happy to merge a version that mirrors #3628 (form field, only written when non-empty),
ideally with the provider auto-signup path covered too.

markwellis:
@hsluoyz I have reworked it based on your feedback. Thanks :pray:

PR #5677, opened 2026-07-29
feat: capture DingTalk title during OAuth login

0to1toeverything:
## What changed

- Parse the `title` field returned by DingTalk's corporate user details API.
- Add the title to `UserInfo.Extra` so the existing provider user mapping can populate the Casdoor user profile.
- Add an HTTP-level regression test covering the DingTalk OAuth user info flow.

## Why

The DingTalk OAuth provider already calls `topapi/v2/user/get`, but it only reads the user's mobile number, email, and union ID. The API also returns the user's job title, which was discarded before the configured `title -> title` user mapping could process it.

## Impact

On DingTalk OAuth sign-in, applications with the existing title user mapping can populate an empty Casdoor user `Title` field. This change does not alter profile overwrite behavior and does not affect other OAuth providers.

## Validation

- `go test ./idp`
- `go test -race ./idp`
- `go vet ./idp`


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/f378240177dbaa565300ba92ce4d41c1a2a6b472

PR #5686, opened 2026-07-31
fix(auth): return generic sign-in error for unknown users

### LINKED ISSUE #5685: fix: Prevent user enumeration during login
### Problem

Currently, when a user attempts to log in with a username that does not exist, Casdoor returns a message indicating that the user was not found.

This behavior allows an attacker to determine whether a particular username exists in the system. It can be used for user enumeration and may expose information about registered accounts.

AndrewBlinets:
Use a single client-facing message for missing users and failed LDAP
lookups to prevent username enumeration (OWASP). Keep user-not-found
in audit via SigninReasonUserNotFound.

fix: #5685

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -591,7 +591,7 @@ func (c *ApiController) Login() {
 				c.ResponseError(err.Error(), nil)
 				return
 			} else if user == nil {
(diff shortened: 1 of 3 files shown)


nomeguy:
closed due to: https://github.com/casdoor/casdoor/issues/5685#issuecomment-5150557878

PR #5694, opened 2026-08-04
feat: Add organization minimum password age policy with centralized e…

### LINKED ISSUE #5693: [feature] Add Minimum Password Age to password policy
Casdoor already provides password policy configuration, but there is currently no way to define the minimum amount of time a password must remain valid before the user can change it again.

This can be important for environments where password changes are subject to security policies or password history requirements.

For example, without a minimum password age, a user can potentially change their password multiple times in a short period and eventually return to a previously used password, depending on the configured password history policy.

A minimum password age provides an additional control to prevent rapid password rotation and makes password history policies more effective.

### Feature description

Add a new password policy setting:

Minimum Password Age

The setting defines the minimum period that must elapse after a successful password change before the user is allowed to change the password again.

For example:

Minimum Password Age = 24 hours

After changing the password, the user cannot change it again until 24 hours have elapsed.

A value of 0 should disable the restriction.

### Expected behavior

After a successful password change:

Casdoor records the timestamp of the password change.
On the next password change attempt, Casdoor checks the elapsed time.
If the minimum password age has not expired, the operation is rejected.
Once the configured period has elapsed, the password can be changed normally.
The validation should be performed server-side and must not rely on frontend validation.

AndrewBlinets:
The purpose is to prevent users from changing their password repeatedly within a short period of time. This is useful for environments where password history and password-change policies are required.

The restriction is disabled by default (`0`), so existing installations and behavior remain unchanged unless explicitly configured.

The validation is performed server-side and covers the password-changing flows to prevent bypassing the policy through another API.

fix: #5693


hsluoyz:
not planned

PR #5697, opened 2026-08-06
fix: sync casbin user rules after user rename

fangker:
• ## What
  - Update `casbin_user_rule` when a user is renamed, so `g` policies keep pointing to the new user ID.
  - Refresh `userEnforcer` policy cache only when the rename actually changes at least one rule.
  - Add a regression test for the rename flow.

  ## Why
  - Prevent stale group mappings after a user rename.
  - Keep the database and in-memory Casbin policy state aligned.
  - Prevent group-related APIs from returning users that are no longer valid after a rename.

  ## Test
  - `go test ./object -run TestUserChangeTriggerRenamesCasbinRule -count=1 -v -tags skipCi`

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -1449,7 +1449,25 @@ func userChangeTrigger(owner string, oldName string, newName string) error {
 		return err
 	}
 
-	return session.Commit()
+	oldId := util.GetId(owner, oldName)
+	newId := util.GetId(owner, newName)
+	tableNamePrefix := conf.GetConfigString("tableNamePrefix")
+	affected, err := session.Table(tableNamePrefix+"casbin_user_rule").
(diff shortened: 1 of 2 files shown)


hsluoyz:
Thanks for catching this — stale `g` rules pointing at the old user ID after a rename is a real bug, and the group member count / group-based permission expansion both read from those rules.

We ended up implementing it a bit differently, going through the existing `UserGroupEnforcer` API rather than updating `casbin_user_rule` directly. The main reason is that this table is created by `adapter.go` with an empty table prefix, so `tableNamePrefix + "casbin_user_rule"` would point at a non-existent table and break renames entirely on deployments that set `tableNamePrefix`.

Fixed in https://github.com/casdoor/casdoor/commit/cdd805c3. Closing this one — thanks again for the report!

Issue #5707, opened 2026-08-10
[feature] Add a global search filter to User API


AndrewBlinets:
## Problem

The current `User API / ApiController.GetUsers` supports filtering users by individual fields, but it does not provide a convenient way to search for a user using a single value across the main user identity fields.

A common use case is a user search field where the input may contain either:

* username/login;
* full name;
* email.

Without a global search filter, clients need to know which particular field should be searched or implement multiple requests and combine the results themselves.

## Proposed solution

Extend the existing `User API / ApiController.GetUsers` with an optional query parameter for global user search.

The proposed parameter name is:

```text
search
```

This is a proposed name only. The final parameter name should follow the naming conventions already used in the Casdoor API. If necessary, it should be renamed to match the project's existing API standards.

The parameter should search across:

* `name` — username/login;
* `displayName` — full name;
* `email` — email address.

The conditions between these fields should use `OR` semantics:

```text
name contains <search>
OR
displayName contains <search>
OR
email contains <search>
```

hsluoyz:
not planned

PR #5708, opened 2026-08-10
feat: add global search param to get-users API

AndrewBlinets:
Allow filtering users by name, displayName, or email via optional search query.

#5707


hsluoyz:
not planned

Issue #5714, opened 2026-08-12
feat: make group metadata and member pagination bounded by page size


yqnops:
## Description

Casdoor's user pagination is bounded at the SQL result level when `groupName` is empty, but the group-related paths are not bounded by `pageSize` at the policy or membership level.

On v3.152.0 (`7325999791deaeab538f86082d70858d0ab4d773`):

1. `GET /api/get-group` always calls `ExtendGroupWithUsers()`.
2. `ExtendGroupWithUsers()` calls `GetAllUsersByGroup()`.
3. `GetAllUsersByGroup()` calls `LoadPolicy()` before retrieving the role members.
4. `GET /api/get-users?groupName=...&p=...&pageSize=...` calls the group membership path once for the count and again for the page query. Both calls load the full policy and materialize all member IDs before SQL `LIMIT/OFFSET` is applied.
5. The SQL query then uses the complete member-name list in an `IN (...)` clause. Therefore reducing `pageSize` limits returned User rows, but does not bound policy loading, member-ID memory, or SQL parameter count.
6. `GET /api/get-groups?p=...&pageSize=...` calls `ExtendGroupsWithUsers()` and repeats `GetAllUsersByGroup()` for every returned group, so a page can reload the policy multiple times.
7. `pageSize` is parsed directly and no server-side maximum is enforced by these handlers.

This becomes visible well before millions of users. For example, with about 21,000 membership policy rows and a group of about 2,400 members, a request for a small page still performs work proportional to the complete policy and/or complete group membership. The request therefore has no meaningful upper bound based on `pageSize`.

Relevant source:

- `controllers/group.go`: `GetGroup()` and `GetGroups()`
- `controllers/user.go`: `GetUsers()`
- `object/group.go`: `GetGroupUserCount()`, `GetPaginationGroupUsers()`, and `ExtendGroup(s)WithUsers()`
- `object/user_enforcer.go`: `GetAllUsersByGroup()`

Related history:

- #3145 identified that group APIs should not fetch all users.
- #3146 limited results to the first 10 users, but truncation does not provide correct pagination.
- #3518 replaced a full user-table scan with Casbin membership lookup, but retained eager membership expansion and full `LoadPolicy()`.

## Expected bounded contract

Could the group APIs provide a documented, server-enforced bounded read contract?

- A group metadata read should not implicitly expand all members. This could be an explicit opt-out/opt-in parameter for compatibility, or a separate metadata endpoint.
[...]

PR #5716, opened 2026-08-12
feat: allow group reads without member expansion

yqnops:
## What

Add an optional `withUsers` query parameter to:

- `GET /api/get-group`
- `GET /api/get-groups`

The default remains `true` for backward compatibility. Only an explicit `withUsers=false` skips `ExtendGroupWithUsers()` / `ExtendGroupsWithUsers()`.

## Why

Group metadata consumers do not always need membership data. Today these endpoints always expand members, which calls `GetAllUsersByGroup()` and reloads the complete Casbin policy. For `get-groups`, that work is repeated for each group in the returned page.

Allowing callers to opt out gives metadata reads a bounded path with respect to membership policy size and avoids unnecessary response data. This is the first, compatibility-preserving part of #5714.

This PR intentionally does **not** claim to fix group-member count or pagination. `GET /api/get-users?groupName=...` still materializes the complete membership set before SQL pagination; that broader storage-level count/cursor work remains tracked by #5714.

## Compatibility

- omitted `withUsers`: existing behavior
- `withUsers=true`: existing behavior
- `withUsers=false`: return group metadata without populating `users`
- unknown values: existing behavior

## Tests

- Added focused unit coverage for default, explicit true, explicit false, and unknown values.
- `go test ./util -run '^TestShouldExtendGroupWithUsers$' -count=1 -v`
- `go test ./controllers -tags skipCi -run '^$' -count=1`
- Parsed `swagger/swagger.json` and `swagger/swagger.yml`
- `git diff --check`

Refs #5714

### DIFF
--- a/controllers/group.go
+++ b/controllers/group.go
@@ -27,6 +27,7 @@ import (
 // @Tag Group API
 // @Description get groups
 // @Param   owner     query    string  true        "The owner of groups"
+// @Param   withUsers query    bool    false       "Whether to include group members (default true)"
 // @Success 200 {array} object.Group The Response object
 // @router /get-groups [get]
 func (c *ApiController) GetGroups() {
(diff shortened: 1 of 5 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/dfd07ed91e1dceea2979ac2716e25c3adaf541f9

PR #5720, opened 2026-08-14

maximaa2001:
## Description
When `columns` is provided on `/api/update-user`, only update those columns (plus service fields like `hash` / `updated_time`). Append admin-sensitive fields only when `columns` is empty (full update path).

## Why
Admin callers currently always get `email`, `phone`, `is_forbidden`, etc. appended even for partial updates like `?columns=tag`, which breaks the `columns` whitelist contract.

## Compatibility
- Casdoor web UI: unchanged (no `columns` query param)
- Empty `columns` + admin: unchanged
- Explicit `columns` + admin: now respects the whitelist only

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -856,7 +856,8 @@ func UpdateUser(id string, user *User, columns []string, isAdmin bool) (bool, er
 		}
 	}
 
-	if len(columns) == 0 {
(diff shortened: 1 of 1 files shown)


maximaa2001:
Superseded by the reworked PR on branch maximaa2001-fix-update-user-admin-columns.

PR #5721, opened 2026-08-14

maximaa2001:
## Summary
Respect the `columns` whitelist on `/api/update-user` for admin callers.

1. **Keep admin-sensitive fields for full updates only**  
   When `columns` is empty, admins still get `email`, `phone`, `is_forbidden`, and the rest appended as before.
2. **Honor explicit `columns` as a partial-update whitelist**  
   When callers pass `columns` (e.g. `?columns=tag`), only those columns are updated — admin status no longer expands the list.
3. **Leave the UI path unchanged**  
   The web console does not send `columns`, so admin edit behavior stays the same.

## Test plan
- [ ] Admin update without `columns` still persists email / phone / isForbidden
- [ ] Admin update with `?columns=tag` updates only `tag` (plus hash / updated_time)
- [ ] `go test ./object/` passes

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -856,7 +856,8 @@ func UpdateUser(id string, user *User, columns []string, isAdmin bool) (bool, er
 		}
 	}
 
-	if len(columns) == 0 {
+	columnsWereEmpty := len(columns) == 0
+	if columnsWereEmpty {
 		columns = []string{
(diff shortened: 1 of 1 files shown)


maximaa2001:
Renamed branch to match Casdoor naming (fix-…).

PR #5722, opened 2026-08-14

### LINKED ISSUE #5723: Admin /api/update-user ignores columns whitelist
## Bug

When an admin (or application) caller updates a user via `/api/update-user` with an explicit `columns` query parameter, Casdoor still appends admin-sensitive fields (`email`, `phone`, `is_forbidden`, etc.) to the SQL update set.  
As a result, `columns` does not act as a partial-update whitelist for admin callers.

## Root cause

In `object.UpdateUser()`, after the default whitelist is applied for an empty `columns` list, this block always runs for admins:

```go
if isAdmin {
    columns = append(columns, "name", "id", "email", "phone", ...)
}
```

So even when the caller passes e.g. `?columns=tag`, the update still includes the full admin-sensitive column set.

## Expected

- If `columns` is empty: keep the current full-update behavior (default whitelist + admin-sensitive fields for admins).
- If `columns` is set: update only the requested columns (plus service fields like `hash` / `updated_time`), regardless of admin status.

## Actual

For admin callers, any non-empty `columns` value is expanded with admin-sensitive fields, so partial updates like `?columns=tag` still write `email`, `phone`, `is_forbidden`, and related columns.

maximaa2001:
## Summary
Respect the `columns` whitelist on `/api/update-user` for admin callers.

1. **Keep admin-sensitive fields for full updates only**  
   When `columns` is empty, admins still get `email`, `phone`, `is_forbidden`, and the rest appended as before.
2. **Honor explicit `columns` as a partial-update whitelist**  
   When callers pass `columns` (e.g. `?columns=tag`), only those columns are updated — admin status no longer expands the list.
3. **Leave the UI path unchanged**  
   The web console does not send `columns`, so admin edit behavior stays the same.

## Test plan
- [ ] Admin update without `columns` still persists email / phone / isForbidden
- [ ] Admin update with `?columns=tag` updates only `tag` (plus hash / updated_time)
- [ ] `go test ./object/` passes

Fixes #5723

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -856,7 +856,8 @@ func UpdateUser(id string, user *User, columns []string, isAdmin bool) (bool, er
 		}
 	}
 
-	if len(columns) == 0 {
+	columnsWereEmpty := len(columns) == 0
+	if columnsWereEmpty {
 		columns = []string{
 			"owner", "display_name", "avatar", "first_name", "last_name",
 			"location", "address", "addresses", "country_code", "region", "language", "affiliation", "title", "id_card_type", "id_card", "homepage", "bio", "tag", "language", "gender", "birthday", "education", "score", "karma", "ranking", "signup_application", "register_type", "register_source",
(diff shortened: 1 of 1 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/061c36f9e519b72a86fb754ab5074a2643c2fb5c

PR #5726, opened 2026-08-14
fix: revoke access when a user is forbidden

### LINKED ISSUE #5725: Token validation ignores IsForbidden for already issued access tokens
## Bug

When an administrator sets `IsForbidden = true` for a user, Casdoor correctly blocks new sign-in and refresh-token grants. However, already issued access tokens remain valid during Casdoor-side token validation:

- `/api/login/oauth/introspect` can still return `active: true`
- `AutoSigninFilter` still accepts the Bearer / `access_token`
- MCP auth still derives claims from the token

As a result, a forbidden user can keep using an existing access token until it expires.

## Root cause

Token validation paths check token existence and `expires_in`, but do not load the end-user or check `user.IsForbidden`. Per RFC 7662, a token for a disabled/forbidden user may be treated as inactive.

## Expected

If the token owner is forbidden (or the user no longer exists), Casdoor token validation should treat the token as inactive / reject it:

- introspection → `{ "active": false }`
- auto-signin / MCP → reject the access token

`client_credentials` tokens (no end-user) should remain unaffected.

## Actual

A forbidden user's existing access token continues to pass introspection and Casdoor Bearer validation until natural expiry or manual token deletion.

### LINKED ISSUE #5727: Forbidding a user does not terminate sessions and tokens
## Bug

When an administrator sets `IsForbidden = true` (or soft-deletes a user), Casdoor blocks new sign-in but does not end existing sessions or expire existing OAuth tokens.

Observed effects:

- Session rows remain visible in the Sessions admin UI
- Refresh may already fail (forbidden check on refresh), but access tokens stay in the token store until natural expiry or manual deletion
- This is weaker than SSO logout-all, which already expires tokens and deletes all sessions

Related: #5725 (token validation should treat forbidden users as inactive). Even with validation checks, sessions should still be cleaned up when an account is forbidden.

## Root cause

`UpdateUser` persists `is_forbidden` / `is_deleted` but does not call the same termination path used by SSO logout-all (`ExpireTokenByUser` + `DeleteAllUserSessions`). Those helpers exist but are only wired to logout / delete-user flows.

## Expected

When `IsForbidden` transitions from `false` to `true` (and similarly for soft-delete):

1. Expire all of the user's OAuth tokens (`expires_in = 0`)
2. Delete all of the user's sessions across applications (including Beego session destroy)
3. Optionally notify RPs (back-channel / SSO logout notifications), consistent with logout-all

Unforbidding a user must not recreate sessions/tokens.

## Actual

Only the user flag is updated; active sessions and token rows remain until logout, manual deletion, or expiry.

maximaa2001:
## Summary
Revoke an account's access immediately when it is forbidden (and keep already-issued tokens inactive during Casdoor validation).

1. **Terminate sessions and tokens on forbid / soft-delete**  
   When `IsForbidden` (or soft-delete) transitions to `true`, call the same path as SSO logout-all: expire tokens, delete all sessions, and send logout notifications.
2. **Treat forbidden owners as inactive during token validation**  
   Add `Token.IsOwnerActive()` and use it in introspection, auto-signin, and MCP auth so leftover access tokens are rejected even if a row still exists.
3. **Leave refresh behavior as-is**  
   Refresh already refused forbidden users; this closes the gap for existing sessions, token rows, and access-token validation.

## Test plan
- [ ] Forbid a signed-in user → Sessions UI no longer shows their sessions
- [ ] Tokens for that user have `expires_in = 0` (or are gone from active use)
- [ ] Introspect a previously issued access token → `active: false`
- [ ] Unforbid the user → no sessions/tokens are recreated automatically
- [ ] Soft-delete also terminates sessions/tokens
- [ ] `go test ./object/ -run 'TestShouldTerminateUserAccess|TestTokenIsOwnerActive'` passes

Fixes #5725  
Fixes #5727

### DIFF
--- a/controllers/token.go
+++ b/controllers/token.go
@@ -606,6 +606,16 @@ func (c *ApiController) IntrospectToken() {
 	}
 
 	if token != nil {
+		ownerActive, err := token.IsOwnerActive()
+		if err != nil {
+			c.ResponseTokenError(object.InvalidRequest, err.Error())
+			return
+		}
+		if !ownerActive {
+			respondWithInactiveToken()
+			return
+		}
+
 		application, err = object.GetApplication(fmt.Sprintf("%s/%s", token.Owner, token.Application))
 		if err != nil {
 			c.ResponseTokenError(object.InvalidClient, err.Error())
--- a/mcpself/auth.go
+++ b/mcpself/auth.go
@@ -154,6 +154,11 @@ func (c *McpController) GetClaimsFromToken() *object.Claims {
 		return nil
 	}
 
+	ownerActive, err := token.IsOwnerActive()
+	if err != nil || !ownerActive {
+		return nil
+	}
+
 	application, err := object.GetApplication(token.Application)
 	if err != nil || application == nil {
 		return nil
--- a/object/token.go
+++ b/object/token.go
@@ -111,6 +111,28 @@ func GetTokenByAccessToken(accessToken string) (*Token, error) {
 	return &token, nil
 }
 
+// IsOwnerActive reports whether the end-user for this token may still use it.
+// client_credentials tokens are not bound to a user account. Per RFC 7662, a
+// token for a forbidden/disabled user is treated as inactive.
+func (token *Token) IsOwnerActive() (bool, error) {
+	if token == nil {
+		return false, nil
+	}
+	if token.GrantType == "client_credentials" || token.User == "" {
+		return true, nil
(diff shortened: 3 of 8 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/b505e642f8df6723f5d1c3f5c2f6ee4d95851ec2

PR #5738, opened 2026-08-18
feat: link OAuth tokens to Beego session for session-scoped revoke

### LINKED ISSUE #5737: Deleting a Beego session does not revoke OAuth tokens minted in that session
## Problem

Casdoor stores **Beego sessions** (`Session` rows + cookie) and **OAuth tokens** (`Token` rows) separately. There is no link between them.

When an admin removes a user session from the Casdoor UI, or when logout runs without a token hint, only the Beego session is destroyed. OAuth token rows keep `expires_in > 0`, so previously issued access/refresh tokens remain valid on Casdoor (refresh, introspection, Bearer against Casdoor APIs) until natural expiry.

Single-token logout via `GET /api/logout?id_token_hint=...` works because it calls `ExpireTokenByAccessToken` for one row. Logout-all uses `ExpireTokenByUser`. **Session-scoped** revoke is missing.

## Steps to reproduce

1. Sign in via OAuth authorization code (browser session + code exchange).
2. Confirm the access token works (e.g. `get-account` with Bearer).
3. In Casdoor admin, open the user's **Sessions** and delete the session used at login.
4. Read the token in admin/API (`get-token`) → **`expiresIn` is unchanged** (`> 0`).
5. Refresh the access token (or call Casdoor with the same Bearer) → **still works** until JWT/`expires_in` lifetime ends.

Same behavior for:
- `POST /api/sso-logout?logoutAll=false` (current session only): session removed, tokens for that browser session stay active.
- `GET /api/logout` without `id_token_hint`: session removed, no token expire unless hint is present.

## Root cause

- `Token` has no field tying a row to the Beego `session_id` that minted it.
- `DeleteSession` / `DeleteSessionId` only destroy Beego sessions and DB `Session` rows; they never update `token.expires_in`.
- OAuth code/tokens are minted without recording `CruSession.SessionID()`.

## Expected

When a Beego session ends (admin delete, single-session logout, logout without hint):

1. All OAuth `Token` rows minted under that session should get `expires_in = 0`.
2. Refresh for those tokens should fail (`InvalidGrant` / expired).
3. Other sessions/devices for the same user should be unaffected.

New logins should persist the minting session id on the token row (including through refresh).

## Actual

- Session UI / Beego cookie: removed.
- Token rows: unchanged `expires_in`.
- No way to revoke "this browser login" without knowing a specific token value.

## Proposed solution

1. Add `session_id` on `Token` (indexed), set when minting (authorization code, implicit, consent, etc.).
2. Inherit `session_id` on refresh.
3. Add `ExpireTokensBySessionIds(sessionIds []string)` → `UPDATE token SET expires_in = 0 WHERE session_id IN (?) AND expires_in > 0`.
4. Call it from `DeleteSession`, `DeleteSessionId`, single-session logout paths.

maximaa2001:
Fixes #5737

## Summary

- Add `session_id` on `Token`, set when minting OAuth codes/tokens and inherited on refresh.
- Add `ExpireTokensBySessionIds` to set `expires_in = 0` for tokens linked to ending sessions.
- Revoke linked tokens when a session is deleted (admin UI), on single-session logout, and on logout without `id_token_hint`.
- Best-effort backfill of `session_id` on authorization_code exchange when the Casdoor session cookie is present.

## Test plan

- [ ] Sign in via OAuth authorization code; confirm the token row has `sessionId` in admin/API.
- [ ] Delete that session in Casdoor admin → token `expiresIn` becomes `0`; refresh fails.
- [ ] Sign in from two browsers; deleting one session does not expire the other device's token.
- [ ] `POST /api/sso-logout?logoutAll=false` expires tokens for the current session only.
- [ ] `GET /api/logout` without `id_token_hint` expires tokens for the current session.
- [ ] Refresh issues a new token row with the same `sessionId`.

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -368,7 +368,7 @@ func (c *ApiController) Signup() {
 			return
 		}
 
-		code, err := object.GetOAuthCode(userId, clientId, "", "password", responseType, redirectUri, scope, state, nonce, codeChallenge, "", c.Ctx.Request.Host, c.GetAcceptLanguage())
+		code, err := object.GetOAuthCode(userId, clientId, "", "password", responseType, redirectUri, scope, state, nonce, codeChallenge, "", c.Ctx.Request.Host, c.GetAcceptLanguage(), c.Ctx.Input.CruSession.SessionID(context.Background()))
 		if err != nil {
 			c.ResponseError(err.Error(), nil)
 			return
@@ -613,6 +613,12 @@ func (c *ApiController) SsoLogout() {
 		// Logout from current session only
 		sessionIds = []string{currentSessionId}
 
+		_, err = object.ExpireTokensBySessionIds(sessionIds)
+		if err != nil {
+			c.ResponseError(err.Error())
+			return
+		}
+
 		// Only delete the current session's Beego session
 		object.DeleteBeegoSession(sessionIds)
 
@@ -711,7 +717,8 @@ func (c *ApiController) GetAccount() {
 
 	accessToken := c.GetSessionToken()
 	if accessToken == "" {
-		accessToken, err = object.GetAccessTokenByUser(user, c.Ctx.Request.Host)
+		beegoSessionId := c.Ctx.Input.CruSession.SessionID(context.Background())
+		accessToken, err = object.GetAccessTokenByUser(user, c.Ctx.Request.Host, beegoSessionId)
 		if err != nil {
 			c.ResponseError(err.Error())
 			return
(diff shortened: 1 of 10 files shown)


nomeguy:
@maximaa2001 you have made too many issues and PRs these days. I guess 80% of the volumes come from you these weeks. This is not good for a public OSS project, because it becomes more of your "perosnal flavor" project instead of public and neutral project. Code is cheap in the AI coding era. I can do 100+ code commits one day, but I won't. Because this behavior will turn this project into my color and this will harm the project. We need more diversity for the contributions.

You should be more conservative about submitting your patches. Maybe save most of them in your own fork. Given that no other persons are feedbacking them: perhaps only you need them, not other people.

PR #5750, opened 2026-08-22
fix: fall back to alipay_open_id when user_id is empty

### LINKED ISSUE #5749: [BUG] AlipayIdProvider fails to get user_id for new apps, only returns openid
The current AlipayIdProvider implementation in Casdoor relies on the user_id field from the Alipay API to populate the user's Id. However, this causes a critical failure for newly registered Alipay apps.

According to the official Alipay documentation ([openid 简介](https://opendocs.alipay.com/mini/0ai2i6?pathHash=13dd5946)), there is a clear notice:

"注意：支付宝未来推出的新产品将只支持 openid。" (Note: New Alipay products launched in the future will only support openid.)

For these new apps, the user_id field is no longer returned in any API responses. Only the alipay_open_id is provided. This breaks the authentication flow for any new Casdoor users trying to link their Alipay account, resulting in the error:

Failed to link user account

Affected Code

The issue originates in the idp/alipay.go file, where the UserInfo.Id is directly mapped from the absent user_id field:

type AlipayUserInfoShareResponse struct {
    // ...
    UserId   string `json:"user_id"`  // ← This field is empty for new apps
}

userInfo := UserInfo{
    Id:          resp.UserId,  // ← Receives empty value, causing link failure
    // ...
}
Proposed Solution

I kindly request that the Casdoor team considers adding support for the alipay_open_id field to ensure compatibility with new Alipay applications. This will resolve the linking error for new users and align Casdoor with Alipay's future direction.

Thank you for your time and for maintaining this great project.

New app (can't link)：

Image

Image

Image

Old app (can be linked)：

Image

BetterAndBetterII:
Fixes #5749

New Alipay apps omit `user_id` on `alipay.user.info.share` and only return `alipay_open_id`. `GetUserInfo` mapped `Id` from `user_id` alone, so linking failed with an empty id.

This parses `alipay_open_id` and uses it when `user_id` is empty. Legacy `user_id` responses are unchanged.

```bash
go test ./idp -count=1 -run TestAlipayGetUserInfoId
```

### DIFF
--- a/idp/alipay.go
+++ b/idp/alipay.go
@@ -181,11 +181,12 @@ type AlipayUserResponse struct {
 }
 
 type AlipayUserInfoShareResponse struct {
-	Code     string `json:"code"`
-	Msg      string `json:"msg"`
-	Avatar   string `json:"avatar"`
-	NickName string `json:"nick_name"`
-	UserId   string `json:"user_id"`
+	Code         string `json:"code"`
+	Msg          string `json:"msg"`
+	Avatar       string `json:"avatar"`
+	NickName     string `json:"nick_name"`
+	UserId       string `json:"user_id"`
+	AlipayOpenId string `json:"alipay_open_id"`
 }
 
 // GetUserInfo Use access_token to get UserInfo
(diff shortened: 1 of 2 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/e6b2338f8955455254ec2f48ef0a0390cbef9ec5

PR #5757, opened 2026-08-24
fix: make the web build's build-temp swap non-destructive

ThomasM-PX:
## Summary
- `web/mv.js`'s `postbuild` step deleted `web/build` *before* renaming `build-temp` into place, so any failure partway through the rename left the app with no build directory at all.
- On Windows this is easy to hit: a directory can stay in a "delete pending" state while another process holds a handle to a file inside it, and if the running server is serving static files straight out of `build/`, the following `renameSync` fails with `EPERM`.
- Reworked the swap to rename the previous build aside, move the new build into place, and only then remove the backup — restoring it automatically if the swap fails. Both the renames and the removals retry on `EPERM`/`EBUSY`/`EACCES`/`ENOTEMPTY`, and a stale backup left by an earlier interrupted run is cleaned up on the next run.

## Test plan
- [x] `yarn build` completes and swaps `build-temp` into `build` correctly with no server holding a lock on the directory (verified: clean run, no leftover `build-temp`/`build-old`, served assets match the new hashes)
- [ ] Verified under real lock contention (server actively serving from `build/` during the build) — the retry loop and rollback path are implemented per Node's documented `EPERM`/`EBUSY` retry semantics but have not yet been exercised by an actual failing rename in this environment

### DIFF
--- a/web/mv.js
+++ b/web/mv.js
@@ -3,6 +3,35 @@ const path = require("path");
 
 const sourceDir = path.join(__dirname, "build-temp");
 const targetDir = path.join(__dirname, "build");
+const backupDir = path.join(__dirname, "build-old");
+
+// On Windows a directory can linger in a "delete pending" state while another
+// process still holds a handle to a file inside it: the Casdoor server serves
+// files straight out of build/, and virus scanners open freshly written ones.
+// Removes and renames then fail with EPERM/EBUSY, so both are retried here.
+const RETRY_ATTEMPTS = 10;
+const RETRY_DELAY_MS = 200;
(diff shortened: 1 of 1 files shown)


hsluoyz:
@ThomasM-PX Thanks for catching this — deleting `build/` before the rename was indeed wrong. `web/` has since been replaced by the shadcn/ui frontend, so this patch no longer applies, but I've taken the same rename-aside approach into the new `web/mv.js` in bc5f2702bb55177e43b12c914ed935d910380832. Closing — thanks again!

PR #5758, opened 2026-08-24
fix: make the syncer Add button navigate to a valid route

ThomasM-PX:
## Summary
- The Add button on the syncer list page (`SyncerListPage.js`) pushed `/syncers/<name>`, but the route registered in `ManagementPage.js` is `/syncers/:organizationName/:syncerName`. The navigation matched no route, so clicking Add rendered an in-app 404 with no network request ever made — it looks like a backend or deployment issue but isn't.
- Added the missing `organization` segment, matching every other syncer link in the same file (`Edit` button, table link) and the equivalent Add buttons on the model, cert and adapter list pages, which already do this correctly.
- `SyncerEditPage` reads both `organizationName` and `syncerName` from route params and needs both to load/create the syncer.

## Test plan
- [x] Clicking "Add" on the syncer list now opens the syncer edit page instead of a 404
- [x] Confirmed the new path matches the registered route and the pattern used by `ModelListPage`, `CertListPage`, and `AdapterListPage`
- [x] Confirmed via decompiled production bundle that the fix (two-segment path) is what ships, matching the registered route

### DIFF
--- a/web/src/SyncerListPage.js
+++ b/web/src/SyncerListPage.js
@@ -50,7 +50,7 @@ class SyncerListPage extends BaseListPage {
 
   addSyncer() {
     const newSyncer = this.newSyncer();
-    this.props.history.push({pathname: `/syncers/${newSyncer.name}`, mode: "add", syncer: newSyncer});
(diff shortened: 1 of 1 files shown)


hsluoyz:
already fixed

PR #5760, opened 2026-08-25
fix: prevent account enumeration in forgot-password flow

dshilovprpl:
## Summary

Prevent account enumeration in the forgot-password flow when `enableErrorMask` is enabled.

## Problem

The existing masking policy only replaced selected error messages. The forgot-password flow still returned structurally different responses depending on whether an account existed:

- The initial account lookup returned contact information for an existing user and an error for an unknown user.
- The verification-code endpoint returned account-dependent errors.
- Replacing every response with an error would prevent legitimate users from completing password recovery.

A response-only middleware is therefore insufficient because the recovery flow still needs a real destination and user identity after the public response has been masked.

## Solution

- Return the same public contact placeholders and response status for every supplied identifier.
- Keep the real recovery user and destination in the server-side session.
- Send a verification code only when an eligible account and provider exist, while returning the same public response in all cases.
- Bind code verification and password reset to the user stored in the recovery session.
- Make Dynamic CAPTCHA behavior independent of account existence.
- Apply this behavior only to `method=forget` when `enableErrorMask` is enabled.

Other verification flows, including login, signup, and MFA, are unchanged. Behavior also remains unchanged when `enableErrorMask` is disabled.

## Testing

- Added regression tests for:
  - forgot-password masking activation;
  - fixed public contact placeholders;
  - email/SMS delivery selection and fallback;
  - account-independent Dynamic CAPTCHA behavior.

- Full controller test suite:

  ```text
  go test ./controllers -count=1
  ok github.com/casdoor/casdoor/controllers

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -1512,10 +1512,15 @@ func (c *ApiController) GetCaptchaStatus() {
 	}
 
 	clientIp := util.GetClientIpFromRequest(c.Ctx.Request)
-	captchaEnabled, err := object.CheckToEnableCaptcha(application, organization, userId, clientIp)
-	if err != nil {
-		c.ResponseError(err.Error())
-		return
+	var captchaEnabled bool
+	if c.isMaskedPasswordRecoveryCaptchaRequest(organization, userId) {
+		captchaEnabled = shouldEnableMaskedPasswordRecoveryCaptcha(application, clientIp)
+	} else {
+		captchaEnabled, err = object.CheckToEnableCaptcha(application, organization, userId, clientIp)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


dshilovprpl:
@hsluoyz Could you please review? I've sent email to the admin@casdoor.org, but didn't receive any response :(

nomeguy:
send issue first

dshilovprpl:
> send issue first

@nomeguy 
This is security issue and according SECURITY.md we shouldn't create issue in GitHub issues. 

I've already sent email with this security issue and now prepared PR. 
Could you please re-open PR? Or I should avoid to use SECURITY rules and prepare GitHub issue for it?

PR #5768, opened 2026-08-28
feat: make LDAP/AD forgot password restriction configurable

### LINKED ISSUE #5745: Feature request: Make LDAP/AD forgot password restriction configurable
Hi Casdoor team,

I would like to request a configurable option for the change introduced in #5676.

Our environment uses Microsoft Active Directory Domain Services (AD DS) running on Windows Server 2022. Casdoor connects to our AD domain through LDAPS, and our users are synchronized from Active Directory to Casdoor.

Our current architecture is:

```text
Casdoor
  ↓ LDAPS
Windows Server 2022
Microsoft Active Directory Domain Services (AD DS)
  ↓
AD Domain Users
```

In our environment, the LDAPS bind/service account has the necessary permissions to reset user passwords in Active Directory.
This is not a theoretical requirement. We have already verified in our production environment that the Casdoor Forgot Password flow can successfully reset passwords for our Microsoft Active Directory users through LDAPS before this change.

Therefore, the existing Casdoor forgot-password flow works correctly:

1. An AD user clicks "Forgot Password" in Casdoor.
2. Casdoor sends the verification code.
3. The user completes the verification.
4. Casdoor resets the user's password through LDAPS.
5. The password is successfully updated in Microsoft Active Directory.

Because this workflow already works correctly in our Microsoft AD + LDAPS environment, the change in #5676 creates a regression for us.

The current implementation rejects all users with `user.Ldap != ""` before sending the forgot-password verification code. This assumes that LDAP users cannot reset their passwords through Casdoor, but this is not true for all LDAP implementations.

For example, Microsoft Active Directory over LDAPS can support password resets when the configured bind/service account has the appropriate permissions.

## Proposed solution

Could this restriction be made configurable instead of being enforced for all LDAP users?

Ideally, Casdoor could add an option to the LDAP provider configuration, for example:

```text
Enable Forgot Password for LDAP Users
[ ] Allow
```

Or:

```text
Allow Password Reset via LDAP
[ ] Enable
```

Recommended behavior:

* Disabled by default, preserving the current behavior introduced by #5676.
* Administrators can explicitly enable it when their LDAP/AD server and bind account support password reset.
* When enabled, LDAP users can continue using Casdoor's existing Forgot Password flow.

A per-LDAP-provider option would be preferable to a global switch because different LDAP providers may have different capabilities and permission configurations.

This would preserve the purpose of #5676 while also supporting environments like ours, where Microsoft Active Directory on Windows Server 2022 with LDAPS is already able to reset user passwords successfully.

Could the team consider adding such a configuration option?

Thank you!

wangweiningse:
The check added in c13cd7a rejects all LDAP users from the forgot-password flow. But LDAP servers like Microsoft AD over LDAPS can reset passwords when the bind account has the permission.

Add an "Enable password reset" option to each LDAP server config (default off, keeping the current behavior). Administrators can turn it on to let LDAP users use the forgot-password flow again.

Fixes #5745

### DIFF
--- a/object/ldap.go
+++ b/object/ldap.go
@@ -41,6 +41,8 @@ type Ldap struct {
 	AutoSync     int    `json:"autoSync"`
 	LastSync     string `xorm:"varchar(100)" json:"lastSync"`
 	EnableGroups bool   `xorm:"bool" json:"enableGroups"`
+
+	EnablePasswordReset bool `xorm:"bool" json:"enablePasswordReset"`
 }
 
 func AddLdap(ldap *Ldap) (bool, error) {
@@ -154,7 +156,7 @@ func UpdateLdap(ldap *Ldap) (bool, error) {
 	}
 
 	affected, err := ormer.Engine.ID(ldap.Id).Cols("owner", "server_name", "host",
-		"port", "enable_ssl", "username", "password", "base_dn", "filter", "filter_fields", "auto_sync", "default_group", "default_groups", "password_type", "allow_self_signed_cert", "custom_attributes", "enable_groups").Update(ldap)
+		"port", "enable_ssl", "username", "password", "base_dn", "filter", "filter_fields", "auto_sync", "default_group", "default_groups", "password_type", "allow_self_signed_cert", "custom_attributes", "enable_groups", "enable_password_reset").Update(ldap)
 	if err != nil {
 		return false, nil
 	}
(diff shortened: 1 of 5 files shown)


wangweiningse:
image
image

hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/bfe03536e3fee86386acdd4f64a047b72d5a8ff5

PR #5772, opened 2026-08-28
fix(verification): pre-check email and phone existence before sending signup verification code (fix #5770)

### LINKED ISSUE #5770: Signup page sends a verification code email even when the email is already registered (wasted send, error only surfaces on submit)
## Description

On the signup page, clicking **Get Code** for an email that is **already registered** sends the verification code email successfully. The duplicate-email error ("check:Email already exists") only appears after the user submits the signup form — so every signup attempt with an existing email wastes one paid email send (e.g. Alibaba Cloud DirectMail charges per message).

## Reproduction

1. Register user `foo@example.com` on any application with an Email provider.
2. Go to the signup page, fill anything, enter the **already registered** `foo@example.com`, click **Get Code**.
3. `POST /api/send-verification-code?type=register` returns `{"status":"ok"}` and the email is actually sent.
4. Only `POST /api/signup` then fails with "Email already exists".

## Root cause

In `controllers/verification.go` `SendVerificationCode()`, the `VerifyTypeEmail` branch only resolves/validates the user for `LoginVerification`, `ForgetVerification`, `ResetVerification` and `MfaAuthVerification`. For the default signup method there is no `HasUserByField(org, "email", dest)` check, so the code is generated and sent regardless.

## Suggested fix

In the `VerifyTypeEmail` branch, when the method is the default signup flow, reject with `check:Email already exists` when `object.HasUserByField(organization.Name, "email", dest)` is true (mirroring `CheckUserSignup`). Symmetrically, the login/forgot flows already reject unknown emails with "the user does not exist, please sign up first" — signup should reject known emails the same way, before spending the email send.

## Environment

- casdoor/casdoor:latest docker image (2026-08)
- Signup items: Email visible + required, verified by code

Jaymr1031-Z:
Fixes #5770

### Problem
On the signup page, clicking **Get Code** for an email or phone number that is already registered would still trigger sending the verification code (wasting paid email/SMS quota). The duplicate check was only performed when the signup form was submitted.

### Solution
In controllers/verification.go SendVerificationCode():
- For VerifyTypeEmail, pre-check if the email already exists in the organization (object.HasUserByField(organization.Name, email, normalizedEmail)) when method is signup. If it exists, immediately return check:Email already exists and skip sending.
- For VerifyTypePhone, pre-check if the phone already exists in the organization (object.HasUserByPhoneAndCountryCode(organization.Name, dest, countryCode)) when method is signup. If it exists, immediately return check:Phone already exists and skip sending.

### DIFF
--- a/controllers/verification.go
+++ b/controllers/verification.go
@@ -309,7 +309,12 @@ func (c *ApiController) SendVerificationCode() {
 			return
 		}
 
-		if vform.Method == LoginVerification || vform.Method == ForgetVerification {
+		if vform.Method == SignupVerification || vform.Method == "" || vform.Method == "register" {
+			if object.HasUserByField(organization.Name, "email", strings.ToLower(vform.Dest)) {
+				c.ResponseError(c.T("check:Email already exists"))
+				return
+			}
+		} else if vform.Method == LoginVerification || vform.Method == ForgetVerification {
 			if user != nil && util.GetMaskedEmail(user.Email) == vform.Dest {
 				vform.Dest = user.Email
 			}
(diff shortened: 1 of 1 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/44a81f4ed53ae6c4d61402e41046e6e13e6ce8b6

PR #5779, opened 2026-09-01
feat: add API to get direct and group roles for user

mifrun:
Add an API for retrieving roles assigned to a user directly and through groups. The query only loads roles related to the requested user.

### DIFF
--- a/controllers/role.go
+++ b/controllers/role.go
@@ -84,6 +84,31 @@ func (c *ApiController) GetRole() {
 	c.ResponseOk(role)
 }
 
+// GetUserRoles
+// @Title GetUserRoles
+// @Tag Role API
+// @Description get the user's direct roles and roles assigned through groups
+// @Param   owner  query    string  true        "The owner of the user"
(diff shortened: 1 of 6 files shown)


sandro78:
Guys (maintainers) please review - this is really useful API. On the moment there is no way to distinguish which roles were got directly on user and which were inherited.

hsluoyz:
Thanks for the PR, and thanks @sandro78 for the ping — the need is real, there was no way to tell a directly-assigned role from an inherited one.

We went with a slightly different shape so we don't have to add another API: each role in `user.roles` (returned by `/api/get-user`) now carries a `sourceGroups` field. Empty means the role was assigned to the user directly, otherwise it lists the groups that granted it. This also covers roles inherited through sub-roles, which the endpoint here didn't handle, and the user edit page marks those roles now too. Done in https://github.com/casdoor/casdoor/commit/b6e2abdc4876e3db4e8ea76a336b60e1e81d2294

Closing this one — thanks again for raising it!

Issue #5799, opened 2026-09-04
[BUG] OIDC provider login fails with "Unknown column 'oidc' in 'where clause'" — missing `Oidc` field in User struct


developerchen1012:
When a provider is configured with type OIDC in the frontend, authenticating fails with the following MySQL error:                                                                                       
Error 1054 (42S22): Unknown column 'oidc' in 'where clause'                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                    
The root cause is that the User struct has no Oidc field, so never creates an oidc column in the user table 

Workaround

  Manually adding an oidc column to the user table (e.g. ALTER TABLE user ADD COLUMN oidc varchar(100);) stops the error and allows login to proceed. However, this is an unverified manual workaround — it is unclear whether other code paths (e.g. account linking, user listing, field display in admin UI) fully support this column, since the User struct has no corresponding Oidc field. Official confirmation and a proper fix are still needed.

HsukqiLee:
Confirmed the root cause on current master: for a provider of type OIDC, getUserByProvider falls through to GetUserByField(organization, provider.Type, providerId) (controllers/auth.go), which issues WHERE oidc = ? (strings.ToLower(field) in GetUserByField). The User struct has a dedicated linkage column for every built-in provider type except a generic oidc column, so xorm Sync2 never creates it and MySQL returns error 1054. Opened #5801 to add the Oidc field (auto-created by Sync2) and include it in the UpdateUser column list and the frontend UserFields, so lookup, account linking, and the admin UI all support it without the manual ALTER TABLE workaround.

hsluoyz:
fixed in: https://github.com/casdoor/casdoor/commit/ebb307cf03ccba051a18d7d314dc67c65076f140

PR #5801, opened 2026-09-04
fix: add missing Oidc field to User struct for OIDC provider login

HsukqiLee:
## What does this PR do?

Fixes the OIDC provider login failure reported in #5799:

```
Error 1054 (42S22): Unknown column 'oidc' in 'where clause'
```

### Root cause

When signing in with a provider of type `OIDC`, `getUserByProvider` falls through to `GetUserByField(organization, provider.Type, providerId)` (`controllers/auth.go`), which issues `WHERE oidc = ?` via `strings.ToLower(field)`. The `User` struct has a dedicated linkage column for every built-in provider type (`github`, `google`, `okta`, `azuread`, ...) **except** a generic `oidc` column, so xorm's `Sync2` never creates it and MySQL returns error 1054.

### Fix

- Add `Oidc string `xorm:"oidc varchar(100)" json:"oidc"`` to the `User` struct, mirroring the existing per-provider columns. On startup, xorm `Sync2` (`object/ormor.go`) auto-creates the `oidc` column, so no manual `ALTER TABLE` is needed.
- Add `"oidc"` to the default updatable-columns list in `UpdateUser`, so account linking via `LinkUserAccount` / `SetUserField` (`bean["oidc"] = providerId`) can persist the value on the normal update path.
- Add `"oidc"` to the frontend `UserFields` list so the admin UI treats the field consistently with the other provider linkage fields.

With the column present, both the lookup (`WHERE oidc = ?`) and the link-back succeed, and OIDC login proceeds without the manual workaround from the issue.

## Testing

- `go build ./...` passes.
- `go vet ./object/` clean.
- The repository's `go test` suite requires a live MySQL (`127.0.0.1:3306`) which is not available in this environment; the schema/column logic is exercised by CI.

### DIFF
--- a/object/user.go
+++ b/object/user.go
@@ -144,6 +144,7 @@ type User struct {
 	Steam           string `xorm:"steam varchar(100)" json:"steam"`
 	Bilibili        string `xorm:"bilibili varchar(100)" json:"bilibili"`
 	Okta            string `xorm:"okta varchar(100)" json:"okta"`
+	Oidc            string `xorm:"oidc varchar(100)" json:"oidc"`
 	Douyin          string `xorm:"douyin varchar(100)" json:"douyin"`
 	Kwai            string `xorm:"kwai varchar(100)" json:"kwai"`
 	Line            string `xorm:"line varchar(100)" json:"line"`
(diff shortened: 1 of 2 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/ebb307cf03ccba051a18d7d314dc67c65076f140

HsukqiLee:
Thanks @hsluoyz — confirmed the fix in ebb307c covers the same three spots this PR touched (the `Oidc` struct field with `xorm:"oidc varchar(100)"`, the `UpdateUser` default column list, and the frontend `UserFields`), so OIDC login and account linking should work without the manual `ALTER TABLE` now. Closing in favor of that commit.

PR #5802, opened 2026-09-05
fix(apple): capture the real name/email from Sign in with Apple

### LINKED ISSUE #3464: The account information registered after Apple OAuth login should use Apple's user information
After using Apple Oauth login, the account information, name, and displayName created by Casdoor are all before the mailbox @. Can I use the name returned by my Apple account as the name to create an account?

thestumonkey:
## The bug

Sign-in-with-Apple accounts have always gotten an invented display name and
username — the email local part, or Apple's opaque `sub` — instead of the
real name, even when the user has one on their Apple ID. This has been
reported before with no fix: #3464.

## Root cause (two bugs stacked)

1. `idp/goth.go`'s `"Apple"` case in `NewGothIdProvider` never requests the
   `"name"`/`"email"` scopes:
   ```go
   Provider: apple.New(clientId, *secret, redirectUrl, nil),
   ```
   Without them, goth's Apple provider never sets `formPostResponseMode`
   (see `configure()` in `providers/apple/apple.go`), Apple is never asked
   for `response_mode=form_post`, and Apple's one-time `user` payload — the
   **only** place it ever sends the real name, and only on an account's
   **very first** authorization ever — is never sent at all.

2. Requesting the scopes alone would not be enough. The POST `/api/callback`
   route (`web.Router("/api/callback", &controllers.ApiController{}, "POST:Callback")`)
   is clearly meant for exactly this form_post case, but the handler only
   ever read `code` and `state`:
   ```go
   func (c *ApiController) Callback() {
       code := c.GetString("code")
       state := c.GetString("state")
       frontendCallbackUrl := fmt.Sprintf("/callback?code=%s&state=%s", ...)
       c.Ctx.Redirect(http.StatusFound, frontendCallbackUrl)
   }
   ```
   Apple's `user` field was silently discarded. And goth's own `apple.Session`
   has no field for a name either — it only decodes the ID token, which
   carries `email` but never a name (`providers/apple/session.go`). So even
   with the scope fixed, the name would still never reach `getUser`.

## The fix

- Request `apple.ScopeName` and `apple.ScopeEmail` when constructing the
  Apple provider.
- Thread the OAuth `state` through `ProviderInfo` → `GothIdProvider` so two
  separate HTTP requests can be correlated: Apple's direct POST to
  `/api/callback` (which now parses the `user` JSON and stashes it in a new
  `AppleUserInfoCache`, keyed by `state`), and the frontend's later POST to
  `/api/login` (which is where `GetUserInfo`/`getUser` run today).
- `getUser`'s `"apple"` branch now does `AppleUserInfoCache.LoadAndDelete(state)`
  and prefers the cached real name/email over the existing
  `GetUsernameFromEmail` fallback for both `DisplayName` and `Username`,
  when present. `LoadAndDelete` makes each entry single-use, matching
  Apple's own one-time delivery — a returning user (no cache entry) falls
  back to exactly the old behavior, unchanged.

Scoped narrowly to the `"apple"` case only — no other goth-backed provider
(Slack, Steam, Line, Amazon, ...) changes behavior, and no new dependencies
were added.

## Testing

- `go build ./...`, `go vet ./...`, and `gofmt -l` are all clean.
- Added `idp/goth_apple_test.go` covering the new `getUser` logic in
  isolation (no cache entry / a fresh entry for this state / an entry
  cached under a *different* state that must not leak across), since this
  part is a pure function testable without live credentials.
- **Not verified against a real Apple sign-in.** That needs a live Apple
  Developer Sign in with Apple key and a real device/browser flow, which I
  don't have available here. The two most Apple-specific pieces I'd
  especially appreciate a maintainer or CI double-checking against a real
  account: whether `response_mode=form_post` actually round-trips cleanly
  end to end with the existing `redirectUrl` handling in this codebase, and
  the exact shape of the `user` JSON Apple sends (I've matched Apple's
  documented `{"name":{"firstName","lastName"},"email"}` shape, but haven't
  been able to confirm it against a live payload).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CFVNgE6Cpmm6yZH5EjVhdg

### DIFF
--- a/controllers/auth.go
+++ b/controllers/auth.go
@@ -904,6 +904,7 @@ func (c *ApiController) Login() {
 				return
 			}
 			idpInfo.CodeVerifier = authForm.CodeVerifier
+			idpInfo.State = authForm.State
 			var idProvider idp.IdProvider
 			idProvider, err = idp.GetIdProvider(idpInfo, authForm.RedirectUri)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/75420fc5406672afb9a97f4541963a0e8d22ad75

PR #5803, opened 2026-09-05
feat: add MailKite email provider

bucabay:
## What this adds

A **MailKite** email provider, so Casdoor can send verification codes, invitation and reset mail through [MailKite](https://mailkite.dev) — same shape as the Resend provider added in #5200.

- `email/mailkite.go` — `MailKiteEmailProvider` implementing `EmailProvider.Send()`. It POSTs to `https://api.mailkite.dev/v1/send` with a Bearer API key (`from`, `to`, `subject`, `html`). It uses `net/http` and `proxy.DefaultHttpClient` like the Custom HTTP provider, so **no new dependency** in `go.mod`.
- `email/provider.go` — `case "MailKite"` in `GetEmailProvider`.
- `web` + `web-old` — `MailKite` added to the Email provider type list and to the same field-visibility conditions as `Resend` (Secret key label, hidden Client ID / Host / Port / SSL rows, no "Test SMTP Connection" button). Both frontends changed identically, per the `web-old`-is-the-reference rule.

The provider type logo uses `social_default.png` for now (same as Custom HTTP Email); if you'd like an `email_mailkite.png` in `casbin/static` I'll open that follow-up.

## How to configure

Providers → Add → Category **Email**, Type **MailKite**:

- **Secret key**: a MailKite API key (`mk_live_…`) from https://app.mailkite.dev
- **From address / From name**: as for the other API email providers (the sender domain has to be verified in the MailKite account)

## How I tested it

- `go build ./...`, `go vet ./email/`, and `gofumpt -l email/` (the repo's `.golangci.yml` formatter) — all clean.
- A local `httptest` harness for `Send()` (request method/path, `Authorization: Bearer …`, `Content-Type: application/json`, `from` rendered as `Name <addr>` and as a bare address when no name is set, multiple recipients, and a non-2xx response surfaced as an error with the body). All three cases pass. **Not committed**, since `CLAUDE.md` asks for no `_test.go` files unless requested — happy to add it if you'd like it in the tree.
- `web`: `yarn typecheck && yarn lint` clean. `web-old`: eslint on the three changed files clean.

## Disclosure

I work on MailKite. We'll maintain this provider going forward, and I'll add a `docs/provider/email/mailkite` page to `casdoor-website` once this is in.

### DIFF
--- a/email/mailkite.go
+++ b/email/mailkite.go
@@ -1,0 +1,83 @@
+// Copyright 2026 The Casdoor Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package email
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/casdoor/casdoor/proxy"
+)
+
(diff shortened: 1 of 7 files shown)


nomeguy:
closed due to advertising

PR #5818, opened 2026-09-08
feat: support configurable password history on password change

### LINKED ISSUE #5817: Support configurable password history on password change
After [#4134](https://github.com/casdoor/casdoor/pull/4134), Casdoor rejects a new password that is identical to the current one during `SetPassword`. That closes the simplest reuse case, but it does not prevent reusing an older password from before the last change.

## Problem

With the current behavior, a user can cycle passwords to bypass the check:

1. Change password from `A` to `B` (allowed — `B` differs from current `A`).
2. Change password from `B` back to `A` (allowed — `A` differs from current `B`).

The user ends up with the same password as before, even though the intent of password-change policy is to avoid reuse of recent passwords, not only the immediately previous one.

Many security baselines expect the new password to differ from the current one **and** from a configurable number of previous passwords. Casdoor currently has no equivalent beyond the current-password check introduced in #4134.

## Background from #4134

The first version of [#4134](https://github.com/casdoor/casdoor/pull/4134) implemented full password history storage. During review, [@dacongda](https://github.com/dacongda) raised two concerns:

1. **Algorithm/salt changes** — if the organization password type or salt changes, historical hashes may become hard to validate correctly.
2. **Safety** — storing previous password material (even hashed) increases stored credential surface.

The PR was then reduced to a single check: **new password must differ from the current password**. Password history storage was removed before merge.

Those concerns are valid for a naive implementation. However, they do not necessarily rule out a carefully scoped history feature:

- Only **hashes** are stored, never plaintext.
- Each history entry records the **hashing algorithm (`type`) and salt** used at the time, so reuse checks still work after an organization migrates to a different password type.
- History is **not exposed** through the User JSON API.
- The feature is **opt-in per organization** and can default to the current #4134 behavior (`0` / `1` = check current password only; no history stored).

## Expected behavior

Add an organization-level setting, e.g. `passwordHistoryCount`, that defines how many recent passwords cannot be reused, **including the current one**:

| Value | Behavior |
|-------|----------|
| `0` or `1` | Reject only if new password equals the current password (same as today after #4134). No history stored. |
| `N > 1` | Reject if new password matches the current password or any of the last `N - 1` stored hashes. Store up to `N - 1` previous hashes per user. |

On successful password change via `SetPassword`:

1. Validate the new password against the current hash and, when `N > 1`, against stored history.
2. Snapshot the current hash into history before replacing it.
3. Trim history to the configured window.

Error messages should distinguish:

- reuse of the **current** password;
- reuse of a **recent** password from history.

maximaa2001:
## Summary

- Add `Organization.passwordHistoryCount` to configure how many recent passwords cannot be reused, including the current one.
- Store previous password hashes on the user (`password_history`) with per-entry `type` and `salt` so reuse checks survive organization algorithm changes.
- Wire `CheckPasswordHistory` and `UpdatePasswordHistory` into `SetPassword`; expose the setting in the organization admin UI (`web` and `web-old`).
- Keep backward compatibility: `0` / `1` checks only the current password (same as #4134); history is stored and checked only when the count is greater than `1`.
- Closes #5817.

## Test plan

- [ ] With `passwordHistoryCount = 0` or `1`, `SetPassword` rejects reusing the current password only
- [ ] With `passwordHistoryCount = 3`, reusing any of the last three passwords is rejected
- [ ] `password_history` in the database stores at most `N - 1` entries
- [ ] Organization admin UI shows the new field and tooltip
- [ ] Upstream CI


hsluoyz:
Thanks for the PR and the detailed write-up in #5817. The problem is real, since only checking the current password is easy to get around by switching back and forth. We've implemented this in https://github.com/casdoor/casdoor/commit/d60b5ae57429ce987eccf0e616c96b38e8c5423f, which follows the same idea with a few changes. It leaves `web-old` alone, as that frontend is now read-only. A history entry with an unsupported hash type is skipped instead of blocking the password change. The count is capped at 24 to limit hashing cost, and old entries are removed when the count is lowered. The error message also says how many recent passwords are checked. Closing this in favor of that commit. Thanks again for the contribution!

PR #5819, opened 2026-09-08
fix: show syncers across all organizations

### LINKED ISSUE #5814: [Bug] Syncers list shows "No data" when organization selector is set to "All"
## Description

In the new console, the **Syncers** list shows **No data** when a global administrator selects **All** in the organization selector, even though syncers exist in a non-`built-in` organization.

The list request is scoped to `built-in` instead of querying across organizations. The saved syncers and imported users were present in the backend; the empty list did not mean that synchronization or saving had failed.

## Environment

- Deployed Casdoor version: `v4.2.0`
- Deployed commit reported by `/api/get-version-info`: `cc39a9395698ca9cf6cf869c963a641c7c015a71`
- Account: global administrator (`built-in/admin`)
- Frontend: new shadcn/ui console
- Observed with two WeCom syncers belonging to a non-`built-in` organization

The same relevant code is also present at master commit `3d51fc26f4547d74f44a36630720598341ce5fe4`. That master revision was checked by source inspection, not by running a separate deployment.

## Steps to reproduce

1. Sign in as a global administrator.
2. Create a syncer belonging to a non-`built-in` organization, for example `example-org`.
3. Select **All** in the console organization selector.
4. Open the **Syncers** list.
5. Compare the list API result with the same request explicitly selecting `example-org`, or with an empty `organization` parameter.

## Expected behavior

With **All** selected, the list should include syncers across the organizations accessible to the global administrator, consistent with the previous console.

## Actual behavior and API evidence

The list shows **No data** when there are no syncers in `built-in`, despite syncers existing elsewhere.

During read-only diagnosis on the affected deployment, the same administrator and pagination/search settings produced the following results. The organization name below is anonymized:

| Request | Returned syncers |
| --- | ---: |
| `GET /api/get-syncers?owner=admin&organization=built-in&p=1&pageSize=10&field=&value=&sortField=&sortOrder=` | 0 |
| `GET /api/get-syncers?owner=admin&organization=example-org&p=1&pageSize=10&field=&value=&sortField=&sortOrder=` | 2 |
| `GET /api/get-syncers?owner=admin&organization=&p=1&pageSize=10&field=&value=&sortField=&sortOrder=` | 2 |

The individual syncer detail endpoints also returned the saved configurations.

## Source analysis

1. The new [SyncerListPage](https://github.com/casdoor/casdoor/blob/3d51fc26f4547d74f44a36630720598341ce5fe4/web/src/pages/SyncerListPage.tsx#L13) gets `organizationName` from `useRequestOrganization()` and [passes it directly to getSyncers](https://github.com/casdoor/casdoor/blob/3d51fc26f4547d74f44a36630720598341ce5fe4/web/src/pages/SyncerListPage.tsx#L45).
2. [useRequestOrganization](https://github.com/casdoor/casdoor/blob/3d51fc26f4547d74f44a36630720598341ce5fe4/web/src/hooks/use-organization.ts#L13) calls `Setting.getRequestOrganization(account)`.
3. [getRequestOrganization](https://github.com/casdoor/casdoor/blob/3d51fc26f4547d74f44a36630720598341ce5fe4/web/src/lib/setting.tsx#L1837) returns `account.owner` when an admin selects `All`. For `built-in/admin`, that is `built-in`.
4. The [previous SyncerListPage](https://github.com/casdoor/casdoor/blob/3d51fc26f4547d74f44a36630720598341ce5fe4/web-old/src/SyncerListPage.js#L276) handled this case explicitly:

```js
Setting.isDefaultOrganizationSelected(this.props.account)
  ? ""
  : Setting.getRequestOrganization(this.props.account)
```

This comparison suggests a frontend regression caused by losing the **All** handling in the syncer list during the console migration.

## Workaround / possible fix

Select the specific organization containing the syncers.

Consider restoring the previous **All** behavior for syncer listing, while preserving explicit organization filtering and the existing restrictions for non-global administrators. A change to the shared organization helper would need to account for its other callers.

jackwalkerlabs:
Fixes #5814.

The Syncers page now sends an empty organization filter when a global administrator selects All, matching `web-old`. Selecting a specific organization still filters the list, and tenant administrators remain scoped to their own organization.

The shared organization hook gains an opt-in All mode. It computes the filter inside the storage-event callback so switching between All and built-in also refreshes the page; existing callers retain their current behavior.

Validation:
- Cypress reproduces the original All → built-in filter error on the base code.
- Two browser regression tests pass against the running Vite frontend with intercepted API responses: All → built-in → All → tenant, visible syncer rows, and tenant-admin isolation.
- Frontend lint, TypeScript checks, and production build pass.

Implemented and locally validated with OpenAI Codex. A maintainer/user will need to handle any required CLA signature; no signature was submitted by Codex.

### DIFF
--- a/web/cypress/e2e/syncers-organization.cy.js
+++ b/web/cypress/e2e/syncers-organization.cy.js
@@ -1,0 +1,81 @@
+// Copyright 2026 The Casdoor Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+describe("Syncer organization filtering", () => {
+  function openSyncers(owner, selection) {
+    cy.intercept("GET", "**/api/get-account*", {
+      status: "ok",
+      data: {owner, name: "admin", displayName: "Admin", isAdmin: true},
+      data2: {name: owner, displayName: owner, enableTour: false},
+    });
+    cy.intercept("GET", "**/api/get-organization-names*", {
+      status: "ok",
+      data: [
+        {name: "built-in", displayName: "Built-in"},
+        {name: "tenant", displayName: "Tenant"},
+      ],
+    });
+    cy.intercept("GET", "**/api/get-syncers?*", (req) => {
+      const organization = new URL(req.url).searchParams.get("organization");
+      const syncers = [
+        {owner: "admin", name: "tenant-syncer", organization: "tenant", type: "Database"},
+      ].filter((syncer) => !organization || syncer.organization === organization);
+      req.reply({status: "ok", data: syncers, data2: syncers.length});
+    }).as("syncers");
+    cy.visit("/syncers", {
+      onBeforeLoad(win) {
+        win.localStorage.setItem("organization", selection);
+        win.localStorage.setItem("language", "en");
+        win.localStorage.setItem("isTourVisible", "false");
+      },
+    });
+  }
+
+  function expectOrganization(organization) {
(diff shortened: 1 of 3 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/e8db3f94dd97527ee1d8c43a264d3b8328607289

PR #5827, opened 2026-09-10
fix: retain draft rows when editing properties

### LINKED ISSUE #5826: The add button for attributes in the group is not responding.
In v4, the add button for attributes within the group does not respond.

Image

Hanabi9248:
Fixes #5826.

Adding a property in the group editor immediately removes the empty row when the table is rebuilt from the properties map. The user editor has the same problem.

Keep editable rows in a shared PropertiesTable, as the old console did, and convert them to the record map on change. This also lets a key be cleared or temporarily duplicate another key without losing the other row. Loading a different properties object resets the table.

Checked adding multiple rows, entering a value before its key, renaming duplicate keys, clearing a key, deleting a row and reloading in a local browser harness rendering the actual group properties field. TypeScript, lint on the changed files and the Vite production build pass. The harness stubs the page data layer; backend persistence and the full Cypress suite were not run.

### DIFF
--- a/web/src/components/crud/PropertiesTable.tsx
+++ b/web/src/components/crud/PropertiesTable.tsx
@@ -1,0 +1,69 @@
+// Copyright 2026 The Casdoor Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/3eb92315410c2de44ccddf0fb217b3b0211e8faa

PR #5858, opened 2026-09-17
feat: render the forced password update inside the auth layout

oxsean:
### Why

When an account is flagged "Need update password", the sign-in stops and the browser is sent to `/account`. For a Casdoor admin that is fine. For someone who only came to sign in to another application it is not: they land in the Casdoor console, on the full profile form with the console's navigation around it, and have to find the yellow banner, then the "Modify password..." button, then a modal, before they are sent back. Applications that use Casdoor purely as their identity provider have no good way to present this step to their users, so they end up patching the sign-in page or building their own password page around the API.

Casdoor already renders sign-in, sign-up and forgot-password as standalone pages in the application's auth layout. The forced password update is the one step in that flow that did not have one.

### What this adds

**A standalone page, `/update-password[/:applicationName]`**, rendered inside the application's auth layout like the sign-in and forgot-password pages. It shows the same old / new / confirm form as the account page's password modal, validates against the organization's password options, uses the password obfuscator when one is configured, calls `set-password`, and then continues the sign-in it was interrupted from: the session survived the password change, so the page returns to the stored `signinUrl` with `silentSignin` and the sign-in page completes the authorization with it, without asking the user to pick the signed-in account again. `goToUpdatePassword()` now goes there instead of `/account`; the account page and its modal are unchanged for anyone who still opens them directly.

Default look, nothing configured:

default

**"Update password items" on the application**, so the page can be tuned per application the same way the sign-in page can. They have the same shape as the signin items (name, visible, label, placeholder, custom CSS, free-form HTML blocks) and are edited with the same table in the UI Customization tab. The default set is Logo, Languages, Title, Old password, New password, Confirm password, Submit button; an application that has none stored gets these on read, like the signin items.

items

The same page after changing a few labels and adding one HTML block:

customized

### Changes

- `object/application.go`, `object/application_util.go`: the `updatePasswordItems` field and its defaults.
- `web/src/pages/auth/UpdatePasswordPage.tsx`: the page.
- `web/src/lib/setting.tsx`, `LoginPage.tsx`, `AuthCallback.tsx`, `SamlCallback.tsx`, `App.tsx`: `goToUpdatePassword(applicationName)` and the two routes.
- `ApplicationUiCustomizationTab.tsx`: the signin items table is now a `PageItemsTable` component used for both lists; `signin-css.ts` carries the default CSS of the new items.
- Two new i18n keys in `application`: "Update password items" and "Submit button" (en and zh).

### Checked by hand

- A password sign-in with a flagged account lands on the new page; a wrong old password is rejected with the existing message; after a successful change the browser continues straight to the application's `redirect_uri` with a code, whether or not "Enable auto signin" is on.
- Editing the items in the application editor, saving, and reloading the page shows the new labels, placeholders, order and HTML block; hiding an item removes it.
- An application saved before this change has no stored items and renders the default set.


hsluoyz:
Thanks for the PR, the problem is real and the direction is right. We landed a slimmer version of it in https://github.com/casdoor/casdoor/commit/0352f4beaa5049b0037af145a2205ef51d433c67: the standalone `/update-password[/:app]` page in the auth layout that resumes the interrupted sign-in, plus the console lock for flagged accounts, but without the new `updatePasswordItems` field and its editor table, since the page can be styled through the application's existing custom CSS. It also passes the application name from the OAuth state in `AuthCallback` (the `application` state there is still `null` at that point) and keeps the old-password field rules of `PasswordModal`. Closing this one in favor of that commit.

oxsean:
@hsluoyz The slimmer approach does make more sense. Thanks for the explanation!
I use Casdoor extensively as the IAM provider for our internal SaaS platform, and I’d love to contribute some of our extensions back to the community. Is there a preferred chat or IM channel where we could discuss proposed changes in advance to make sure they align with the project’s direction and contribution guidelines?

hsluoyz:
@oxsean see: https://discord.gg/5rPsrAzK7S

PR #5861, opened 2026-09-18
feat: add Message-Authenticator protection to RADIUS

Hawaii-ol:
# Add Message-Authenticator protection to RADIUS

## Summary

This change adds `Message-Authenticator` protection to Casdoor's RADIUS MFA client and validates protected requests handled by Casdoor's built-in RADIUS server.

The RADIUS MFA client now signs every `Access-Request` and requires every corresponding `Access-Accept`, `Access-Reject`, or `Access-Challenge` to contain a valid `Message-Authenticator`. The built-in RADIUS server remains compatible with legacy clients: it only enforces `Message-Authenticator` when the incoming request contains the attribute, and it only adds the attribute to responses for those protected requests.

## Background

The Blast-RADIUS vulnerability (CVE-2024-3596) demonstrated that traditional RADIUS packets without a message signature are susceptible to mitm attacks. By exploiting weaknesses in the legacy MD5-based Response Authenticator construction, an attacker can modify a valid response and inject malicious authorization attributes into the forged packet.

Major RADIUS servers and vendors have since updated their security policies in response to this vulnerability. For EAP packets, RFC 2869 explicitly requires the Message-Authenticator attribute. The receiver must discard the packet if this attribute is missing or invalid. For PAP/CHAP, the standard does not mandate a Message-Authenticator; However, modern RADIUS implementations provide configurable or default policies to validate or require it. For example, FreeRADIUS provides explicit policies for requiring Message-Authenticator on Access-Requests and can enforce it conditionally for clients.

## How Message-Authenticator Works

Classic RADIUS over UDP uses a 16-byte Request or Response Authenticator based on MD5. For `Access-Request` packets, the Request Authenticator is random rather than a keyed integrity check. Response packets use an MD5 construction that includes the shared secret, but this construction is not equivalent to a modern message authentication code.

`Message-Authenticator`, defined as RADIUS attribute type 80, adds a keyed HMAC-MD5 over the complete RADIUS packet. During calculation, the attribute is present with a 16-byte zero value; that value is then replaced by the resulting HMAC. For a response, the calculation uses the original request authenticator in the authenticator field.

Despite using MD5 internally, HMAC-MD5 is not vulnerable to the chosen-prefix collision technique used against the legacy RADIUS Response Authenticator. The attribute therefore provides an independent integrity and authenticity check for the packet and binds its code and attributes to the shared secret.

`Message-Authenticator` does not encrypt RADIUS traffic. Usernames and most other attributes remain visible on the network, and the deployment still depends on the strength and protection of the shared secret. RADIUS over TLS or DTLS remains the preferred long-term transport where it is available.

### RADIUS authenticator package

A new `radius/authenticator` package is added to:

- Signs `Access-Request` packets with a 16-byte `Message-Authenticator`.
- Verifies that protected requests contain exactly one correctly sized and valid attribute.
- Signs `Access-Accept`, `Access-Reject`, and `Access-Challenge` responses using the original request authenticator.
- Verifies response `Message-Authenticator` values with constant-time comparison.
- Verifies the traditional RADIUS Response Authenticator in addition to the new HMAC check.
- Rejects missing, malformed, duplicate, or invalid `Message-Authenticator` attributes where protected communication is required.

### RADIUS MFA client

RADIUS MFA now uses `ExchangeWithMessageAuthenticator` instead of sending packets directly through `layeh.com/radius.Exchange`. As a result:

- Every MFA `Access-Request` includes a valid `Message-Authenticator`.
- The existing RADIUS library continues to validate the traditional Response Authenticator.
- Casdoor additionally requires and validates `Message-Authenticator` on every authentication response.
- A response that is unsigned, malformed, duplicated, or modified is treated as a RADIUS authentication error.

This prevents the MFA client from silently accepting a response protected only by the vulnerable legacy response construction.

### Built-in RADIUS server

The built-in RADIUS server applies conditional enforcement for backward compatibility:

- If an `Access-Request` contains `Message-Authenticator`, the server validates it before any authentication logic runs. Invalid protected requests are dropped without a response.
- Responses to a valid protected request include a valid `Message-Authenticator`.
- If a legacy `Access-Request` does not contain the attribute, the server keeps the previous behavior and does not add the attribute to its response.
- Accounting requests and responses are unchanged.

This provides end-to-end protection for upgraded clients without immediately breaking older clients that communicate with Casdoor's built-in server.

### External RADIUS servers used by MFA

This is intentionally stricter than the previous behavior. Any external RADIUS server used for Casdoor MFA must return a valid `Message-Authenticator` in `Access-Accept`, `Access-Reject`, and `Access-Challenge` responses. Servers that ignore the signed request but return an unsigned response will no longer interoperate with RADIUS MFA.

### Clients of Casdoor's built-in RADIUS server

Existing unsigned clients remain compatible. Upgraded clients can opt into protected exchanges simply by sending a valid `Message-Authenticator`; Casdoor will then validate the request and sign the response.

This compatibility mode is useful for migration, but it means the built-in server does not force Blast-RADIUS protection for every client. Deployments that require universal enforcement should ensure all clients send the attribute and should use network controls or a protected RADIUS transport to prevent downgrade to an unsigned exchange.

## Testing

The new tests cover:

- Signing and validating `Access-Request` packets.
- Detection of missing, duplicate, malformed, and tampered attributes.
- Signing and validating `Access-Accept`, `Access-Reject`, and challenge-compatible responses.
- Independent validation of the traditional Response Authenticator.
- Rejection of unsigned responses by the protected exchange helper.
- A local UDP request/response exchange using signed packets.
- Preservation of the original request authenticator when the server mutates a request into an `Access-Challenge`.

The affected `object` package and the application entry point also compile successfully with the new dependency structure.


hsluoyz:
Thanks for raising this, the Blast-RADIUS concern is valid. I've landed a reworked version in ab56581f40471789539593c3f773b921e0b99da6: the MFA client now always signs its Access-Requests, and there's a new "Require Message-Authenticator" switch on the RADIUS MFA provider (on by default for new providers) so existing deployments against unpatched RADIUS servers don't break. The built-in server validates and signs when the client sends the attribute, same as your approach. Closing this PR in favor of that commit.

PR #5869, opened 2026-09-20
fix: correct Ukrainian (uk) i18n translations

hz61p1:


hsluoyz:
@hz61p1 thanks for the contribution!

PR #5870, opened 2026-09-21
fix: validate update-user identifiers and authorize the resolved user

civil-king:
When `update-user` resolves `userId` and `owner`, it can replace the query `id` before updating the user. Validate that both identifiers refer to the same user when supplied together, and require `owner` when using `userId`.

Authorize the resolved user with the existing strict `CheckUserPermission` helper before merging request fields. This covers UUID-only requests as well as ID-based requests, while retaining self updates, organization-admin updates within the organization, and global-admin updates. It applies the same self/admin boundary used by the existing user permission helper; API-policy delegation alone does not bypass that boundary.

Validation:

- `gofmt` and `git diff --check` pass.
- 17 isolated regression cases pass against both the current upstream source and the local source. The harness extracts the actual `UpdateUser` and `CheckUserPermission` functions, substituting in-memory storage and controller dependencies. Cases cover conflicting identifiers, missing owner, UUID-only access, cross-user/cross-organization rejection, and permitted self/admin/app updates.
- The same harness fails against the unmodified upstream controller, including the conflicting-identifier and UUID-only unauthorized update cases.
- Full application build, HTTP-filter integration, and database integration tests have not been run. The local harness is not part of this minimal production-code commit.


hsluoyz:
not planned

PR #5872, opened 2026-09-21
fix: delete empty-owner records in record retention cleanup

### LINKED ISSUE #5871: record_retention_days never deletes audit records written for unauthenticated requests (empty `owner`)
**Describe the bug**

`Organization.RecordRetentionDays` is meant to bound the growth of the `record` audit table and enforce a deletion period. However, the cleanup only deletes rows whose `owner` equals the name of an organization that has a positive retention configured. Records written for **unauthenticated** requests (e.g. `login`, `login/oauth/access_token`, `oauth/register`, `device-auth`, `webauthn/signin/finish`) are stored with an **empty `owner`**, so they are never matched by the cleanup and are retained forever, regardless of any retention setting. These are also the rows that store the client IP address.

**Root cause (v4.7.0 source)**

`object/record.go` — `AddRecord()`:

```go
record.Owner = record.Organization
```

For a request without an authenticated user/organization context, `Organization` is empty, so `Owner` becomes `""`.

`object/record_cleanup.go` — `getOrgRecordRetentionDays()` builds a map keyed by **organization name** (only orgs with `RecordRetentionDays > 0`), and `cleanupOrgRecords()` deletes with:

```go
Where("owner = ?", owner)
```

No organization is named `""`, so rows with an empty `owner` never match and are never deleted.

**Steps to reproduce**

1. On any organization, set `Record Retention Days` to a small positive value.
2. Trigger some unauthenticated endpoints (a failed login, an OAuth token request, a registration attempt).
3. Wait for the cleanup (runs at startup and daily at 00:00), or lower the system clock past the cutoff.
4. Observe: rows with a populated `owner` older than the cutoff are deleted, but rows with an empty `owner` (the login/token/register entries, which carry the client IP) remain.

**Impact**

The retention feature cannot bound the part of the `record` table that typically grows fastest and that carries the most privacy-sensitive field (client IP of login/token/register attempts). For deployments that rely on `record_retention_days` to enforce a data-retention/erasure period (e.g. GDPR), this is a silent gap: the setting appears to be configured but never removes these entries.

**Expected behavior**

Record retention should be able to bound *all* audit rows, including those from unauthenticated endpoints. Possible directions:

1. In `AddRecord()`, when the organization/owner is empty, fall back to a well-known owner (e.g. the `built-in` organization) so the row is governed by a retention policy.
2. Or in `CleanupRecords()`, add an explicit sweep for empty-`owner` rows governed by a configurable global/`built-in` retention value.

**Version:** 4.7.0 (verified against source). The cleanup logic (`record_cleanup.go`) is recent; the `record.Owner = record.Organization` assignment is long-standing.

**Related:** #5692 (closed, "[Questions] There is no built-in retention/cleanup mechanism…") — that question predates the cleanup feature; this report is about the feature (now present) not covering empty-`owner` rows.

Tyagiquamar:
Problem: record retention cleanup only deletes rows whose owner matches an organization with RecordRetentionDays set. Rows from unauthenticated requests (login, token, register) are stored with empty owner, so they are kept forever, including client IPs. Fixes #5871. Root cause: AddRecord sets Owner from Organization, which is empty without auth context, while CleanupRecords only sweeps per-organization owners. Fix: CleanupRecords now also sweeps empty-owner rows using the strictest configured retention; no retention configured means nothing is deleted, as before. Tests: new TestGetRetentionCutoffs in object/record_cleanup_test.go covers empty map, single org, and strictest-wins. gofmt clean, go vet ./object/ clean, go test ./object/ -run TestGetRetentionCutoffs passes (3/3 subtests) in golang:1.25.8-bookworm.


Tyagiquamar:
Retriggering CI: the standalone Go-Linter job failed while downloading golangci-lint (HTTP 504). The Build workflow Go-Linter already passed with the same config. Reopening to rerun the lint workflow.

hsluoyz:
replaced by: https://github.com/casdoor/casdoor/commit/af0b022c2cd954197f5872bb0a1c41a597803dc2

